PR #4, opened 2019-10-14
Update README.md

chrismgonzalez:
Fixed an apparent type in the README.md under the Roadmap section.  

Added:
4. (optional) Build Android / iOS clients that also use these API endpoints


QuincyLarson:
@chrismgonzalez Nice catch!

PR #13, opened 2019-10-15
Adding a story where organizers can see participants attendance rate

MirzaChilman:
Adding a user story where organizers can see the attendance rate of the participants. There's a time where the seat is limited but got lots of people in waiting list. At the event day, some people doesn't come without a proper notice. This is a loss as there are some people that way more eager to join but probably got late notice of the event.

With attendance rate we can utilize it to message the participants that their attendance rate is low and the impact is that if it reach some point it will make them to not automatically accepted to get a seat on the event.

The acceptance should be manually triggered by the organizers instead of automatically by system (where if the seats is available then he/she will automatically get the seat)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ I can add a venue sponsor to the event with a link to their website as a way of 
 
 I can add a food sponsor to the event with a link to their website as a way of thanking them for food.
 
+I can see how many times a participant has come to the event as well as their attendance rate
+
 ### Roadmap
 
 1. Design the schema


mblonyox:
Should we add a ticketing system like eventbrite does? So organizer can easily validate quickly the user.

QuincyLarson:
@MirzaChilman Good idea. Ideally we'd be able to show the organizer the participant's entire rsvp/attendance history. We can incorporate this into our schema.

MirzaChilman:
> Should we add a ticketing system like eventbrite does? So organizer can easily validate quickly the user.

I never use eventbrite so I can't say a lot about that, will take a look into that

nik-john:
@all-contributors please add @MirzaChilman   to doc

PR #18, opened 2019-10-15
Add a nickname for a user

bernhard-hofmann:
Many users will prefer a nickname to their own name, so they should be able to define one.

### DIFF
--- a/data/ddl.sql
+++ b/data/ddl.sql
@@ -1,9 +1,10 @@
 create table users (
     id uuid primary key,
-    first_name text not null ,
+    first_name text not null,
     last_name text not null,
     email text not null,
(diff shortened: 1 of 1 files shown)


vkWeb:
@bernhard-hofmann They can put whatever they like in first name or last name. This should not be in MVP.

@QuincyLarson should be closed.

francocorreasosa:
Agree with @vkWeb, I make `last_name` optional in #26 so that should be enough to support nicknames.

QuincyLarson:
Thanks, @bernhard-hofmann. I agree with @vkWeb and @francocorreasosa that we can add this later if we need to.

PR #26, opened 2019-10-15
Adapt users to be external ID providers friendly

francocorreasosa:
- [x] Making `last_name` not required since it is not required in Meetup.com right now.
- [x] Adding `provider` as a way to track external ID providers like Google, Facebook, Auth0 when the user uses one of them.
- [x] Adding `provider_id` as a way to track which user is on one of those external providers.
- [x] Renaming `password` to `password_digest` to make clear that we'll need to use a hash instead of the actual plain text password, and making it optional in case of Social/IdP signup.

### DIFF
--- a/data/ddl.sql
+++ b/data/ddl.sql
@@ -6,14 +6,14 @@ create table users (
     password_digest text
 );
 
-create table social_provider (
+create table social_providers (
     id uuid primary key,
     name text
 );
 
-create table social_provider_user (
+create table social_provider_users (
     id uuid primary key,
-    provider_id uuid references social_provider(id) not null,
+    provider_id uuid references social_providers(id) not null,
     provider_user_id text not null,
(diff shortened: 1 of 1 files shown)


bernhard-hofmann:
Please make provider user a separate table to allow one user to use multiple provider logins and still be the same user entity.

francocorreasosa:
Done :)

ElSaico:
Another argument for making `last_name` optional is that some cultures allow for legal names to be mononyms (e.g. Indonesia).

francocorreasosa:
@bernhard-hofmann the current implementation allows one user to have as many social_provider_users as wanted.
https://github.com/freeCodeCamp/chapter/pull/26/files#diff-131f930065283444c7d6d03c472b044aR14

QuincyLarson:
This looks good to me.

PR #27, opened 2019-10-15
Add waitlist support in RSVPs

francocorreasosa:
Adds `in_waitlist` column that will help implement a whitelist system as the one that meetup.com has right now.

### DIFF
--- a/data/ddl.sql
+++ b/data/ddl.sql
@@ -62,11 +62,9 @@ create table rsvps (
     user_id uuid references  users(id),
     event_id uuid references events(id),
     date timestamptz not null,
-    on_waitlist boolean not null,
+    on_waitlist boolean not null default FALSE,
     primary key (user_id, event_id)
 );
-
-alter table rsvps alter column on_waitlist set default FALSE;
 
 create table user_bans (
     user_id uuid references users(id),
--- a/data/ddl.sql
+++ b/data/ddl.sql
@@ -61,8 +61,12 @@ create table user_groups (
 create table rsvps (
     user_id uuid references  users(id),
     event_id uuid references events(id),
+    date timestamptz not null,
+    in_waitlist boolean not null,
     primary key (user_id, event_id)
 );
+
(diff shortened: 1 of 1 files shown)


bernhard-hofmann:
White list or wait list? 😕

francocorreasosa:
Small typo in the commit name, the code is using the correct wording 😄 sorry about that

francocorreasosa:
Fixed, sorry again.

data/ddl.sql
```diff
@@ -61,11 +61,15 @@ create table user_groups (
...
     user_id uuid references  users(id),
     event_id uuid references events(id),
+    date timestamptz not null,
+    in_waitlist boolean not null,
     primary key (user_id, event_id)
 );
 
+alter table rsvps alter column in_waitlist set default FALSE;
+
 create table user_bans (
```
dmmulroy: Postgres enables inlining default values.  You can put this is inline in the create table statement: `in_waitlist boolean not null default false,`

dmmulroy: I'd also want to make sure w/ @QuincyLarson that wait listing is an MVP feature. I suspect that's its going to be easy to start bikeshedding the database before we actually have the API hooked up and using it (see various open PRs).

QuincyLarson:
@francocorreasosa I would say waitlists aren't MVP, like @dmmulroy pointed out. 

But then I'm sure people would immediately ask for this feature.

So yes - I think we can include them. I agree that `on_waitlist` sounds more natural.

francocorreasosa:
Changed it, thanks

data/ddl.sql
francocorreasosa: done, thanks

QuincyLarson:
@francocorreasosa Thanks again for creating this PR. 

And thanks to @dmmulroy and @vaibhavsingh97 for looking over this, too.

nik-john:
@all-contributors please add @francocorreasosa to code

PR #29, opened 2019-10-15
Initial stab at a GQL API and UI

orta:
This is a possible start based on the example inside [GraphQL Sequelize](https://github.com/mickhansen/graphql-sequelize/tree/master/examples)

I've ran out of time for today, but at least someone can continue from here rather than it be abandoned. running `yarn start` in the api package will launches the graphql, which has a working models section for both GQL and the db. It's set up to support associations fine with diamond dependencies in the current setup.

No pressure for this to be _the way_


packages/api/src/models/Group.js
```diff
@@ -0,0 +1,65 @@
...
+import { sequelize } from '../db';
+import { User } from './User';
+
+class Group extends Model {
+  static tableName = 'groups';
+  static Creator = undefined;
+
+  static associate(models) {
+    console.log("-----")
+    // Group.Groups = Group.belongsTo(User, {
```
Guusy: :eye: can remove this comment ?

packages/api/src/index.js
```diff
@@ -0,0 +1,48 @@
+// @ts-check
```
orta: Basically you get the tooling for TypeScript without using it

packages/api/src/index.js
```diff
@@ -0,0 +1,48 @@
...
+    users: {
+      type: userGroupsConnection.connectionType,
+      args: userGroupsConnection.connectionArgs,
+      resolve: userGroupsConnection.resolve
+    }
+  }
+})
+
+const schema = new GraphQLSchema({ query: root });
+const app = express();
```
orta: Someone in chat said folks agreed against graphic, in which case this PR can be closed

packages/api/src/index.js
shovon: Nice. A great start.

email2vimalraj:
I believe we should focus on putting the Open API specification first before diving into the code.

vaibhavsingh97:
> I believe we should focus on putting the Open API specification first before diving into the code.

We can use swagger for that, should I open a PR with basic specification?

QuincyLarson:
@orta Thanks for opening this. It's great to spike on this, though I agree we shouldn't merge any code until we've agreed the spec.

@email2vimalraj Would you agree with @vaibhavsingh97 that Swagger is the best place for this?

I'm closing this PR so it doesn't accidentally get merged, but we can continue the discussion here.

Wizyma:
Do we really need swagger with GraphQl ? The documentation of GraphQl is autogenerated, and there is tool like graphql-voyager to visualise the data flow wich is way more friendly than swagger.

email2vimalraj:
@QuincyLarson yes there is another PR which is opened by another member related to swagger. 
@vaibhavsingh97 it would be great if you can help us with that PR

PR #30, opened 2019-10-15
Add Next.js with API routes boilerplate

yakovlevyuri:


email2vimalraj:
We should first have the Open API specification built before even starting the code to get clarity.

QuincyLarson:
@yakovlevyuri Thanks for setting this up. I agree with @email2vimalraj that we should plan things first, then do this.

Once we're ready, we would welcome your help with this, though. I'm going to close this PR for now.

PR #37, opened 2019-10-15
Created tags table-gave users table tag id-updated readme with this u…

PrestonElliott:
Created tags table. Updated users table to add tag id column. Updated ReadMe with this user story added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ So far we have only two user roles: participants and group organizers
 
 - After the event, I will automatically get emails notifying me of subsequent events.
 
+- I can filter all events in my location by tag/interests.
+
 #### As an organizer
 
 - I can create a group.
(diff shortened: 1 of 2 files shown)


QuincyLarson:
@PrestonElliott Thanks for adding this, and for updating the readme as well.

email2vimalraj:
Sorry for commenting after merging it. Just had a question on this. Does tags belongs to a user or event?

If event, it has to be one to many (event to tags). In that case, we’ve to create another table to maintain the relationship. Please clarify

PrestonElliott:
@email2vimalraj  for example, the JavaScript tag is associated with a React event and JavaScript tag is also associated with a Vue.js event. This would provide the ability to filter/search events on the "JavaScript" tag and return both the React and Vue.js events. 

The tag shouldn't belong to a user, but it should enable users to search for events by tag.

nik-john:
@all-contributors please add @PrestonElliott  to doc code

PR #38, opened 2019-10-15
README - Docker image -> Docker container

ScottBrenner:
Proposing a quick change in the README - this will be a self-hosted Docker _container_, not image.


QuincyLarson:
@ScottBrenner Thanks for pointing this out. Now I know the difference between a Docker image and a Docker container. 😄

PR #55, opened 2019-10-16
feat: adding logo to sponsor table

wendelnascimento:
I guess it would be nice to have the sponsor logo (a link to the logo would be stored on the database) as meetup.com currently displays it and from my experience the sponsors would want to have their logo displayed alongside their name and website.


data/ddl.sql
```diff
@@ -47,6 +47,7 @@ create table sponsors (
     id uuid primary key,
     name text not null,
     website text,
+    logo text,
```
viniciusalonso: Maybe could more intuitive to change this variable name to `logo_url` or `logo_path`.

wendelnascimento: Sure, logo_path seems better. I'll change it

francocorreasosa:
This is a good change, but I'm not totally sure this should go to the MVP. Since this change implies handling file uploading to a server and managing that extra data.

I'm not closing this one but leaving it here till we get to a consensus on the scope.

wendelnascimento:
@francocorreasosa I've got your point, maybe we can make the user submit the image url instead of submitting the file itself, so we don't need to handle this. I think that for the MVP it will work well and it's just one optional field on the request

snekolen:
Does this require putting the logo onto the website using html?

vkWeb:
> @francocorreasosa I've got your point, maybe we can make the user submit the image url instead of submitting the file itself, so we don't need to handle this. I think that for the MVP it will work well and it's just one optional field on the request.

This sounds good. We should apply this change @francocorreasosa :)

viniciusalonso:
I'm not sure but is there a user story to it? If does not exists is better us write one to detail this specific case.

vkWeb:
> I'm not sure but is there a user story to it? If does not exists is better us write one to detail this specific case.

@viniciusalonso Its on the user story. Check out README.

vkWeb:
@allella Resolve and squash merge.

wendelnascimento:
> @allella Resolve and squash merge.

I've resolved the conflicts, guess we can merge this

chrismgonzalez:
I think this would be good to add to the MVP as an image URL.  I've added it to the MVP user stories.  I think we are ready to merge this one in when ready.

allella:
@vkWeb I'm going to be mostly not near a computer this weekend, so @chrismgonzalez or the other mods may be faster at merging.  Though, I'm around for a bit tonight if this is still ready to roll.

I've put your name up to Quincy for write perms too.

allella:
A logo is part of the MVP https://github.com/freeCodeCamp/chapter/issues/103 so taking the advice above and merging this in.

nik-john:
@all-contributors please add @wendelnascimento to code

PR #57, opened 2019-10-16
Reuse user_groups table table for user bans

midned:
With this we can avoid using an extra table for banning users from groups.

We can actually use the already-existent pivot table to hold that information

### DIFF
--- a/data/ddl.sql
+++ b/data/ddl.sql
@@ -75,6 +75,7 @@ create table event_sponsors (
 create table user_groups (
     user_id uuid references users(id),
     group_id uuid references groups(id),
+    banned boolean default false,
     primary key (user_id, group_id)
 );
 
@@ -86,8 +87,3 @@ create table rsvps (
     primary key (user_id, event_id)
 );
 
-create table user_bans (
-    user_id uuid references users(id),
-    group_id uuid references groups(id),
-    primary key (user_id, group_id)
-);


dmmulroy:
I intentionally made these separate tables for the use case that a user leaves an organization, deletes their account, or something similar we can still easily maintain the ban information.  But I'm also open to this if y'all think it'd be better 👍

midned:
> I intentionally made these separate tables for the use case that a user leaves an organization, deletes their account, or something similar we can still easily maintain the ban information. But I'm also open to this if y'all think it'd be better 👍

I think that's a good reason actually. We could handle that by not allowing the user to leave a group where he or she is banned; and also we won't show the user in the list of members if the banned flag is true.

In the case of account deletion then maintaining the ban record won't give any information since we won't have any information of the banned user (unless we use soft deletes).

I can see advantages in both approaches, I'm personally 65% in favor of using the same table with the `banned` flag, and 35% for the use of separate tables

data/ddl.sql
```diff
@@ -75,6 +75,7 @@ create table event_sponsors (
 create table user_groups (
     user_id uuid references users(id),
     group_id uuid references groups(id),
+    banned boolean default false,
```
francocorreasosa: It would be useful to convert this into a date time field, so we can keep track of how many time an user has been banned. In the future we could also implement auto unbans (or temporal bans). 
```suggestion
    banned_at timestamptz,
```

midned: I totally agree with this change. @dmmulroy @vaibhavsingh97 What do you guys think?

francocorreasosa:
Also, on top of my previous comment, I’m not totally sure we want users to be banned from the entire platform. Who should be allowed to ban an user from all groups?

midned:
> Also, on top of my previous comment, I’m not totally sure we want users to be banned from the entire platform. Who should be allowed to ban an user from all groups?

This change only allows for users to be banned from a group

francocorreasosa:
Oops you're right, misread the table name. Edited my comment.

data/ddl.sql
vaibhavsingh97: > we can keep track of how many time an user has been banned.

`banned_at` field will be updated many times, so how can we track if the user is banned 1 time or 2 times? We will only get the last banned date if we are using `timestampz`

> In the future we could also implement auto unbans (or temporal bans).

So probably we will need to have a separate table for this thing as then we have too many things to consider. Also, maintaining another table will record logs like the reason for banning, no. of times a user gets banned, also the state of ban i.e temp/permanent.

bernhard-hofmann:
I prefer the separate table for a few reasons:
1. As mentioned by @dmmulroy , the row can persist even after the user leaves a chapter
1. The `user_bans` table can (and should IMO) also have columns for `reason` and `created_at` so that we know why a user was banned and when.

latchkostov:
I prefer a separate table for the reason named. We should be able to keep track of multi bans and information regarding those bans. In addition, a bans table allows a user to be removed from the organization, and we don't have to do any weird one off things like not allow a user to leave an org because they are banned.

ceciliaconsta3:
I agree with using a separate ban table. Is this an MVP feature? Maybe not since we need to verify users (phone number?) so a banned person doesn't just create a new account with an alternate email address. Either way, if I'm understanding all correctly, sounds like this PR needs to be closed.

vkWeb:
@ceciliaconsta3 Yes indeed banning a user is an MVP feature. I agree with you all that we should keep `user_bans` in a separate table to handle hard deletion.

Closing this as the consensus is: Banned users should be in separate table.

Thanks, @midned for trying to help us. This was your first time here so I know it feels bad to see your PR closed but don't worry we need a ton of help in implementing MVP features. You can find issues we need help with [here](https://github.com/freeCodeCamp/chapter/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22) :)

PR #59, opened 2019-10-16
Reorder table definitions so the script can run

dmmulroy:
The `events` table has a dependency on `tags` so `tags` must be declared/created first.


viniciusalonso:
This PR is duplicated #51

PR #88, opened 2019-10-18
Format Contributing section in README

ScottBrenner:

Quick formatting change to the Contributing section in the README - "Contribute" shouldn't be capitalized here, and the sentence shouldn't be in a header like this.


allella:
I think someone was using a heading to draw attention too it, so we could compromise with a bold or some other formatting that's less semantic than a H3 tag.  You want to apply some formatting to retain that earlier desire, or just have it be a normal looking link?

PR #119, opened 2019-10-21
Add github ci workflow

Lobarr:

Closes #XXXXX

Description of changes
- Integrated github action for CI

### DIFF
--- a/.github/workflows/ci-workflow.yaml
+++ b/.github/workflows/ci-workflow.yaml
@@ -5,6 +5,7 @@ jobs:
     name: build and test
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         node-version: [10.x]
     steps:
--- a/.github/workflows/ci-workflow.yaml
+++ b/.github/workflows/ci-workflow.yaml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [11.x]
+        node-version: [10.x]
     steps:
       - uses: actions/checkout@v1
       - name: Use Node.js ${{ matrix.node-version }}
--- a/.github/workflows/ci-workflow.yaml
+++ b/.github/workflows/ci-workflow.yaml
@@ -1,0 +1,23 @@
+name: CI Workflow 
+on: [push, pull_request]
+jobs:
+  build:
+    name: build and test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [11.x]
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with: 
+          node-version: ${{ matrix.node-version }}
+      - name: Install dependencies
+        run: npm ci
+      - name: Build next.js
+        run: npm run next:build
+      - name: Run tests
+        run: npm run test
+        env:
+          NODE_ENV: development


.github/workflows/ci-workflow.yaml
```diff
@@ -0,0 +1,23 @@
+name: CI Workflow 
+on: [push, pull_request]
+jobs:
+  build:
+    name: build and test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
ScottBrenner: No need to use a `matrix` strategy if there's only one version of Node being tested against.

.github/workflows/ci-workflow.yaml
```diff
@@ -0,0 +1,23 @@
...
+        node-version: [11.x]
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with: 
+          node-version: ${{ matrix.node-version }}
+      - name: Install dependencies
+        run: npm ci
```
ScottBrenner: I don't know much about JavaScript stuff, but I imagine the commands used here should match  what's in https://github.com/freeCodeCamp/chapter/blob/93f6aad27c81a86f655f27ca63d9069a442d06ff/package.json#L9-L23

.github/workflows/ci-workflow.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Build next.js
+        run: npm run next:build
+      - name: Run tests
+        run: npm run test
```
ScottBrenner:
```suggestion
        NODE_ENV=test npx jest --coverage --verbose
```

based on https://github.com/freeCodeCamp/chapter/blob/93f6aad27c81a86f655f27ca63d9069a442d06ff/package.json#L21

.github/workflows/ci-workflow.yaml
```diff
@@ -0,0 +1,23 @@
+name: CI Workflow 
+on: [push, pull_request]
+jobs:
+  build:
+    name: build and test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [11.x]
```
ScottBrenner:
```suggestion
        node-version: [10.x]
```

based on https://github.com/freeCodeCamp/chapter/blob/bea3dcc1c8174daa3dc942b59380e1cf8c67b373/docker-compose.yml#L4

Recommend reworking the below steps in run in a `node:10-alpine` Docker container as well.

.github/workflows/ci-workflow.yaml
Lobarr: Included this incase testing multiple versions would be required later. Could remove if not.

ScottBrenner:
The workflow doesn't actually run, since the repository is configured so that
> Only repository collaborators may create workflows.

so there's no way to see if it actually works 😬

.github/workflows/ci-workflow.yaml
Lobarr: The `npm run test` script, in this case, would be equivalent to the one defined in the package.json hence would resolve to `NODE_ENV=test npx jest --coverage --verbose`

.github/workflows/ci-workflow.yaml
vaibhavsingh97: @ScottBrenner `npm ci` install a project with a clean slate. For more info: https://docs.npmjs.com/cli/ci.html.

.github/workflows/ci-workflow.yaml
vaibhavsingh97: Agreed with @Lobarr 👍

ScottBrenner: Cool, got it!

.github/workflows/ci-workflow.yaml
```diff
@@ -0,0 +1,23 @@
+name: CI Workflow 
+on: [push, pull_request]
+jobs:
+  build:
+    name: build and test
+    runs-on: ubuntu-latest
+    strategy:
```
ScottBrenner:
```suggestion
    strategy:
      fail-fast: false
```

Am a fan of disabling the [`fail-fast`](https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) strategy, to let the full workflow run even when something fails early. Suppose it's up for debate.

timmyichen:
Does github CI support docker?  Later when we have integration tests with a test database we'd need to be able to run our full docker setup to properly test it.

.github/workflows/ci-workflow.yaml
vaibhavsingh97: what's the use case here? suppose we have test cases are running and if any of the test cases fails then CI should fail, correct me if I am wrong here

ScottBrenner:
Yes, see https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idservices.

timmyichen:
Oh, neato - thanks!

.github/workflows/ci-workflow.yaml
Lobarr: I'm okay with this too so I'd add it.

vaibhavsingh97: @Lobarr It's open to discussion, you don't necessarily have to add it

ScottBrenner: This is helpful with the `matrix` strategy, say the version matrix is like
```
        node-version: [9.x, 10.x, 11.x]
```

When `fail-fast` is `true`, say anything fails in the `9.x` builds - it cancels all the other builds. But we're still interested in the results for other versions, no? Disabling `fail-fast` continues all in-progress jobs if any `matrix` job fails.

vaibhavsingh97: Ohh! Gotcha - Thanks 👍

nik-john:
@all-contributors please add @Lobarr   for platform

nik-john:
@all-contributors please add @Lobarr for platform

PR #120, opened 2019-10-21
Initialize sequelize

Zeko369:
- [x] Add psql to docker compose
- [ ] Initialize sequelize and sequlize-ts
- [ ] Create one model, migration and example route for that model


package-lock.json
```diff
@@ -5224,7 +5224,8 @@
         },
```
vaibhavsingh97: can you please help me, why you updated `package-lock.json`?

docker-compose.yml
```diff
@@ -8,6 +8,16 @@ services:
...
+  db:
+    image: postgres
+    environment:
+      - POSTGRES_USER=root
+      - POSTGRES_PASSWORD=password
+      - POSTGRES_DB=chapter
+    ports:
+      - '54320:5432'
+    volumes:
+      - ./postgres-data:/var/lib/postgresql/data
```
vaibhavsingh97:
```suggestion
    volumes:
      - ./postgres-data:/var/lib/postgresql/data
    ports:
      - '54320:5432'
```

package-lock.json
Zeko369: Somehow I didn't add package.json

allella:
Pinging @oyewoas on this Sequelize topic since that skill was offered in #11

oyewoas:
> Pinging @oyewoas on this Sequelize topic since that skill was offered in #11

Alright @allella

docker-compose.yml
```diff
@@ -8,6 +8,16 @@ services:
       - .:/usr/chapter
     ports:
       - '8000:8000'
+  db:
+    image: postgres
```
timmyichen: We should probably peg this to a specific version e.g. `postgres:11`

docker-compose.yml
```diff
@@ -8,6 +8,16 @@ services:
...
+    environment:
+      - POSTGRES_USER=root
+      - POSTGRES_PASSWORD=password
+      - POSTGRES_DB=chapter
+    volumes:
```
timmyichen: Same as above - not necessary to persist the psql data (wouldn't need a volume in this case) - let me know your thoughts

.gitignore
```diff
@@ -62,3 +62,6 @@ typings/
 
 # .DS_Store
 .DS_Store
+
+# PSQL data
+postgres-data/
```
timmyichen: I think we don't need to persist the psql data on the host - thoughts?

docker-compose.yml
```diff
@@ -8,6 +8,16 @@ services:
       - .:/usr/chapter
     ports:
       - '8000:8000'
```
timmyichen: we should also ensure that `app` depends on `db` using docker-compose's `depends_on` (otherwise server may try to connect before postgres initializes)

vaibhavsingh97:
@Zeko369 Let me know if you stuck somewhere. We can help you, just ping me on discord or any mods. Also, please resolve merge conflict

QuincyLarson:
@Zeko369 Thanks again for pushing forward with the sequelize / sequelize-ts integration.

Have you had a chance to make the suggested changes and finish your checklist?

This is now critical path. Please let us know if we can do anything to help.

vaibhavsingh97:
@Zeko369 can you please give me push access so that I can help you out, and help you complete the checklist

timmyichen:
Alternatively since this PR is good on its own it could be merged and follow up PRs could be made

config/db.json
```diff
@@ -0,0 +1,26 @@
...
+    "password": "password",
+    "database": "chapter",
+    "host": "127.0.0.1",
+    "dialect": "postgres",
+    "port": 54320
+  },
+  "test": {
+    "username": "postgres",
+    "password": "password",
+    "database": "chapter",
```
timmyichen: We should use a different db for testing, one that we bring up and tear down when testing. Not essential for this pr, just leaving a note

package.json
```diff
@@ -46,21 +46,30 @@
...
     "@types/jest": "^24.0.19",
     "@types/morgan": "^1.7.37",
     "@types/next": "^8.0.6",
-    "@types/node": "^12.11.1",
+    "@types/node": "^12.11.2",
+    "@types/pg": "^7.11.2",
     "@types/react": "^16.9.9",
     "@types/react-dom": "^16.9.2",
+    "@types/styled-components": "^4.1.19",
+    "@types/validator": "^10.11.3",
```
timmyichen: added but unused (we dont have the normal validator package anyway)

server/models/index.ts
```diff
@@ -0,0 +1,13 @@
+import { Sequelize } from 'sequelize-typescript';
+import { Chapter } from './chapter';
+
+export const initSequelize = () =>
+  new Sequelize({
+    dialect: 'postgres',
+    username: 'postgres',
+    password: 'password',
+    database: 'chapter',
```
timmyichen: i think these are better to pass as env variables to the docker-compose instance and use `process.env.DB_USERNAME` or equivalent here (env vars can then be set for production)

server/index.ts
```diff
@@ -4,12 +4,17 @@ import morgan from 'morgan';
...
+import chapterRouter from 'server/routers/chapter';
+
+import { initSequelize } from 'server/models';
 
 const app: express.Application = express();
 
 nextjs.nextApp.prepare().then(async () => {
   const port = process.env.PORT || 8000;
 
+  initSequelize();
```
timmyichen: is this enough or do we have do something like `await initSequelize().authenticate()`?  connecting to the db should be async

server/models/index.ts
vaibhavsingh97: Also, same for all the fields

server/models/index.ts
```diff
@@ -0,0 +1,13 @@
...
+    host: 'db',
+    port: 5432,
+    models: [Chapter],
```
vaibhavsingh97: instead of defining models like this, we can dynamically load and define models. Also, what happens if we have associations in a model?

jacobbogers:
Hi, how can I help out? My skills are mainly on DB and backend of systems, (front end is still very weak)

package.json
Zeko369: Sequelize has it as a peer dependency, so we do use it, but it's just not in our dependencies

timmyichen:
On second thought @Zeko369 any particular reason for the pattern of having `initSequelize()` generate the sequelize instance vs having it be created and exported as part of the `db.ts` file since it's synchronous?  We'd need to access the sequelize instance in other places e.g. when we want to do transactions

server/db.ts
```diff
@@ -9,5 +8,5 @@ export const initSequelize = () =>
     database: process.env.DB_NAME || 'chapter',
     host: 'db',
     port: 5432,
-    models: [Chapter],
```
Haxified: you should always do this with path.join() instead

server/migrations/20191025082945-create-chapter.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('Chapters', {
```
timmyichen: lowercase c chapters

package.json
```diff
@@ -46,21 +46,30 @@
...
   },
   "devDependencies": {
     "@testing-library/react": "^9.3.0",
+    "@types/bluebird": "^3.5.28",
     "@types/express": "^4.17.1",
     "@types/jest": "^24.0.19",
     "@types/morgan": "^1.7.37",
     "@types/next": "^8.0.6",
-    "@types/node": "^12.11.1",
+    "@types/node": "^12.11.2",
```
ScottBrenner: Isn't Node 10 being used currently?
https://github.com/freeCodeCamp/chapter/blob/c1cf448ba0851e626e65013ae8867f8b7d94cb1a/docker-compose.yml#L4

allella:
@timmyichen it looks like @vaibhavsingh97 is still showing as requesting changes. @vaibhavsingh97 can you review and approve if this is ready.

Also, the package-lock.json is showing in conflict.

Zeko369:
@allella @timmyichen I completely forgot about this PR, I fixes most of the requests, I only have one question. What are we going to do about the seeds?, default `sequzelie-cli` doesnt work with ts files. I'll try `sequelize-cli-typescript`. Also the build from master is failing I'll look into it?

Zeko369:
@nik-john I dont know how you managed to get the CI pass for #158 , because I just merger master and you had a TS bug that stops it from compiling it on the CI.
```jsx
<StyledComponentsThemeProvider theme={themeObject}>
  <>
    <CssBaseline />
    <Component {...pageProps} />
  </>
</StyledComponentsThemeProvider>
```, you missed the fragment :)

vaibhavsingh97:
> On second thought @Zeko369 any particular reason for the pattern of having initSequelize() generate the sequelize instance vs having it be created and exported as part of the db.ts file since it's synchronous? We'd need to access the sequelize instance in other places e.g. when we want to do transactions

@Zeko369 what are your thoughts? IMO, we should intitalize the database connection and export the connection to app, so that we can reuse it to query the database.

nik-john:
@all-contributors please add @Zeko369   for code tool

nik-john:
@all-contributors please add @Zeko369 for code tool

PR #121, opened 2019-10-21
Add styled-components, config

nik-john:

This PR adds [Styled Components](https://www.styled-components.com) for styling. This lib is fairly familiar to most folks who have been in the React community for some time. From the offical docs:

> Apart from the improved experience for developers, styled-components provides:
> 
> - **Automatic critical CSS**: styled-components keeps track of which components are rendered on a page and injects their styles and nothing else, fully automatically. Combined with code splitting, this means your users load the least amount of code necessary.
> - **No class name bugs**: styled-components generates unique class names for your styles. You never have to worry about duplication, overlap or misspellings.
> - **Easier deletion of CSS**: it can be hard to know whether a class name is used somewhere in your codebase. styled-components makes it obvious, as every bit of styling is tied to a specific component. If the component is unused (which tooling can detect) and gets deleted, all its styles get deleted with it.
> - **Simple dynamic styling**: adapting the styling of a component based on its props or a global theme is simple and intuitive without having to manually manage dozens of classes.
> - **Painless maintenance**: you never have to hunt across different files to find the styling affecting your component, so maintenance is a piece of cake no matter how big your codebase is.
> - **Automatic vendor prefixing**: write your CSS to the current standard and let styled-components handle the rest.

To add on, **theming** is going to be important for Chapter. We can use the built-in CSS-in-JS solution for styling but considering that folks would like to white-label their own Chapters, and for DRY and general consistency we need to have a global theme and a Styleguide and I am yet to see something that is as mature as styled-components for this. I don't see any flip sides to this as it is an overwhelmingly popular lib in the community. It's best we decide on this at the outset so we don't have get into refactor hell. 

Oh and did I say Styled Components fully supports **server-side rendering**? ❤️ 

**PS:** I have also added a sample global theme file where we can add any global colors/shadows/text styles etc. This file can be potentially overriden if someone wants to white-label their Chapter instance.

I also see us needing some form of UI library that will help with Grid layout - suggestions welcome. [styled-bootstrap](https://github.com/bootstrap-styled) seems to be a solid option


.babelrc
```diff
@@ -1,15 +1,28 @@
...
       {
-        "root": ["."],
+        "root": [
+          "."
+        ],
         "alias": {
           "client": "./client",
-          "server": "./server"
+          "server": "./server",
+          "styles": "./styles"
```
timmyichen: will `styles/` ever be accessed from `server/`?  if not, can we put it in `client/`?

vaibhavsingh97:
> I also see us needing some form of UI library that will help with Grid layout - suggestions welcome. styled-bootstrap seems to be a solid option

Can we use CSS Grid, and flexbox as a fallback ?

.babelrc
timmyichen: Since `next.js` is taking care of SSR and our current setup has us grabbing components out of `client/` from `pages/` where the SSR magic happens, I think it makes sense for styles to also be in `client/`.  I haven't toyed around with any advanced config around SSR though so I'm not sure if we'd need to import anything from `styles/` to the `server/` directory. Would you happen to know of any instances where it would be needed to do such an import?

allella:
@abidRahim mentioned using Styled Components in #11 so I'm pinging on this Styled Components topic as it might be a good time to jump in.

ScottBrenner:
Recommend rebasing this PR to run the proposed changes against the CI workflow!

.babelrc
nik-john: @timmyichen Sorry - I've been swamped all day. I'm just seeing this. So the scenario I was talking about is when we need to **inline critical above-the-fold CSS** like mentioned [here](https://github.com/emotion-js/emotion/issues/1178#issuecomment-520493094). Lighthouse throws warnings when moderate sized CSS files are loaded via `link` as the import is render-blocking. An instance of this is if we are using a light enough Grid/Flex framework, it might make sense to inline it. 

Here are some instances where we do inlining: [1](https://github.com/zeit/next.js/blob/9c4eefcdbf4058b65c8b88fe98387166acad127b/examples/with-babel-macros/pages/_document.js#L9), [2](https://github.com/zeit/next.js/blob/9c4eefcdbf4058b65c8b88fe98387166acad127b/examples/with-cxs/pages/_document.js#L17), [3](https://github.com/zeit/next.js/blob/5782651e92ef07845e9d251981ac3ba2ced27e11/examples/with-fela/pages/_document.js#L27)

The way we are hooking up `styled-components` on [this PR](https://github.com/freeCodeCamp/chapter/pull/121/files#diff-8f3a5cf5019173ef1f56a54e45e44747R20) is also by inlining the styles although it isn't as straight forward as that.

nik-john:
@ScottBrenner Done 👍 CI workflow looks sick 🔥

joelrozen:
I've not used styled-bootstrap, but have implemented styled-system (https://styled-system.com/) into many projects and it is a dream to work with. I can't talk to its pros and cons, but will leave here as a suggestion

nik-john:
@here Could we merge this PR if there aren't any major comments please? I'll rebase ir In an hour

allella:
I'll defer to one of the more tech-stack aware maintainers, like @francocorreasosa, to merge this one.

pages/_app.tsx
```diff
@@ -0,0 +1,15 @@
+import App from 'next/app';
+import React from 'react';
+import { ThemeProvider } from 'styled-components';
+import theme from 'styles/theme';
+
+export default class MyApp extends App {
+  render() {
+    const { Component, pageProps } = this.props;
+    return (
+      <ThemeProvider theme={theme}>
```
vaibhavsingh97: are we going to provide dark theme in the future?

.babelrc
```diff
@@ -1,15 +1,28 @@
...
+      "styled-components",
+      {
+        "ssr": true,
+        "displayName": true,
+        "preprocess": false
+      }
     ]
   ]
-}
+}
```
vaibhavsingh97:
```suggestion

```

pages/_app.tsx
SeanRParker: If so I'd recommend with this media query `prefers-color-scheme:dark {}`

vaibhavsingh97: Browser coverage is very less for `prefers-color-scheme:dark {}`
image

nik-john: It's something we can look at - one of the advantages of using a global theme is that it's one source of truth for the entire application and whitelabelling and re-theming are a walk in the park

SeanRParker: Wow, in that case, I'd favor something with more adoption then; like css vars you could change with a toggle or something. I'm sure ThemeProvider is fine too.

nik-john:
Merging this in now, will raise another PR for the UI lib

nik-john:
@all-contributors please add @nik-john   for doc tool

nik-john:
@all-contributors please add @nik-john for doc tool

PR #128, opened 2019-10-22
Rewrite routing

Haxified:

I rewrote how the routing works in the server folder, making it (imo) much easier to read and understand, using a class based approach. I also removed all uses of `export default`, as many members of the TypeScript community think of it as a bad practise, as it causes inconsistent naming.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5262,7 +5262,8 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -5627,7 +5628,8 @@
         },
         "safe-buffer": {
           "version": "5.1.2",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "safer-buffer": {
           "version": "2.1.2",
@@ -5675,6 +5677,7 @@
         "strip-ansi": {
           "version": "3.0.1",
           "bundled": true,
+          "optional": true,
           "requires": {
             "ansi-regex": "^2.0.0"
           }
@@ -5713,11 +5716,13 @@
         },
         "wrappy": {
           "version": "1.0.2",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "yallist": {
           "version": "3.0.3",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         }
       }
     },
--- a/server/index.ts
+++ b/server/index.ts
@@ -1,9 +1,9 @@
 import 'module-alias/register';
 import express from 'express';
 import morgan from 'morgan';
-import nextjs from 'server/lib/next';
+import { nextjs } from 'server/lib/next';
 import { responseErrorHandler } from 'express-response-errors';
-import exampleRouter from 'server/routers/exampleRouter';
+import { ExampleRouter } from './routers/exampleRouter';
 
 const app: express.Application = express();
 
@@ -18,7 +18,7 @@ nextjs.nextApp.prepare().then(async () => {
   app.use(express.json());
   app.use(express.static('public'));
 
-  app.use(exampleRouter);
+  app.use(new ExampleRouter().router);
 
   app.use(responseErrorHandler);
 
--- a/server/lib/next.ts
+++ b/server/lib/next.ts
@@ -6,7 +6,7 @@ const isDev = process.env.NODE_ENV !== 'production';
 const nextApp = next({ dev: isDev });
 const handle = nextApp.getRequestHandler();
 
-const nextjs = {
+export const nextjs = {
   nextApp,
   handle,
   render: (
@@ -16,5 +16,3 @@ const nextjs = {
     query: { [s: string]: any } = {},
   ) => nextApp.render(req, res, page, query),
 };
-
-export default nextjs;
--- a/server/routers/Router.ts
+++ b/server/routers/Router.ts
@@ -1,0 +1,11 @@
+import * as express from 'express';
+
+export abstract class Router {
+  public router: express.Router = express.Router();
+
+  public constructor() {
+    this.registerRoutes();
+  }
+
+  protected abstract registerRoutes(): void;
+}
--- a/server/routers/exampleRouter.ts
+++ b/server/routers/exampleRouter.ts
@@ -1,26 +1,27 @@
-import express from 'express';
-import asyncHandler from 'express-async-handler';
+import { Router } from './Router';
+import { Request, Response } from 'express';
 import { BadRequestError } from 'express-response-errors';
 
-const router = express.Router();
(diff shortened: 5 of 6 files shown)


Zeko369:
I think that we should go through this and decide if this is better, because we're trying to make it as simple as it could be for newcomers to contribute, and this just seems like an unnecessary abstraction.
Tagging @timmyichen @QuincyLarson

Haxified:
I don't really think this is making anything too hard. OOP isn't very hard to understand, and I think even if you didn't understand it you could get a pretty good understanding of how the system worked without understanding OOP. Also, the old way this was written uses none of TypeScript's features, meaning js would of been an easier option anyway.

GeniaT:
@Haxified Would this apply we are going to use a OOP approach in the rest of the code base? If yes, it will definitely be a barrier to newcomers as most of node courses / examples aren't written that way. If not, consistency should be king. OOP can be easy to read but to contribute properly in OOP is another challenge...

Haxified:
Well, React is basically 100% OOP (obviously functional components etc but you still need the ideas of oop). OOP isn't exactly a very complex subject, and as I said before, you could just copy how the example controller works.

slikts:
Modern React with hooks has little or no semblance with OOP.

Verbosity or code volume is a significant source of complexity, and it's not justified here, since all that the abstract class does is initialize a router. Using inheritance is also not in line with the trend favoring composition.

Haxified:
A couple of points:
- I understand hooks have no direct correlation to OOP, but imo if you don't understand OOP you probably shouldn't be contributing to projects like this.
- Imo, just because you *can* make everything using functional components, dosen't mean you have to. Sometimes class based components are easier to develop and read.
- Currently, the abstraction is very simple (although imo still better then the system you had before). This dosen't mean that it always has to be this way. In the future, you could add more to the base class, making development easier.
- If we're talking about making it easy for beginners, is TS even the right way to go? The code before was valid es6, meaning there is no point of adding an extra later of abstraction if nobody's gonna use it. 
- In personal projects I normally take stuff like this further, by seperating stuff out into services, controllers, entities, stuff like that, using decorators and stuff. I understand that this is not at all understandable by beginners, and thats why i opted to go with this simpler routing system.

slikts:
The basic point is that verbosity makes code harder to read and understand, so the changes do the opposite of their description.

Nekoniri:
I don't see how this simple Routing abstraction would make it much more difficult from beginners to understand. This is really the basics of TypeScript we're talking about. It's much easier to read, and it encourages beginning TS developers to learn to properly structure their projects.

slikts:
Verbosity is harder to read, and this is particularly verbose with all the access modifiers. There's a segment of TS users who write code as if it were C# or Java, but that's not a consensus approach.

For an abstraction to be a simplification it needs to hide details, but this arguably just shuffles them around.

There's also good reasons to generally consider inheritance as poor design, so the opposite of proper.

Haxified:
I don't see how this is verbose at all.

GeniaT:
> * is TS even the right way to go? The code before was valid es6, meaning there is no point of adding an extra later of abstraction if nobody's gonna use it.

I read somewhere that we would start declaring files as .ts but with almost no Typescript involved to keep it easy but in same time we prepare the ground to gradually add types in the future.

slikts:
It doubles the number of changed lines for no change in functionality. The explicit `public` access modifiers in particular are verbosity carried over from C#. The verbosity could be justified if it were a simplifying abstraction, but it's not that either.

Haxified:
It doubles the line count until we start adding actual features. Also adding `public` isn't necessary, but i always like to define the access modifiers. If people think its verbosity i will happily remove them. This is definitely justified as it improves code readability. Compare the old routes to the new system and tell me it dosen't look better.

slikts:
I've obviously looked at the previous version and it's more readable since there's less code without the unnecessary abstraction. Just looking more like C# doesn't make code more proper, readable or better, and in this case it's even the opposite.

Haxified:
So i asked several members of my community (the official typescript discord server) with this image: https://imgur.com/a/iO7wbKI . These are the responses I got:
- https://imgur.com/a/r70it0U
- https://imgur.com/a/7SOrZxq
- https://imgur.com/a/BcJLN6Y
- https://imgur.com/a/wG71mNi
- https://imgur.com/a/Cgl5FDU

slikts:
Not sure if a poll could be more biased than asking a group selected by you in a server that you own. The answers could be relevant if they were substantiated with arguments, but the closest that one comes is to mention reusability, which doesn't really apply to an abstraction that does nothing.

I'd like to ask what is supposed to be unreadable about just defining routes with `app.get()`. It's clear at a glance what it does, as opposed to having to understand an abstraction that adds several steps of indirection and increases the code volume, while not actually abstracting away any details.

Haxified:
First - I didn't select the group, i sent out a public message in servers and those were the responses. I also dont see how me owning the server has anything to do with it, its not like i ban people if they do not agree with me. Second, I was going to do something like this:
```typescript
@Get('/sample-route')
private sampleRoute() { // }
```
but I decided against it because I don't think many of the contributors working on chapter would understand them, and def not beginners/

allella:
This conversation is getting a little heated, so there are apparently strong feelings on both directions.

Are there code-related suggestions to meet somewhere in the middle on this one?

Haxified:
Theres only one person who's disagreed against this approach, vs several people who have agreed.

matjack1:
@allella I would have personally extracted the callback from `router.get('/api/sample-route', ...)` in a separate function.

So that it would be a bit easier to read and could grow well by extracting each function in its own file. Could be a compromise to avoid the objects while retaining the idea.

If I had to chose between what we have now and the suggested PR I would go with the PR. But still if you are intersted in my suggestion I could provide an example.

Haxified:
I had a chat with @timmyichen last night and he seemed on board with the idea

timmyichen:
I have this really bad habit of addressing things on discord and then forgetting to document it on GitHub.

I was mostly fine with it, but the solution offered by @matjack1 seems like a sensible compromise. Being able to see all the routes defined clearly in one place is a imo the biggest strength among these changes.

On the other hand, in TS we want to avoid declaring types unless necessary, doubly so in this project since we are going for a "typescript-lite" approach where contributors should write as little TS as possible, and manually typing every route handler seems counterproductive to that goal.

matjack1:
I'll send a stub later today then so that you can have a look, cannot right now :)

allella:
@matjack1 it sounds like we should explore your suggestion.

ThomasRoest:
some resources that might be interesting:
- https://dev.to/santypk4/bulletproof-node-js-project-architecture-4epf
- https://github.com/goldbergyoni/nodebestpractices

matjack1:
OK, I've opened https://github.com/freeCodeCamp/chapter/pull/134 :)

allella:
Closing as #134 achieved a consensus.

PR #134, opened 2019-10-23
Refactor: extract routes in folder

matjack1:

This is a followup PR to #128.

I'm proposing an intermediate version between the existing solution and what we have in #128.

We are not introducing objects but we are separating functions in files, so we retain a clean router and smaller route files.

I haven't any strong preference between this or #128, I like both :) Maybe this is just a bit simpler?

### DIFF
--- a/server/routers/exampleRouter.ts
+++ b/server/routers/exampleRouter.ts
@@ -1,26 +1,13 @@
 import express from 'express';
-import asyncHandler from 'express-async-handler';
-import { BadRequestError } from 'express-response-errors';
+
+import sampleRoute from 'server/routers/routes/sampleRoute';
+import sampleAsyncRoute from 'server/routers/routes/sampleAsyncRoute';
+import sampleErroredRoute from 'server/routers/routes/sampleErroredRoute';
 
 const router = express.Router();
 
-router.get('/api/sample-route', (req, res) => {
-  const { query } = req;
-  res.json({ message: 'Response from the server', query });
-});
-
-router.get(
-  '/api/sample-async-route',
-  asyncHandler(async (req, res) => {
-    await new Promise(resolve => setTimeout(resolve, 3000));
-
-    const { query } = req;
-    res.json({ message: 'Delayed response from the server', query });
-  }),
-);
-
-router.get('/api/sample-errored-route', () => {
-  throw new BadRequestError('This route will always error with code 400');
-});
+router.get('/api/sample-route', sampleRoute);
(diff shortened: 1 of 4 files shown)


allella:
@Haxified @timmyichen @slikts @ThomasRoest @Nekoniri @Zeko369 

Thoughts on this PR vs #128 vs the existing routes?

Nekoniri:
Seems like a decent compromise!

server/routers/routes/sampleRoute.ts
```diff
@@ -0,0 +1,4 @@
+export default function(req, res) {
```
timmyichen: this uses a regular function def while the other uses arrow functions - should we use the same for consistency?

timmyichen:
also LGTM, minus a minor consistency thing.  does this pass TS linting as well?  (we should add that to CI)

vaibhavsingh97:
@timmyichen Will send PR for the same

server/routers/routes/sampleRoute.ts
vaibhavsingh97: nice catch  👍 , @matjack1 can you please make changes?

server/routers/exampleRouter.ts
```diff
@@ -1,23 +1,13 @@
...
-    await new Promise(resolve => setTimeout(resolve, 3000));
-
-    const { query } = req;
-    res.json({ message: 'Delayed response from the server', query });
-  }),
-);
+router.get('/api/sample-route', sampleRoute);
+router.get('/api/sample-async-route', sampleAsyncRoute);
 
 router.get('/api/sample-errored-route', () => {
```
timmyichen: also, any reason why this also doesn't have its own separately defined handler? 🙃

allella:
@slikts are you good with this approach?

vaibhavsingh97:
@slikts, please review

slikts:
It's terse, doesn't use inheritance and doesn't add an unnecessary abstraction, so the previous objections don't apply.

matjack1:
Thank you all. I've updated the PR taking onboard the suggestions and the linting passes on my machine, I'll add it to CI in a separate PR if that's OK.

vaibhavsingh97:
@slikts and @timmyichen thanks for the review 👍

nik-john:
@all-contributors please add @matjack1   for code

nik-john:
@all-contributors please add @matjack1 for code

PR #140, opened 2019-10-24
feat(sponsors): add component for adding a sponsor to an event

### LINKED ISSUE #97: As an organizer I can add a venue sponsor to the event with a link to their website as a way of thanking them for hosting.

joelrozen:
#97

Closes #97

- [x] adds form to add a sponsor to an event.
- [ ] display sponsor on event listing page.

@vkweb - as discussed I started this before the change to TS and I don't know TS (yet!) :(


package.json
```diff
@@ -37,6 +37,7 @@
   },
   "homepage": "https://github.com/freeCodeCamp/chapter#readme",
   "dependencies": {
+    "axios": "^0.19.0",
```
Haxified: i would recommend using fetch

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <button type="submit">Add Sponsor</button>
+      </form>
+    </div>
+  );
+};
+
+AddSponsor.propTypes = {
+  eventId: string,
+  chapterId: string,
+};
```
Haxified: No need for prop types when your using typescript

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { useState } from 'react';
+import { string } from 'prop-types';
+import axios from 'axios';
+
+const AddSponsor = ({ eventId, chapterId }) => {
```
Haxified: type this

joelrozen: I have never used typescript, but have given it a go. I'm just not sure what type handlesubmit and handlechange are supposed to be?

client/components/AddSponsor.tsx
joelrozen: fixed

package.json
joelrozen: changed to use fetch

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+interface SponsorData {
+  name?: string;
+  website?: string;
+  type?: string;
+}
+
+const AddSponsor: React.FC<AddSponsorProps> = ({ eventId, chapterId }) => {
+  const [isSubmitting, setSubmitting] = React.useState<boolean>(false);
```
timmyichen: No need to specify types if they can be implied from the initial value, so this and `success` can just be `React.useState(false)`

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,86 @@
...
+    if (isSubmitting) {
+      return false;
+    }
+
+    e.preventDefault();
+    setSubmitting(true);
+
+    try {
+      const data = JSON.stringify(values);
+      await fetch(`/${chapterId}/events/${eventId}/sponsors`, {
```
timmyichen: add a TODO in a comment to indicate this route doesn't exist yet

pages/add-sponsor.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const AddSponsorPage = () => {
+  // TODO - Dynamically get Chapter and EventId s
+  const eventId = '123abc';
+  const chapterId = 'abc123';
+
+  return (
+    <div>
+      <AddSponsor eventId={eventId} chapterId={chapterId} />
+      <style jsx>{`
```
timmyichen: I believe we switched from using styled jsx to styled components

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,86 @@
...
+    } catch (error) {
+      // TODO: Error Handling
+    } finally {
+      setSubmitting(false);
+    }
+  }
+
+  return (
+    <div>
+      {success && <div>Sponser has been added!</div>}
```
timmyichen: `sponser` -> `sponsor`

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <form onSubmit={handleSubmit}>
+        <label>Sponsor Name: </label>
+        <input
+          name="name"
+          type="text"
+          placeholder="Glitter and Sparkle Co"
```
timmyichen: ✨

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+interface SponsorData {
+  name?: string;
+  website?: string;
+  type?: string;
+}
+
+const AddSponsor: React.FC<AddSponsorProps> = ({ eventId, chapterId }) => {
+  const [isSubmitting, setSubmitting] = React.useState<boolean>(false);
+  const [values, setValues] = React.useState<SponsorData>({ type: 'FOOD' });
```
timmyichen: for this i think it would be better to make the fields of `SponsorData` not optional, and instead pass in a default value of empty string.  this will let you do some validations on the values without having to check for null/undefined-ness.  e.g.

```
interface SponsorData {
  name: string;
  website: string;
  type: string;
}

...

React.useState<SponsorData>({ name: '', website: '', type: 'FOOD' })
```

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+interface AddSponsorProps {
+  eventId: string;
+  chapterId: string;
+}
+
+interface SponsorData {
+  name?: string;
+  website?: string;
+  type?: string;
```
timmyichen: since the possible values for this is known we should just make this an enum:

```
type: 'FOOD' | 'BEVERAGE' | 'OTHER';
```

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,86 @@
...
+          type="text"
+          placeholder="www.glitter.co"
+          onChange={handleChange}
+        />
+        <br />
+        <br />
+        <label>Sponsor Type: </label>
+        <select name="type" onChange={handleChange}>
+          <option value="FOOD">Food</option>
+          <option value="BEVEREGE">Beverege</option>
```
timmyichen: `Beverege` => `Beverage`

pages/add-sponsor.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { AddSponsor } from '../client/components';
```
timmyichen: you can just use `client/components` :)

timmyichen:
Btw let me know if you need any help with TS, either here or on Discord 🙃

timmyichen:
Oops sorry for the string of comments, but can you also revert the changes to `package-lock.json` since `axios` is no longer required? Thanks 😛

client/components/AddSponsor.tsx
timmyichen: what editor are you using - do they have typescript plugins?

joelrozen:
> Oops sorry for the string of comments, but can you also revert the changes to `package-lock.json` since `axios` is no longer required? Thanks 😛

Do you want me to just add package-lock.json and yarn.lock to .gitignore?

vaibhavsingh97:
>Do you want me to just add package-lock.json and yarn.lock to .gitignore?

Nope, it's needed in the project and it's recommended to present in git

nik-john:
We should be using either yarn or npm, there's no need the use both. I would recommend just using npm now that it's pretty much caught up with yarn

client/components/AddSponsor.tsx
vaibhavsingh97: +1 on this

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,86 @@
...
+      {success && <div>Sponser has been added!</div>}
+      <form onSubmit={handleSubmit}>
+        <label>Sponsor Name: </label>
+        <input
+          name="name"
+          type="text"
+          placeholder="Glitter and Sparkle Co"
+          onChange={handleChange}
+        />
+        <br />
```
vaibhavsingh97: can have component which helps in adding whitespace

pages/add-sponsor.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const AddSponsorPage = () => {
+  // TODO - Dynamically get Chapter and EventId s
+  const eventId = '123abc';
+  const chapterId = 'abc123';
+
+  return (
+    <div>
```
vaibhavsingh97: instead of wrapping component over `<div>` we can use `<> </>`. It's more of a react way of doing it

client/components/AddSponsor.tsx
timmyichen: or just use CSS 🙃

client/components/AddSponsor.tsx
joelrozen: @timmyichen I use vscode. I have the typescript hero extension but nothing else - any suggestions?

client/components/AddSponsor.tsx
joelrozen: changed. Though if we are allowing implicit types why use ts at all? (curiosity question)

client/components/AddSponsor.tsx
joelrozen: added

pages/add-sponsor.tsx
joelrozen: yes we did (praise!) however I just copy/pasted the existing index page :)

client/components/AddSponsor.tsx
joelrozen: agree, nicer ux :) changed

client/components/AddSponsor.tsx
joelrozen: true I didn't think of that, I kept them optional because they were initially not set, didn't think to set them to nothing haha. Have fixed.

client/components/AddSponsor.tsx
joelrozen: Agree and fixed. I originally did try and do by declaring an enum SponsorTypes and then trying to reference that but I couldn't get it to work.

client/components/AddSponsor.tsx
joelrozen: well that's embarassing

client/components/AddSponsor.tsx
joelrozen: I'll change up the form fields to use styled components :)

pages/add-sponsor.tsx
joelrozen: Yup I originally did this and my editor threw me a bunch of errors about something so I assumed it wasn't supported in TS... but now I tried again and it just... works. 
Sometimes I hate code.

client/components/AddSponsor.tsx
joelrozen: @Haxified - CI fails now saying eventId and chapterId are missing in props validation?

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,106 @@
...
+    if (isSubmitting) {
+      return false;
+    }
+
+    e.preventDefault();
+    setSubmitting(true);
+
+    try {
+      const data = JSON.stringify(values);
+      await fetch(`/${chapterId}/events/${eventId}/sponsors`, {
```
timmyichen: seeing `'fetch' is not defined` - can you install the npm package `isomorphic-fetch` so you can `import fetch from 'isomorphic-fetch`?  we're gonna need this anyway when doing more server-rendered stuff

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,106 @@
...
+const Input = styled.input`
+  margin-bottom: 15px;
+`;
+
+const ResponseDiv = styled.div`
+  margin: 15px 0;
+`;
+// TODO: Style div based on response reveived
+
+const AddSponsor: React.FC<AddSponsorProps> = ({ eventId, chapterId }) => {
```
timmyichen: regarding this i think you can just do `const AddSponsor = ({ eventId, chapterId }: AddSponsorProps) => {` but someone else might have opinions on whether using `React.FC` is better

vaibhavsingh97: I think `React.FC` is better as it automatically adds children for you, and check the return type

joelrozen: I have added a return type of JSX.Element to AddSponsor but still getting eventId and chapterId are not validated. I need to guidance on how to fix these please?

vaibhavsingh97: I am flying, but will help you out in evening 😅

timmyichen: Oh I just realized this was an eslint error, not a typescript error.  We should turn off the `react/prop-types` rule since we're not using `PropTypes`.  Can you add this to `eslintrc.json`?  Just a new line under `rules` that says something like `"react/prop-types": 0`.

Once done, you can also get rid of the return type you added, since it is implied by the JSX you're returning in the FC (if you're always returning the same thing TS will be able to infer the type)

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,113 @@
+import * as React from 'react';
+import styled from 'styled-components';
+import fetch from 'isomorphic-fetch';
+
+interface IAddSponsorProps {
```
vaibhavsingh97: can move to `interface` at one single place for a component? I mean for Sponsor component, we can have single place for all the `interface`

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,113 @@
...
+  chapterId: string;
+}
+
+interface ISponsorData {
+  name: string;
+  website: string;
+  type: 'FOOD' | 'BEVERAGE' | 'OTHER'; // TODO: Add VENUE
+}
+
+const Form = styled.form`
```
vaibhavsingh97: same as above, please move all styles in a single file for a component

joelrozen:
> LGTM, just a minor thing, move `AddSponsor.tsx` to `Components` folder in `interfaces` and `styles`.
> FYR:
> 
> ```
> |
> |__interfaces
> |   |__Components
> |      |_AddSponsor.tsx
> |__styles
> |   |__Components
> |      |_AddSponsor.tsx
> ```

done!

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,93 @@
...
+        setSubmitting(false);
+      } else {
+        setResponseMsg(
+          `${values.name} has been added as a ${values.type.toLowerCase} sponsor.`,
+        );
+        setSubmitting(false);
+      }
+    });
+
+    setSubmitting(false);
```
timmyichen: since the above call is async this will be invoked immediately.  we should probably be using async/await instead of `.then()`s

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,93 @@
...
+      body: data,
+    }).then(res => {
+      if (res.status != 200) {
+        setResponseMsg('Uh oh, something went wrong.'); // TODO: more descriptive error messages
+        setSubmitting(false);
+      } else {
+        setResponseMsg(
+          `${values.name} has been added as a ${values.type.toLowerCase} sponsor.`,
+        );
+        setSubmitting(false);
```
timmyichen: since `setSubmitting(false)` is being called regardless of the condition we can pull this out of the if/else statement (that way we'd only need it once)

client/components/AddSponsor.tsx
timmyichen: using async await usually looks something like this:
```
try {
  async doAsyncThing()
} catch (e) {
  // e.message has the error message
}

File: client/components/AddSponsor.tsx [hunk 21]
```diff
@@ -0,0 +1,93 @@
...
+const AddSponsor: React.FC<IAddSponsorProps> = ({ eventId, chapterId }) => {
+  const [isSubmitting, setSubmitting] = React.useState(false);
+  // we specify type as food as state is only updated on field change. Since field
+  // defaults to food and the user doesn't change it, state would be empty
+  const [values, setValues] = React.useState<ISponsorData>({
+    name: '',
+    website: '',
+    type: 'FOOD',
+  });
+  const [responseMsg, setResponseMsg] = React.useState<string>();
```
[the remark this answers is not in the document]
joelrozen: @timmyichen see the comment ln 17 - 18. The 'type' is only updated in state if the form field is changed. If we set the initial state to an empty string we face the situation where the user leaves "food" selected (ie doesn't change it) in the drop down, and then we would be submitting an empty string to the backend for the type.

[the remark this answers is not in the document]
timmyichen: Ah this was referencing `responseMsg` - sorry for the late reply

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,92 @@
...
+import {
+  IAddSponsorProps,
+  ISponsorData,
+} from 'client/interfaces/components/AddSponsor';
+
+const AddSponsor: React.FC<IAddSponsorProps> = ({ eventId, chapterId }) => {
+  const [isSubmitting, setSubmitting] = React.useState(false);
+  // we specify type as food as state is only updated on field change. Since field
+  // defaults to food and the user doesn't change it, state would be empty
+  const [values, setValues] = React.useState<ISponsorData>({
```
Zeko369: Have we discussed using `useReducer` instead of `useState` with an object for this? For this specific form it's OK, but if we had more fields maybe `useReducer` hook would make it more readable, since we're just implementing our own reducer later in the `handleChange`?

client/interfaces/components/AddSponsor.tsx
```diff
@@ -0,0 +1,10 @@
+export interface IAddSponsorProps {
+  eventId: string;
+  chapterId: string;
+}
+
+export interface ISponsorData {
+  name: string;
+  website: string;
+  type: 'FOOD' | 'BEVERAGE' | 'OTHER'; // TODO: Add VENUE
```
Zeko369: Maybe better create enum for this since it's going to be used more than once?

Zeko369:
```ts
enum SponsorDataType {FOOD, BEVERAGE, OTHER}

interface ISponsorData {
  name: string;
  type: SponsorDataType;
}

const sponsorData: ISponsorData = {
  name: 'Foobar',
  type: SponsorDataType.FOOD
}
```

client/components/AddSponsor.tsx
Zeko369: I know that we need `isomorphic-fetch` for `getInitalProps`, because that is done on the server side, but this will always be done on the client, so do we need `isomorphic-fetch` here or should the normal `fetch` be fine for this?

client/components/AddSponsor.tsx
timmyichen: I don't think this has been discussed as a pattern we want to use - feel free to create a new issue for it!

client/components/AddSponsor.tsx
timmyichen: Not needed, although we _could_ use isomorphic fetch here for consistency.  Shouldn't matter either way

QuincyLarson:
Note: Per https://github.com/freeCodeCamp/chapter/issues/163 we have decided to use React-hook-form.

joelrozen:
I will refactor tomorrow

joelrozen:
@timmyichen and @QuincyLarson changes complete. Also now makes (basic) use of react-hooks-form as per #163

client/components/AddSponsor.tsx
```diff
@@ -0,0 +1,73 @@
...
+} from 'client/interfaces/components/AddSponsor';
+
+const AddSponsor: React.FC<IAddSponsorProps> = ({ eventId, chapterId }) => {
+  const [responseMsg, setResponseMsg] = React.useState('');
+
+  const { register, handleSubmit, errors } = useForm();
+
+  const onSubmit = async data => {
+    const { name, website, type }: ISponsorData = data;
+    console.log({ name, website, type });
```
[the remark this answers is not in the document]
joelrozen: Ahhh oops

vaibhavsingh97:
Merging, Thanks for the PR @joelrozen 💯

Shub1427:
So are we going with `isomorphic-fetch` then?? It hasn't been maintained for 4 years now
Ref: https://github.com/matthew-andrews/isomorphic-fetch/issues/133

vaibhavsingh97:
@Shub1427 that's not the case here, please feel free to raise another PR if you want to use axios. This PR is there for long time, so i thought let's merge this and then work actively on fixing issues with PR.

PR #146, opened 2019-10-26
add email service with nodemailer

eolculnamo2:

Creates Email Service using nodemailer that will be used later for our email tickets.

Dependencies Added:
1) nodemailer
2) chai
3) sinon
4) sinon-chai.


package.json
```diff
@@ -18,7 +18,7 @@
     "speccy:watch": "npx nodemon --watch 'api/swagger.json' --exec 'npm run speccy'",
     "speccy": "npx speccy serve -p 8001 api/swagger.json",
     "start": "npm run prod:server",
-    "test": "NODE_ENV=test npx jest --coverage --verbose",
```
Haxified: it seems you added mocha and chai without documenting it in the pr?

server/services/MailerService.ts
```diff
@@ -0,0 +1,69 @@
+import nodemailer, { Transporter, SentMessageInfo } from 'nodemailer';
+import { IMailerService } from 'types/services';
+import Utilities from 'server/util/Utilities';
+
+// @todo add ourEmail, emailUsername, emailPassword, and emailService as
+// environment variables when they become available. Temporary placeholders
+// provided until updated info available.
+export default class MailerService implements IMailerService {
```
Haxified: why do you need to implement IMailerService? its not like there are gonna be multiple class definitions following the same guidelines

package.json
eolculnamo2: You're absolutely right. I should have mentioned it in the PR. I'm new to contributing to online projects outside of work -- I'll get that updated.

server/services/MailerService.ts
eolculnamo2: That's true, but the purpose of interfaces is not only to define shared types, but also to show a high level abstraction of a class. In most Object Oriented projects (such as C# and Java), it's generally considered a good practice to include the interface on services and I think it makes sense for NodeJS w/ TS, but I guess its really up to the FCC community. If you all believe I should get rid of it, I certainly will.

server/services/MailerService.ts
```diff
@@ -0,0 +1,75 @@
...
+    htmlEmail: string,
+    backupText?: string,
+  ) {
+    this.emailList = emailList;
+    this.subject = subject;
+    this.htmlEmail = htmlEmail;
+    this.backupText = backupText || '';
+
+    // to be replaced with env vars
+    this.ourEmail = 'ourEmail@placeholder.place';
```
vaibhavsingh97: can you please replace this placeholder with env, and define this in `env.sample`.

vaibhavsingh97: Also, please update the readme, to copy env from `env.sample`

eolculnamo2: Added :)

README.md
```diff
@@ -80,6 +80,9 @@ Install dependencies:
 npm install
 ```
 
+Setup Environment Variables:
+Copy and paste env.sample into your .env file
```
ScottBrenner: How about using [a ".env" file in Docker Compose](https://docs.docker.com/compose/environment-variables/#the-env-file)?

eolculnamo2: I honestly don't know much at all about Docker -- Do you mean that we should use environment variables in Docker Compose for default variables that can be overwritten by an .env file?

allella:
@vaibhavsingh97 @eolculnamo2 would we need to enable all the transports, by default, to support the common options?

Looks like SMTP is the default and then SES and sendmail (local relay) need to be enabled.

So, would be be installing everything even if people only will use one?

jacobbogers:
Hi, I can look into this

jacobbogers:
Hi fellas, I need to dive into `node-mailer` ( I am only familiar with `mailgun-js`  ), for me to make any sensible decision about design.. I will take this opportunity to learn node-mailer very well,

vaibhavsingh97:
> would we need to enable all the transports, by default, to support the common options?

Nope, by default only SMTP will be enabled, we will just provide options to enable all other providers.

> Would we be installing everything even if people only will use one transport?

Nope, We will give options to users, if they want to use other services.

allella:
> Nope, by default only SMTP will be enabled, we will just provide options to enable all other providers

How would installing new mailer packages work for the app? Would the administrator need to run a command on the server, or is there a way to install new packages directly from the application admin UI and/or configuration files?

jacobbogers:
@allella  nodemailer uses plugins for non SMTP transports (has some build in),  to avoid manually installing them, they need to already exist as a preset (needs its own config page to fill in params) to be selected, I am looking into that as well to put in MVP, it seems to be very trivial.

vaibhavsingh97:
@eolculnamo2 Can you please resolve conflicts so that we can merge 😄

eolculnamo2:
Will do :)

vaibhavsingh97:
Also, Can you please resolve @ScottBrenner comment, also as you had added `mocha` and `chai` for test frameworks, so can you please add the same in the readme [here](https://github.com/freeCodeCamp/chapter#testing) and [here](https://github.com/freeCodeCamp/chapter#tech-stack).

allella:
We had more current conversation in [today's meeting on using nodemailer and SMTP with free / cheap tiers of 3rd party transactional mailing services](https://github.com/freeCodeCamp/chapter/wiki/Meeting-Notes-February-23,-2020#email-service).

Issue #144, opened 2019-10-24
Browser Support Matrix


nik-john:
## Browser Support Matrix

We should define a Browser Suppor Matrix like @ceciliaconsta3 mentioned [here](https://github.com/freeCodeCamp/chapter/issues/114#issuecomment-545878412). For the MVP I personally think it's good enough to support latest - 2 Chrome (desktop, android), Firefox, Safari (Desktop, iOS), IE10+ & Edge.

However, because this is going to be an totally public facing app allowing us no control over what Geography, Device or Networks it will be used in, we will need to support pretty much all possible scenarios including Opera Mini and UC Browser. Even if we don't develop for these two, the app should work on them and we should definitely follow Progressive Enhancement.

The steps involved are:

### Define the Audience

- Geography: I'm assuming here that we plan to eventually support **all geographies**, the only requirement being that you have a working internet connection and some smart phone. 

- Devices (Form Factors): Again, here I assume we will support all Android, iOS and Desktop devices of any form factor, with a min. device width of ~200px.

#### Worldwide browser usage

![image](https://user-images.githubusercontent.com/1117182/67529577-66538e00-f671-11e9-97ee-5000f8e1a653.png)

![image](https://user-images.githubusercontent.com/1117182/67529600-779c9a80-f671-11e9-9c05-ff9373af5d9e.png)

<img width="553" alt="Screen Shot 2019-10-24 at 3 23 25 PM" src="https://user-images.githubusercontent.com/1117182/67529925-6dc76700-f672-11e9-8c6c-a0aedd3120fc.png">
<img width="620" alt="Screen Shot 2019-10-24 at 3 23 37 PM" src="https://user-images.githubusercontent.com/1117182/67529926-6e5ffd80-f672-11e9-8d0e-82eb087bb533.png">
<img width="570" alt="Screen Shot 2019-10-24 at 3 23 50 PM" src="https://user-images.githubusercontent.com/1117182/67529927-6e5ffd80-f672-11e9-8e84-f8b74b2cf71c.png">
<img width="363" alt="Screen Shot 2019-10-24 at 3 24 03 PM" src="https://user-images.githubusercontent.com/1117182/67529928-6e5ffd80-f672-11e9-82ad-4a9bfa45dbe6.png">
[...]

Issue #47, opened 2019-10-16
Our Roadmap


kognise:
Below I'll try to assemble a checklist of things we have to get done before shipping a semi-stable beta that's close to a finished product.

As I see it so far:

- [x] Sort Out Additional [Tech Stack](https://github.com/freeCodeCamp/chapter#tech-stack) Decisions ~~#24 #34 #145 #76 #32~~
- [x] Determine Web Browser Support ~~#144~~
- [x] Sort out Authentication ~~#40 #33~~
- [ ] Finalize an API ~~#17~~ ~~#46~~ #257 #120 #253 
- [x] UI/UX mockups #114
- [ ] Add Initial Documentation to Wiki #12
- [ ] Build super basic MVP based on the [MVP Project Board](https://github.com/freeCodeCamp/chapter/projects/1)

This is vastly simplified, maybe we eventually want to move to a task management solution? Anyways, I'd love feedback on what this is missing. I tried to link up some issues/PRs but I'm probably missing a lot.

chrismgonzalez:
@kognise This looks awesome and gives us some direction over the coming days.  I don't mind chiming in on API discussions, but I have been focusing on basic docs right now - I'm happy to be CC'd on that topic.  I don't have any experience developing API's, though I am learning.

allella:
@kognise I think Chris is more working on the 3rd bullet, "Build a list of contributors and make the project more welcoming"

martindavid:
Love this checklist. I'm super keen to help, but minimum experience on working with OSS project. So I'm looking forward to see how people can contribute in this initial stage

allella:
@kognise I'll try to supplement what @chrismgonzalez is doing by summarizing who's contributing on the various issues and seeing if we can tap into the [couple hundred volunteers sitting on the sidelines waiting for some direction](https://github.com/freeCodeCamp/chapter/issues/11#issuecomment-542447016
).

SeanFromIT:
Do we have some way of prioritizing the backlog? Project Board maybe?

kognise:
@SeanFromIT I think we could consider using either GitHub Projects or Trello - @QuincyLarson what are your thoughts on this?

martindavid:
Really like if we can use Github Projects so that we can have one source to go. I found Trello and Github Projects not have much differences, unless you're using the Trello Powerup

MirzaChilman:
I agree with @martindavid , it's better to make it centralized

allella:
[...]

PR #148, opened 2019-10-26
add globalstyles and minimal css reset

ThomasRoest:

Adds a globalstyles component with a minimal reset. Any other ideas/improvements welcome ofcourse!

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -2,6 +2,7 @@ import App from 'next/app';
 import React from 'react';
 import { ThemeProvider } from 'styled-components';
 import theme from 'styles/theme';
+import { GlobalStyles } from 'styles/global';
 
 export default class MyApp extends App {
   render() {
@@ -9,6 +10,7 @@ export default class MyApp extends App {
     return (
       <ThemeProvider theme={theme}>
         <Component {...pageProps} />
+        <GlobalStyles />
       </ThemeProvider>
     );
   }
--- a/styles/global.ts
+++ b/styles/global.ts
@@ -1,0 +1,50 @@
+import { createGlobalStyle } from 'styled-components';
+
+export const GlobalStyles = createGlobalStyle`
+  html {
+    box-sizing: border-box;
+  }
+
+  *,
+  *::before,
+  *::after {
+    box-sizing: inherit;
+  }
+
+  body,
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6,
+  ul,
+  ol,
+  li,
+  p,
+  pre,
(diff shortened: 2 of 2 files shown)


pages/_app.tsx
```diff
@@ -2,13 +2,15 @@ import App from 'next/app';
...
 import theme from 'styles/theme';
+import { GlobalStyles } from 'styles/global';
 
 export default class MyApp extends App {
   render() {
     const { Component, pageProps } = this.props;
     return (
       <ThemeProvider theme={theme}>
         <Component {...pageProps} />
+        <GlobalStyles />
```
vaibhavsingh97:
```suggestion
        <GlobalStyles />
        <Component {...pageProps} />
```

styles/global.ts
```diff
@@ -0,0 +1,50 @@
...
+    color: inherit;
+    font: inherit;
+    letter-spacing: inherit;
+  }
+
+  embed,
+  iframe,
+  img,
+  object,
+  video {
```
vaibhavsingh97: Also, add the following tags:

```html	
article,
aside,
details,
figcaption,
figure,
footer,
header,
hgroup,
nav,
section,
summary
```

ThomasRoest: what do you mean by adding these elements? Do they need some reset in some form?

ThomasRoest: not sure about IE6/7/8/9. I don't think we should bother supporting those browsers

allella: We have a browser support conversation on #144.

vaibhavsingh97: > what do you mean by adding these elements? Do they need some reset in some form?

Yes, I meant adding reset rules for the mentioned tags

ThomasRoest: That's not necessary

vkWeb:
@ThomasRoest We are going to use Material-UI so we don't need to add reset styles explicitly. `<CssBaseline />` will do the job for us: https://material-ui.com/components/css-baseline/

vaibhavsingh97:
> @ThomasRoest We are going to use Material-UI so we don't need to add reset styles explicitly. `<CssBaseline />` will do the job for us: [material-ui.com/components/css-baseline](https://material-ui.com/components/css-baseline/)

I don't know if I am missing anything, but have we reached the conclusion that which UI framework we will use?

vkWeb:
@vaibhavsingh97 Quincy finalized it: https://github.com/freeCodeCamp/chapter/issues/47#issuecomment-547175059

vaibhavsingh97:
Awesome!! thanks @vkWeb 👍

PR #158, opened 2019-10-31
Nik john/add material ui ssr

nik-john:

Material UI stylesheet also need to be rendered on the server side. This PR adds this change and adds a sample component as per [Official Material UI Docs](https://github.com/mui-org/material-ui/tree/master/examples/nextjs)


pages/index.tsx
```diff
@@ -1,16 +1,9 @@
-import * as React from 'react';
+import React from 'react';
 import SomeComponent from 'client/components/SomeComponent';
-import styled from 'styled-components';
-
-const Title = styled.h1`
-  font-size: 50px;
-  color: ${({ theme }) => theme.colors.primary};
-`;
```
timmyichen: can we add back in some example of using styled components for future contributors?

ThomasRoest:
thoughts about folder structure? My preference would be

```
MyComponent
      --- index.tsx
      --- styles.ts
```

pages/index.tsx
nik-john: @timmyichen Yup, done.

nik-john:
@ThomasRoest Agreed. Not sure why we didn't do that already, but I've made the change

PR #162, opened 2019-11-09
Add TypeORM models

eolculnamo2:

Adds sequelize models as show in the db diagram. Potential improvement would be to use HasMany relationships over simply referencing the Id in some case.

### DIFF
--- a/server/migrations/20191117185956-add-initial-tables.js
+++ b/server/migrations/20191117185956-add-initial-tables.js
@@ -1,0 +1,23 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    /*
+      Add altering commands here.
+      Return a promise to correctly handle asynchronicity.
+
+      Example:
+      return queryInterface.createTable('users', { id: Sequelize.INTEGER });
+    */
+  },
+
+  down: (queryInterface, Sequelize) => {
+    /*
+      Add reverting commands here.
+      Return a promise to correctly handle asynchronicity.
+
+      Example:
+      return queryInterface.dropTable('users');
+    */
+  },
+};
--- a/server/models/Chapter.ts
+++ b/server/models/Chapter.ts
@@ -6,8 +6,16 @@ import {
   UpdatedAt,
   DataType,
   PrimaryKey,
+  HasOne,
+  HasMany,
+  BelongsToMany,
 } from 'sequelize-typescript';
-import { IChapter } from 'types/models';
+import { IChapter, ILocation, IUser, IUserBan, IEvent } from 'types/models';
+import { Location } from './Location';
+import { User } from './User';
+import { UserBan } from './UserBan';
+import { Event } from './Event';
+import { UserChapter } from './UserChapter';
 
 @Table
 export class Chapter extends Model<IChapter> {
@@ -28,7 +36,19 @@ export class Chapter extends Model<IChapter> {
   details!: any;
 
   @Column
-  creatorId!: number;
+  creator_id!: number;
+
+  @HasMany(() => Event)
+  events!: IEvent[];
+
+  @HasOne(() => Location)
+  location!: ILocation;
+
+  @BelongsToMany(() => User, () => UserChapter)
+  users!: IUser[];
+
+  @HasMany(() => UserBan)
+  user_bans!: IUserBan[];
 
   @CreatedAt
   @Column
--- a/server/models/Event.ts
+++ b/server/models/Event.ts
@@ -5,8 +5,19 @@ import {
   Table,
   UpdatedAt,
   PrimaryKey,
+  HasMany,
+  HasOne,
+  BelongsToMany,
+  ForeignKey,
+  BelongsTo,
 } from 'sequelize-typescript';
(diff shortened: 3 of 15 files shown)


Zeko369:
You should defiantly add realtions on the ORM level

eolculnamo2:
That’s what I thought. Will add.

timmyichen:
thanks for adding these!  just a quick note, model fields should be `snake_case` as is standard for SQL/postgres.  additionally, each new model will need a corresponding [migration file](https://sequelize.org/master/manual/migrations.html#migration-skeleton) - you could do all of them in one file, or split them up into many - the latter is probably preferred

eolculnamo2:
Updated relationships and casing.

Tag was missing event_id which is required for the one to many relationship. I've added that to the class.

Also, I'm not 100% sure I understand the one-to-one relationships from sequelize-typescript (If someone could double check) :)

allella:
@eolculnamo2 the schema was showing only one tag per event, which seems like an oversight.

Is Discord, I just asked "The [mock-ups](https://www.figma.com/proto/q7DikyL3N0c4CUWxHNa97i/Chapter-Prototype?node-id=1%3A2&scaling=scale-down) already show multiple tags per event and it seems a very safe assumption 1 tag per event is not enough.

That said, do we need an [MVP user story](https://github.com/freeCodeCamp/chapter/issues?q=is%3Aissue+is%3Aopen+label%3AMVP) that says "As an Admin, I can apply optional tags to an event"? Or, do we think this isn't a necessary MVP story. If not, then should we still building an event_tags table now to show the future intention of multiple tags?"

eolculnamo2:
The event tags decision is probably not my call, but it seems simple enough to include with the MVP and its already in our mock ups.

Is there a way to automatically build the migrations file from what I already have or does that need to be manually typed? (I ask because it reminds me of C#'s Entity Framework which build it for you)

timmyichen:
As far as I know there is no way to automatically generate migrations file.  If we were using non-typescript sequelize it would be fairly straightforward as the migration file would look almost identical to the model declaration object, but since we are using typescript-sequelize, the model declarations are classes.

Zeko369:
@timmyichen @eolculnamo2 If we used `TypeORM`, which has pretty much identical API to `sequelize-typescript` we could have auto generated migrations based on model files. We haven't discussed that since `sequezlie` was in the initial stack

vkWeb:
I agree with @allella that we need multiple tags per event and it should be in MVP. We can control the maximum number of tags organizers can apply client side (or server-side). 

I am updating the database, er diagram and api schemas as well. I'll send a PR soon. 

<hr>

@timmyichen, @Zeko369, @nik-john, @allella I was looking at the api routes and I realized that the routes are revolving around `chapters`. What if an organization doesn't have a chapter? 
Just, for example, assume we have `Coders.org` as an organization and they don't want any chapters under them, they host events under the organization's name. In that case, our routes will return 404, right? I think we should include /events, /events/eventId and so on as routes. What do you all think?

vkWeb:
Blocking this till we resolve the architectural issues on #251.

timmyichen:
@Zeko369 I'm down to use TypeORM but we should loop @QuincyLarson into this conversation as I believe it was he who initially proposed sequelize.

Zeko369:
Blocked because of #252

Zeko369:
We're switching to `TypeORM`, but since `sequelize-ts` syntax is pretty much `TypeORM` syntax, this PR will need little changes to make it work with `TypeORM`.
**DONT DO MIGRATATIONS TILL THEN**, beacuse we'll use TypeORM generate migrations, so you dont need to write raw SQL, just check if TypeORM was right :nerd_face: 

So this is blocked till #257 is merged

eolculnamo2:
Awesome, will get this done soon! I'll shoot to have it done sometime tomorrow or the next day.

eolculnamo2:
Things have come up lately that have made it hard to finish this in a timely manner :/ 
I've been working on this issue, but if someone who knows TypeORM already wants to knock it out, let me know.

Zeko369:
@eolculnamo2 When I finish TypeORM migration PR #257 I'll take over this

ankorGH:
@Zeko369 , if you are not working on it I would like to take a go at it.

Zeko369:
Today I'll finish the dev docs on TypeORM and how we're using it so I'll tag you in the pr, and then you can finish this migration @ankorGH

ankorGH:
sure

vaibhavsingh97:
@ankorGH you can now work on this PR, since https://github.com/freeCodeCamp/chapter/pull/257 is merged.

Zeko369:
@ankorGH If you don't have enough time, I can do this tomorrow

ankorGH:
@Zeko369 I am on it.

ankorGH:
@Zeko369 @eolculnamo2 can you grant me access to make commits to this PR or I should create a new PR

Zeko369:
@ankorGH I think that for the sake of simplicity and to get it done sooner it would be better to just create a new PR from your fork

Zeko369:
Closing and moving to #287

Issue #123, opened 2019-10-22
React Class Components vs Functional Components w/hooks


timmyichen:
Title says it all. What's our opinion on using class components vs functional components with hooks?

chrismgonzalez:
I'm down for Hooks since that is the latest and greatest.  Plus it will save us a refactor later.

timmyichen:
I'm also down for hooks, but I'm wondering if it will be yet another barrier to entry given that most resources online are written using class components.

chrismgonzalez:
In that case I would just point to React docs.  They're going to have to learn anyways, right?

timmyichen:
I suppose, but the less things people have to learn, the more people will contribute 🙃

chrismgonzalez:
Let's see what others think over the next 24 hours?  I am interested to hear the opinions of others.

Abott1222:
I think say we use fc + hooks and also be considerate in thinking about giving low hanging fruit as starter issues - that might lead to them understanding hooks better

imVinayPandya:
I recommend Functional Component + hooks. As @chrismgonzalez mentioned that it will save time for refactoring.

abidRahim:
Functional Components + Hooks. 
@timmyichen Converting Class components to Hooks won't be much of a problem, besides there are VS Code extensions to do the same (in worst case scenario).

rohanfaiyazkhan:
Functional components + hooks 100%.

Hooks will allow us to abstract out complex bits of logic which if anything lowers the barrier to entry since newcomers won't have to deal with them.

Danex2:
Hooks 100% of the way, I think they'll be a lot easier to pick up than most people think.

vardhanapoorv:
I think hooks is a good way to go with!

timmyichen:
Sounds like it's pretty unanimous! :+1: Lets go hooks! ↩️

mohanwer:
Hooks 👍

xarielx:
Functional Component + hooks, GitHub should add a polling button,  is there a pull request to update the documentation regarding this? @timmyichen

timmyichen:
Negative - feel free to create one!

xarielx:
Should we install this to facilitate future discussion of this kind? https://github.com/apps/polls @timmyichen @chrismgonzalez

QuincyLarson:
/polls "React Class Components" "Functional Components w/hooks"

QuincyLarson:
[...]

PR #166, opened 2019-11-12
Add Redux/Thunk with bare minimum setup

### LINKED ISSUE #145: Decide which state management tool to use
Hey there,

A state management tool is still to be defined and added to the [tech-stack](https://github.com/freeCodeCamp/chapter#tech-stack). 
Continuing the discussion from Discord, a few options we have: [Redux ](https://redux.js.org/), [MobX ](https://mobx.js.org/README.html), use of [ContextAPI](https://reactjs.org/docs/context.html), ... Please share your view and list any other interesting option that may exist.

I personally didn't work with MobX and use ContextAPI as a substitute to Redux doesn't seem a right choice to me. So it would be Redux for me. 

What tool would you go for and why?

Shub1427:

## Details on PR
Closes #145

This PR adds a bare minimum setup for redux/thunk for Chapter app. Fetch calls made is the PR is just for demo purpose and will be replaced with proper modals, once either a mock server is integrated or the api server is merged to the repo.

* Separates `fetch` proxy lib, in server/browser, with different packages(which are actively maintained).
* Using Redux Hooks for store state and dispatching actions from the component itself.

## TODO:

* Add `redux-persist` for data persistence.
  - The only issue is SSR support from `redux-persist`, https://github.com/rt2zz/redux-persist/issues/457
* Add SSR support to cache and re-hydrate redux state in react app
* Typescript integration with Redux is still very hacky, any suggestions or later RFCs might resolve them.

> Some changes:
> * I have added `modules` folder to keep page related layout components in them, all the UI specific generic components are put in `components` dir.
> * Also, I have kept Http Service in a separate layer, so that if needed, we can replace it to whatever API service we want to use. Specific http services (eg, Chapter Service) should be derived or be composed of HttpService.

> PS: Also, this is my first ever biggest PR for a community project, if there are improvements needed to be done in the PR, please let me know, I would be happy to improve myself.


client/containers/home.tsx
```diff
@@ -0,0 +1,40 @@
+import Home from 'client/modules/home';
```
Zeko369: I suggest first importing installed modules, then local files

client/hoc/with-store.tsx
```diff
@@ -0,0 +1,27 @@
...
+    window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__) ||
+  compose;
+
+const store = createStore(
+  rootReducer,
+  composeEnhancers(applyMiddleware(thunk)),
+);
+
+export function withStore(Component: React.ComponentType) {
+  return class extends React.Component {
```
Zeko369: Can this be a plain function component?

package.json
```diff
@@ -45,13 +45,18 @@
     "express-async-handler": "^1.1.4",
     "express-response-errors": "^1.0.4",
     "fork-ts-checker-webpack-plugin": "^1.5.1",
+    "immer": "^3.2.0",
+    "isomorphic-unfetch": "^3.0.0",
```
Zeko369: We have `isomorphic-fetch` in PR #140 , can you explain why we should use `unfetch` instead of `fetch`?

Shub1427: The main reason to use `isomorphic-unfetch` is for it's small bundle size. But in my opinion, if we are ok to use bigger bundled libs, `axios` would be a preferred option, in my opinion.

Shub1427:
@Zeko369 Using hooks in react needs to be done carefully, and I am no master of it yet. I have come across various tweets/articles mentioning bottlenecks of hooks (which can be resolved, obviously, but with extra care). Check this repo - https://github.com/ryardley/hooks-perf-issues

I am not against of using hooks, but I would prefer using `PureComponent` with redux for page modules and functional components (with hooks, if needed) for UI components.

package.json
Zeko369: I think we should discuss this further, but I think that bundle size isnt that big of a problem for now

Zeko369:
@Shub1427 I think that we should discuss `hooks`/`React.PureComponents`/`React.Component` on the team meeting. @QuincyLarson Could you add this to the agenda?

QuincyLarson:
A couple notes: on https://github.com/freeCodeCamp/chapter/issues/123 we resolved to primarily use hooks. We've also agreed to use Thunk.

allella:
Meeting Notes with more on the decision to use hooks.
https://github.com/freeCodeCamp/chapter/wiki/Meeting-Notes-November-15,-2019

Shub1427:
Thanks all. Will update my PR accordingly.

To keep this in context, this PR is just a setup and does not include, data persistence, neither will it include any SSR support(I was having trouble setting it up with nextjs, but will figure it out). Will open a separate PRs supporting each/both of them.

package.json
Shub1427: I have updated the PR to use `node-fetch` in server and `github/fetch` in browser, as both these packages are actively maintained and quite stable.

This change is on behalf of the discussion @Sonicrida and I had, with approval of few other channel members.

client/services/http-service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { RequestInfo, RequestInit, Response } from 'node-fetch';
+
+let fetch: (url: RequestInfo, init?: RequestInit) => Promise<Response>;
+
+if (typeof window === 'undefined') {
+  fetch = require('node-fetch');
+} else {
+  fetch = window.fetch ? window.fetch : require('whatwg-fetch');
+}
```
Shub1427: I am not able to get any better/easier way to conditionally load `fetch` proxies. If anyone know about some custom webpack configs that we can use to make this code even more simpler, it would be great.

There is [this PR](https://github.com/zeit/next.js/pull/5159), which looks like a sugarcoat over `typeof window === 'undefined'`. So it doesn't seem to solve my problem.

client/store/reducers/chapter.ts
```diff
@@ -0,0 +1,39 @@
...
+  category: '',
+  loading: false,
+  error: '',
+};
+
+const reducer = (
+  state = initialState,
+  action: chapterTypes.IChapterActionTypes,
+) =>
+  produce(state, draft => {
```
Zeko369: Instead of using immer, why not just.
```
const reducer = (state = initialState, action: IChapterActionTypes) => {
  switch (action.type) {
    case FETCH_START:
      return {
        ...state,
        error: '',
        loading: true
      }
    case FETCH_SUCCESS:
      return {
        ...state,
        ...action.payload,
        error: '',
        loading: false
      }
    case FETCH_FAIL:
      return {
        ...state,
        error: action.error,
        loading: false
      }
    default:
      return state;
  }
}
```

client/services/http-service.ts
Zeko369: Why did we go with this instead of using `isomorphic-fetch` or something else that is plug and play. If I remember correctly it was because it "wasn't maintained", honestly why would you need to maintain `axios` or `isomorphic-fetch`?

This seems like a really hacky solution to me :confused:

pages/_app.tsx
```diff
@@ -1,14 +1,29 @@
 import App from 'next/app';
 import React from 'react';
+import { Provider } from 'react-redux';
+import { applyMiddleware, compose, createStore } from 'redux';
+import thunk from 'redux-thunk';
 import theme, { themeObject } from 'styles/theme';
 import { ThemeProvider as MaterialUIThemeProvider } from '@material-ui/core/styles';
```
Zeko369: This was done by @nik-john but just to be consistent could you please order imports since you're already working on this file?
1. npm libs (`react`, `next`, `@material-ui`)
2. our packages (`styles`, `components`)
3. local depenedcies (none in this case but `./someRandomComponent`)

I'll add this to the docs later

client/services/http-service.ts
Shub1427: @Sonicrida shared some stats on vulnerability issues found in `axios`, which I recently found that was a false alarm from `snyk` tests ([Details here](https://github.com/axios/axios/issues/2447)). I am ok with using anyone of these, but there has been conflicts between other members. I am thinking to open this as an issue and resolve it for all (removing this from blocker).

Shub1427: Discussions also concluded that, non maintained libs have many essential issues opened for a long time and haven't been looked into, where `isomorphic-fetch` hasn't been maintained for last 4 years.

client/store/reducers/chapter.ts
Shub1427: Vanilla JS destructuring works only with one level nested JS Objects, and can get pretty complicated to maintain when API responses get a little bit complicated (using nested JSON Objects). This is the biggest `WHY` that I like to use `immerjs`. More benefits can be found [here](https://immerjs.github.io/immer/docs/introduction#benefits).


__Why not `ImmutableJS`__

ImmutableJS uses it's own set of data structures, which for new comers or even with core JS devs might become a barrier to contribute. Where as `ImmerJS` keeps everything similar to pure js, but also giving us benefits of Object freezing, easy immutability with nested objects and more.

pages/_app.tsx
Shub1427: I haven't gone through that part of documentation, but will surely update my branch accordingly.

client/services/http-service.ts
nik-john: https://github.com/lquixada/cross-fetch is a well maintained alternative

client/store/reducers/chapter.ts
nik-john: How would one enforce immutability of state via. native JS without tons of boilerplate?  @Zeko369

Zeko369: @Shub1427 I agree with you that this is better for more complex stores

Shub1427: Will resolve this @Zeko369

pages/_app.tsx
Shub1427: Fixed it. :100:

client/services/http-service.ts
Shub1427: Resolving this as per decisions made - https://github.com/freeCodeCamp/chapter/issues/258

package.json
```diff
@@ -94,6 +99,7 @@
   "lint-staged": {
     "*.{ts,tsx,js,jsx}": [
       "npm run pretty",
+      "npm run lint",
```
Shub1427: CI tests make sure lint errors do not pass. Sometimes people (like me) forget to check linting errors :sweat_smile: . Thus, it seems like a great idea to make this check as a precommit hook. :100:

Shub1427: cc: @timmyichen

Zeko369: YESSSSSSSSSSSSSSSSSSSSSS

client/interfaces/components/AddSponsor.tsx
```diff
@@ -1,7 +1,4 @@
-export interface IAddSponsorProps {
-  eventId: string;
-  chapterId: string;
-}
+// export interface IAddSponsorProps {}
```
Zeko369: Maybe better to collocate this with the component

client/modules/home.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { AppStoreState } from 'client/store/reducers';
+import { chapterActions } from 'client/store/actions';
+
+const useStyles = makeStyles(() => ({
+  root: {
+    padding: 15,
+  },
+}));
+
+const renderCardContent = (loading, name, description) => {
```
Zeko369: 1. dont call component `render` + component name, this isn't a render prop
2. Maybe it would be better to move the logic up to the parent component
3. For something like this where you only have a return it would be better to use a shorthand ES6 syntax
```
const foo = (a: number, b: number) => a + b;
```

client/store/types/chapter.ts
```diff
@@ -0,0 +1,45 @@
+import { ThunkAction } from 'redux-thunk';
+import { AppStoreState } from '../reducers';
+
+export const FETCH_START = 'fcc/chapter/CHAPTER_START';
+export const FETCH_SUCCESS = 'fcc/chapter/CHAPTER_SUCCESS';
+export const FETCH_FAIL = 'fcc/chapter/CHAPTER_FAIL';
+
+export interface IChapterModal {
+  id?: string;
```
Zeko369: Something we fetch from the backend will always have an ID, this was probably done so we could have it work client side for a form or something like that, but since it's only used for fetching make it mandatory.

Zeko369:
@Shub1427 In the review I criticized your code, here I'm going to say that this PR is awesome :rocket: I really like the setup and cant wait to start working with it. :nerd_face: 
PS. you still need to fix things from the review :stuck_out_tongue_winking_eye:

package.json
timmyichen: 🎉 nice, thanks!

timmyichen: Actually, on second thought, should we bake this into CI instead of making it a pre commit hook? I can imagine a scenario where you'd want to make some dev commits and it could be annoying to have to lint unfinished code in between each commit

Shub1427: Sounds great! Will remove it. I also found there is a pre-push hook. So, It looks before push the code should be linted anyhow, can put it there? @timmyichen

client/modules/home.tsx
Shub1427: Not sure how everyone else thinks about this, but I usually tend to separate big component hierarchy, into separate functions, just for readability's sake. This also helps in separating out JS loops and conditionally created components inside render.

This could've been called, `getCardContentWithLoader()` for better naming or even separated out to it's own smaller UI Component for re-usability. I think latter seems a better option as per ReactJS norms, where composition is preferred.

Learnt a new pattern, here. :sweat_smile:

client/store/types/chapter.ts
Shub1427: This was kept, so that same modal can be used for POST body as well, where `id` might be `undefined`/`null`.

If that's not the case, will make `id` as _required_.

client/interfaces/components/AddSponsor.tsx
Shub1427: Are you suggesting to keep interfaces, side-by-side to respective files, where they get referenced?

Shub1427:
> Just a few little changes, also I didn't write it down since I still haven't finished the style guide, but could you add a new line between types of import (npm libs, external modules we creates, local modules)

Are you working on some Contributing Guidlines on Coding? GREAT!!
Even I was thinking to do so after this PR. But if you are doing it, thanks.
Please open an issue regarding that so everybody can also contribute their thoughts.




> @Shub1427 In the review I criticized your code, here I'm going to say that this PR is awesome rocket I really like the setup and cant wait to start working with it. :nerd_face:

I never took them as criticism, but a carefully done assesment, helping me out to do better.
Thanks for that @Zeko369.

> PS. you still need to fix things from the review :stuck_out_tongue_winking_eye:

Well I was thinking to skip that and __live my life happily ever__ after. :stuck_out_tongue_winking_eye: :satisfied:

client/store/types/chapter.ts
Zeko369: Yeah, for that case yes, but we're currently not doing that. Anyways just keep it like this for the furure and if we end up not using we can just remove one question mark

client/interfaces/components/AddSponsor.tsx
Zeko369: Maybe a interfaces folder insade of components? Or .d.ts file insde of components folder? Or maybe keep it in the same file? If I think about it this actually seems like a quite elegant solution, so keep it like that maybe?

client/modules/home.tsx
Zeko369: hmmm, I would definatly call it getCardContent or something like that, which implies that it returns JSX. Also this can be made into a shorter syntax. But I would separete the loading out into the parent component and use `?` insead of `variable &&` and `!variable &&`

client/store/types/chapter.ts
Zeko369: OR maybe we could have something like this
```
interface Chapter {
  name: string;
}

interface DBChapter extends Chapter {
  id: string;
}

const a: Chapter = { name: "Foobar" };
const b: DBChapter = { id: "something", name: "string" };
```

Shub1427: Feels like having an extra interface, just for a single key, is an overkill. Keeping it optional seems more simple. We can move to this architecture, if required later. Hope, you agree with me here. :+1:

Shub1427: Will resolve it for now.

client/interfaces/components/AddSponsor.tsx
Shub1427: Moved it to `/client/interfaces`, which is recreating Response types from `/types/models.d.ts` (represents DB models). Responses may or may not differ, but in terms of `interface` naming conventions, that we will be following in this project, this change now looks good.

client/modules/home.tsx
Shub1427: The `Component` looks much better now. :tada:

client/modules/home.tsx
```diff
@@ -0,0 +1,55 @@
...
+const useStyles = makeStyles(() => ({
+  root: {
+    padding: 15,
+  },
+}));
+
+const Home: React.FC = () => {
+  const classes = useStyles();
+
+  const { error, loading, name, description } = useSelector(
```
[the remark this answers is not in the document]
Shub1427: I didn't get it. Wrapping it into a function which would just return what `useSelector` is doing, is just extra overhead, don't you think?

You are suggesting something like this, right:
```jsx
  const chapterSelector = () => useSelector(
    (state: AppStoreState) => ({
      error: state.chapter.error,
      loading: state.chapter.loading,
      name: state.chapter.name,
      description: state.chapter.description,
    }),
  );
```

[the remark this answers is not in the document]
Zeko369: Sorry for the late replay, didn't see this message.
No my idea was to have a
```
const chapterSelector = (state: AppStoreState) => ({...});
const state = useSelector(chapterSelector);
```
since this will probably be used more then once?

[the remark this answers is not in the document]
Shub1427: I am still a bit unsure about this. Can we extract it later. Since later we can find some very strong points of doing this, if this case comes up. If we find duplicate state-maps we can replace this with your idea.

For now I am just thinking it as a re-usable Connected Component, where selectors are not supposed to be duplicated, but the Component could.

Hope you are fine with me here. 😊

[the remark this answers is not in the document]
Shub1427: I was hoping for this to get merged, so that I can open a PR with the TODO list I have mentioned in the PR description. There are some gotcha's with that TODO list, which I want to resolve with everybody's help here.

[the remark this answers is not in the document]
Zeko369: Yeah, I think that's completely fine for now, we can change it later.

client/store/types/chapter.ts
```diff
@@ -0,0 +1,45 @@
...
+  error: string; // Should reflect a generic Error Type here
+}
+
+interface IChapterFetchStartAction {
+  type: typeof FETCH_START;
+}
+
+interface IChapterFetchSuccessAction {
+  type: typeof FETCH_SUCCESS;
+  data: IChapterModal;
```
Zeko369: Wouldn't it be better to use generic Flux names -> `type` and `payload`?

package.json
```diff
@@ -41,11 +41,12 @@
     "@material-ui/icons": "^4.5.1",
```
Zeko369: fix conflict

client/store/types/chapter.ts
Shub1427: Dindn't know that it's kind of a norm. Thanks for letting me know. Will do it 🎉

PR #257, opened 2019-11-21
Sequelize -> TypeORM

Zeko369:
- [x] Update README.md
- [x] Convert chapterModel to typeORM syntax
- [x] Setup package.json scripts (`ts-node` nonsense)
- [x] Example migrations


timmyichen:
LGTM!  Just need to add / update some docs as well. Thanks for doing this 🎉

allella:
@timmyichen I have a list of small README changes, so I'll update to TypeORM in the docs in a forthcoming PR so you all can focus on the technical stuff.

Zeko369:
@allella I already changed it in the README

timmyichen:
@Zeko369 your readme changes revert some things (like contributors), might need to rebase?  also, i was thinking we could add some basic docs for running migrations, etc.  perhaps in a separate `server/README.md`

Zeko369:
Crap, I haven't pulled from upstream, I'll get that back in. Also I'm working on a simple guide for TypeORM features and how to use them, probably will create both Wiki post and `server/README.md`

Haxified:
I like this other than how we're using BaseEntity. I find the repository method much easier and simpler to use.

Zeko369:
@Haxified I think that ActiveRecord pattern is simpler and easier to understand, and since the most used ORM for js/ts is sequelize a lot more people will be familiar with that pattern, also people coming from rails :heart: 
I'll create 2 routers, one with repo mode, and one with ActiveRecord and then people can vote

Haxified:
@Zeko369 oops, didnt see this 👍. Fair enough, although I would like to see the 2 routers side by side. The thing i dislike about the repository method is how you need to run database.getRepository to get the repositories. This is easily fixed with dependency injection, and I would definately use it if this project wasn't aimed at beginners, although if you guys think dependency injection would be good then i could def put in a pr.

Zeko369:
So I created an example of TypeORM "repository" and "ActiveRecord" patterns. My opinion is that we should use ActiveRecord but if anyone has any opinions on why we should use repository mode and not ActiveRecord feel free to comment here

Take a look at server/controllers/chapterController.ts, `index` vs `index_repo`
Vote :rocket:  for ActiveRecord Pattern and :tada: for Repository Pattern

Zeko369:
Will use ActiveRecord pattern since no one voted for RepoPattern

Haxified:
fair enough, i'm happy with both and i've used both a lot

Zeko369:
I think that we should get this merged so that we can work on converting models from sequelize-ts to typeorm #162 , and I'll create another PR for seeds

PR #287, opened 2019-12-08
Add Models using TypeORM

ankorGH:

Add models using typeorm as shown in db diagram.


Zeko369:
Could you order imports by the importance?
1. npm module
2. aliased modules
3. local imports

server/controllers/chapterController.ts
```diff
@@ -21,8 +21,22 @@ export default {
   },
 
   async create(req: Request, res: Response) {
-    const { name, description, category, details } = req.body;
-    const chapter = new Chapter({ name, description, category, details });
+    const {
```
Zeko369: This unfortunately wont work at property at runtime, beacuse here we implicitly convert location (string) from body into a Location object, change this to location_id and creator_id (for, now later we'll get that from logged in user), and query location and user before inserting a new chapter

ankorGH: @Zeko369  i think at runtime TypeORM converts an id to an entity object

Zeko369: Yeah, I was wrong, it actually work, but sending `location_id` and `creator_id` doesn't, so do you think we should change it to `location` and `creator` in `api docs` or make 2 more queries to the db here?

QuincyLarson:
@ankorGH How is this going? Are you ready for @Zeko369 to take a second look at this? Thanks again for your help.

server/controllers/chapterController.ts
ankorGH: @Zeko369 sorry for the delay. I think we should change it to `location` and `creator`. Since, TypeOrm fails if it is not a valid id we can choose to provide helpful error message instead of making 2 more request to the db.

Zeko369: Yeah, let's do that then

server/models/UserBan.ts
```diff
@@ -0,0 +1,24 @@
+import { Entity, ManyToOne, JoinColumn } from 'typeorm';
+import { BaseModel } from './BaseModel';
+import { User } from './User';
+import { Chapter } from './Chapter';
+
+@Entity({ name: 'user_bans' })
+export class UserBan extends BaseModel {
```
Zeko369: Could we maybe instead of doing this create a `joinTable` https://typeorm.io/#/many-to-many-relations? Also the same for `EventSponsor` and `UserChapter`

server/controllers/chapterController.ts
ankorGH: Alright cool, I am on it.

server/models/UserBan.ts
Zeko369: What do you think?

ankorGH: yeah, we could. I was thinking of a situation where it doesn't extend `BaseModel` but I didn't test it first.

Zeko369: https://typeorm.io/#/many-to-many-relations/many-to-many-relations-with-custom-properties :cry: , for them to have `createdAt` and `updatedAt` we need custom entities, so maybe we could create BaseModel that only has `createdAt` and `updatedAt` and add ID to all models that need it?

ankorGH: I think what we currently have will also work since `UserBan` also needs an `id`.

Zeko369: Ok, keep it like this then

ankorGH:
@QuincyLarson I am great. You? Hopefully, I and @Zeko369 will get this PR soon.

server/controllers/chapterController.ts
```diff
@@ -21,20 +21,51 @@ export default {
...
+      details,
+      location,
+      creator,
+    });
 
     try {
       await chapter.save();
       res.status(201).json(chapter);
     } catch (e) {
+      if (e.code === '23503' && e.message.includes('foreign key constraint')) {
```
Zeko369: This will be a bitch to setup so maybe we should extract this for later reuse

server/controllers/chapterController.ts
```diff
@@ -26,30 +26,46 @@ export default {
...
+        if (e.detail.includes('location')) {
+          return res.status(404).json({ message: 'location not found' });
+        }
+        if (e.detail.includes('user')) {
+          return res.status(404).json({ message: 'creator not found' });
+        }
+      }
+
       res.status(500).json({ error: e });
```
imVinayPandya: Is is good idea to throw full error stack in api response?

server/controllers/chapterController.ts
```diff
@@ -26,30 +26,46 @@ export default {
...
+      location,
+      creator,
     });
 
     try {
       await chapter.save();
       res.status(201).json(chapter);
     } catch (e) {
+      if (e.code === '23503' && e.message.includes('foreign key constraint')) {
+        if (e.detail.includes('location')) {
```
imVinayPandya: Just an edge case, if e.detail is **null** or **undefined** than it will throw an error like ".includes of undefined". To prevent this we can use destructing with default value.

PR #291, opened 2019-12-15
Add schema sync instructions

AryanJ-NYC:

I noticed that instructions for syncing the database with the schema (something that is necessary to initially use the application) were lacking.

I went ahead and added a script plus instructions to use the script upon app standup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,8 @@ docker-compose up
 
 Wait for the logs to show "server started on port 8000", then navigate to `localhost:8000` to view the app.
 
+To sync the database with the schema, run `npm run typeorm:sync-schema`.
+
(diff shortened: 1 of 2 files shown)


Zeko369:
We shouldn't be doing this, we use migrations for this purpose.
can you maybe add a timestamp of your stream where you discuss this or write why here?

AryanJ-NYC:
Ah, you're completely right. My bad. I basically did it because I followed the README and found the Chapters table wasn't in the DB after following the instructions.

AryanJ-NYC:
OK, yeah. I found the migrations. 🤦‍♂ 

Should we include instructions on the front README to migrate after `docker-compose up`?

AryanJ-NYC:
OK, yeah. So the migrations aren't running programatically. Will fix.

PR #294, opened 2019-12-15
Add migrations to db.ts

AryanJ-NYC:

Migrations were _not_ running programatically. Specifically, `await connection.runMigrations();` was not doing anything as it did not find any migrations.

This was due to the custom config passed into `createConnection` lacking a `migrations` property. I've went ahead and added this.

Some additional context here: https://github.com/freeCodeCamp/chapter/pull/291

### DIFF
--- a/server/db.ts
+++ b/server/db.ts
@@ -1,16 +1,18 @@
 import 'reflect-metadata';
-import { createConnection } from 'typeorm';
+import {
+  createConnection,
+  getConnectionOptions,
+  ConnectionOptions,
+} from 'typeorm';
 
-export const initDB = createConnection({
-  type: 'postgres',
-  username: process.env.DB_USER || 'postgres',
-  password: process.env.DB_PASSWORD || 'password',
-  database: process.env.DB_NAME || 'chapter',
-  host: process.env.DB_URL || 'db',
-  port: 5432,
-  entities: [__dirname + '/models/*.ts'],
-  migrations: [__dirname + '/migrations/**/*.ts'],
-  synchronize: false,
-  logging: false,
-  name: 'default',
-});
+export const initDB = (async () => {
+  const connectionOptions = await getConnectionOptions();
+  return createConnection({
+    ...connectionOptions,
+    username: process.env.DB_USER || 'postgres',
+    password: process.env.DB_PASSWORD || 'password',
+    database: process.env.DB_NAME || 'chapter',
+    host: process.env.DB_URL || 'db',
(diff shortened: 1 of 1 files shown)


server/db.ts
```diff
@@ -9,6 +9,7 @@ export const initDB = createConnection({
   host: process.env.DB_URL || 'db',
   port: 5432,
   entities: [__dirname + '/models/*.ts'],
+  migrations: [__dirname + '/migrations/**/*.ts'],
```
Zeko369: Use await getConfig or something like that I dont know the specific api

AryanJ-NYC: I'm unsure I understand. Are you referring to adding https://www.npmjs.com/package/getconfig?

Zeko369: No, sorry, I was thinking about importing the config, not sure if typeorm can do it itself, but just importing the  typeorm config file and spreading it here?

AryanJ-NYC: Ah, understood.

If that's the case, we should switch to an `ormconfig.js` (rather than `.json`) so we can support environment variables.

Additionally, it looks like `createConnection` and `ormconig.json` have two different ports (because of the `54320:5432` mapping in `docker-compose`). Unsure how that should be handled. Would require some `if insideOfDockerContainer` type of logic.

We should probably:
a) Change the `docker-compose` mapping to the norm (`5432:5432`)
b) Convert `ormconfig.json` to `ormconfig.js` to support environment variables
c) Finally get rid of the custom config in this file and default to `ormconfig.js`.

Thoughts?

AryanJ-NYC: @Zeko369 OK, so I was mistaken. I see what you were saying and I think the above implementation satisfies that.

Zeko369: I wouldn't change the port mapping because if someone has `psql` running on the system it's on the 5432, so this would make it possible for people to have both

image

Zeko369: We'll talk about this in the meeting in a hour

AryanJ-NYC: OK, that's cool. I haven't done that so this PR should be safe.

AryanJ-NYC:
@Zeko369 Can you try bringing down all your containers and bringing them back up? `docker-compose down && docker-compose up`?

AryanJ-NYC:
It works. Standing up the server takes quite a while but it definitely works...

Screen Shot 2019-12-16 at 8 15 56 AM

We should probably have someone else try though. I'd rather be sure it works rather than it being just on my machine.

PR #298, opened 2019-12-15
Add chapter CRUD API

### LINKED ISSUE #105: API: As an administrator, I can create a chapter.

Zeko369:
Cleaning up seeds, and "finishing" chapter api from #105 

Closes #105

### DIFF
--- a/server/factories/user.factory.ts
+++ b/server/factories/user.factory.ts
@@ -1,0 +1,13 @@
+import Faker from 'faker';
+import { define } from 'typeorm-seeding';
+import { User } from '../models/User';
+
+define(User, (faker: typeof Faker) => {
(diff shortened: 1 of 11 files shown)


timmyichen:
The above answered on stream - right now only auth method will be through Google OAuth

server/factories/user.factory.ts
```diff
@@ -0,0 +1,13 @@
+import Faker from 'faker';
+import { define } from 'typeorm-seeding';
+import { User } from '../models/User';
+
+define(User, (faker: typeof Faker) => {
+  const location = new User({
```
vaibhavsingh97: @Zeko369 it's should be `const user`

PR #300, opened 2019-12-16
Add first pass at header component

AryanJ-NYC:

This adds a basic Header to all pages of Chapter.

# Screenshots
Screen Shot 2019-12-15 at 9 15 16 PM
Screen Shot 2019-12-15 at 9 19 35 PM


client/components/PageLayout/Header.tsx
```diff
@@ -0,0 +1,17 @@
...
+const Header = ({ classes }) => (
+  <header className={`${classes.pageRoot} ${classes.headerContainer}`}>
+    <div>Search Bar here</div>
+    <img src="/freecodecamp-logo.svg" alt="The freeCodeCamp logo" />
+    <nav>
+      <ul className={classes.navList}>
+        <li className={classes.navListItem}>Events</li>
+        <li className={classes.navListItem}>Chapters</li>
+        <li className={classes.navListItem}>Login</li>
+        <li className={classes.navListItem}>Register</li>
```
Shub1427: It's good to create this component as a separate `ui` component and re-use it by passing props to it, (i.e. more generic component, that can take props and populate it's list items).

The point is to have separation of concerns (UI Component and Connected or Component passing data as props). This way, we can create more stable UI snapshots, that can later on even get separated into our own UI Components  (maintained and displayed in `Storybook` or some other UI Documentations).

AryanJ-NYC: I'm a bit confused by the ask. What are you suggesting I do? This is not a "connected component." It simply displays UI.

Shub1427: Sorry for my bad English. What I meant to say, is to separate this Component into a generic UI component and re-use it in parent component, which can pass it the Data as props, either from Redux Store or via some inner maintained state. 

Hope this clears the doubt 👍

AryanJ-NYC: Ah, understood. Why do you feel that's the best approach?

At this early stage, it's very hard to tell if and how this component will be re-used. In my opinion, it'll be easier to prepare for re-use at the point of its first re-use.

I'd rather avoid making this component generic and take in its hrefs (which it'll soon have) and link names as props when this'll likely be its only use app-wide.

AryanJ-NYC:
> Also, it helps to create more stable Snapshots, as the code won't change due to Data logics.

When does this code change with any logic?

client/components/PageLayout/Header.tsx
AryanJ-NYC: > Also, it helps to create more stable Snapshots, as the code won't change due to Data logics.

When does this code change with any logic?

AryanJ-NYC: And as for Storybook, I'm not super familiar with Storybook. Does this not meet the Storybook spec in any way? Should developers be optimizing for Storybook?

Shub1427: I agree with your point about overengineering, but my point was to keep Figma designed components as separate UI components and re-use those components for more flexibility (which doesn't seem an overkill). Since requirements can change and so do our logic and designs, it is better to separate concerns and put different components for each, one which handles logic and one which handles the designs, which u already have mentioned, that u will separate them into two, later on.

I am suggesting we can do it right now, and keep a separate `components/ui` for all our design components.

AryanJ-NYC: Ah, got it. So your ask is that I put it in the `components/ui` folder?

Shub1427: * https://dev.to/tomekbuszewski/high-level-view-and-logic-separation-in-react-39n0
* https://medium.com/teamsubchannel/react-component-patterns-e7fb75be7bb0

Your changes are just fine according to this. What I am suggesting is to put all our view/styled-components in one place, containers/layout components in one place, and Pages will use all these layout components to structure themselves.

Something like:
* `/client/ui` - for all designed/styled components, whose styles most-probably won't be changed in parent (unless really-really necessary).
* `/client/components` - Components that can re-use views to be layout properly, or pass data and manage logic in themselves.
* `/pages` - NextJS Components that uses above components, to present a page.

## According to above suggestion:

- `/client/ui/header.tsx` - Styled component that is generic and can take props like:
```js
{
  brand: PropTypes.node, // Since, we can pass an image here or an svg, or just a string as a brand name
  links: PropTypes.arrayOfType(PropTypes.shapeOf(someObject)), // Since, we can pass object that contains details on `<Link>` from `next`, with extra details like, it's a dropdown-ui or just a simple re-route link.
  action: PropTypes.node, // Since, it could be a searchbar or plain text or something else for different pages
}
```

```tsx
import React from 'react';

const Header = (props) => (
  <header className={`${classes.pageRoot} ${classes.headerContainer}`}>
    {props.action && <div className={styles}>props.action</div>}
    {props.brand && <span className={styles}>props.brand</span>}
    <nav>
      <ul className={classes.navList}>
        {mapToElements(props.links)}
      </ul>
    </nav>
  </header>
);
export default Header;
```

--------------------

* `/client/components/layout/header.tsx` - maintains the state of header-ui
* `/client/components/layout/index.tsx` - manages layout using above header and other ui components if needed.
> For above components we don't have to provide styles for UI, but just for their layout, as it's might be necessary

-------------------------------

*  `/pages/home.tsx` - re-uses `/client/components` and creates a page, adds any server logic in `getInitialProps`.

--------------------------------------------------

I hope you get my point now 👍

AryanJ-NYC: OK, definitely understand your point. Thanks for pointing that out.

I actually fundamentally disagree though (d'oh!). I believe in colocation. Basically, since `Header` is only used in the `PageLayout` component, the file should be close to the `PageLayout` component. This makes it much easier to find (it's in the same exact folder) and avoids the need for users to dig through a completely different folder.

I 💯 agree with having UI components, I just don't think that they should be banished to some other far away folder just because they are UI components. Colocation is key, in my opinion.

I'd love to hear other opinions though. I'll ping the chat.

timmyichen: There are lots of different ways to organize these components (I'm for a third option not yet mentioned) but I don't think it's terribly important right now.  Reorganizing components would be pretty trivial in the near future.

However, I do not think the header is a generic component - there is only one header and what it displays will be relatively static (would only change between authed / non-authed states).

This seems like the solution should be simple, eg:
* `client/components/SomeGenericReusableComponent.tsx`
* `client/layout/Header.tsx`

client/components/PageLayout/pageLayout.styles.tsx
```diff
@@ -0,0 +1,41 @@
+import { makeStyles } from '@material-ui/core';
```
timmyichen: This file should be a `.ts` file as it doesn't have JSX

client/components/PageLayout/Header.tsx
timmyichen: I do, however, think that links should be in some kind of config object, e.g:

```javascript
const headerLinks = [
  { name: 'events', label: 'Events', href: '/events' },
  ...etc
]
```

Shub1427: I agree, there are various ways to structure a project, but still don't you guys think, keeping our own UI stuff separate, is a good way to support composition and help specific people to work on specific tasks. Say, a css-designer and a developer wants to work on a component, keeping UI component in a separate file, helps the designer to focus on designs, without conflicting with logical dev changes. While, the Dev than can use that UI component, create a composed component and add/change logic, without conflicting the design changes.

I wanted to go this way, as Chapter seems to be a project, whose end-user would be random, and designs could be very different considering each end-user.

AryanJ-NYC: @Shub1427 Do we currently have CSS designers that will work on one file and developers that will work on another? Is this a current use case.

@timmyichen Good call on headerLinks being in config object.

Shub1427: This is an opensource project, anybody can contribute. The more the better!

AryanJ-NYC:
@timmyichen Changes made.

client/components/PageLayout/pageLayout.styles.ts
```diff
@@ -0,0 +1,41 @@
+import { makeStyles } from '@material-ui/core';
+
+const useStyles = makeStyles(theme => ({
+  headerContainer: {
+    'align-items': 'center',
+    backgroundColor: theme.palette.secondary.main,
+    color: 'white',
+    display: 'grid',
```
vaibhavsingh97: We should use a material-UI grid instead? thoughts?
Ref: https://material-ui.com/components/grid/#grid

AryanJ-NYC: @vaibhavsingh97 Good call. Done.

vaibhavsingh97:
@timmyichen can you please review and merge

client/config.ts
```diff
@@ -0,0 +1,6 @@
+export const headerLinks: { name: string; label: string; href: string }[] = [
```
Zeko369: Maybe just extract this into an interface

Zeko369: but keep it in the same file

Zeko369:
Just a small fix, and then I'll merge it

client/config.ts
AryanJ-NYC: Done. However, this sounds a bit subjective and should be documented somewhere. Preferably a linter file.

PR #302, opened 2019-12-16
feat(types): Add postgres error constants enum

Haxified:

Made a new enum based on a conversation in #299. Now, instead of doing `if(e.code === '23503')`, you can do `if(e.code === PostgresErrorCodes.FOREIGN_KEY_VIOLATION)` (imo, much cleaner.) 

In #299, a library was suggested, but I felt like adding an extra dep was a bad idea when the implementation was very simple.


Zeko369:
YES

Issue #270, opened 2019-11-30
Auth: As an owner or chapter admin I can create events and set their venue type, venue, video url, tags and capacity.


QuincyLarson:
Depends on https://github.com/freeCodeCamp/chapter/pull/814

The UI is looking solid, so the main outstanding issue is to add auth and test it, similarly to 
https://github.com/freeCodeCamp/chapter/issues/273

- [x] As an owner I can create events in any Chapter
- [x] As an admin of Chapters, I can create events in those Chapters (but no others)
- [x] As any other role, I cannot create events

The main thing to check is that requests from unauthorised users fail, but it's also helpful to confirm that the UI reflects their role.

Zeko369:
Hey @madaleneaza-design how is the design for this coming along? I was thinking about building the api for this tomorrow, and also doing a skeleton in react for this, so if you already have something I'd like to implement that

allella:
Venue Type is an addition field discussed in 
https://github.com/freeCodeCamp/chapter/issues/378
and 
https://github.com/freeCodeCamp/chapter/issues/382

allella:
@katien 

https://github.com/freeCodeCamp/chapter/issues/270 is a relevant user story for the UI side.

We don't have a more granular issue for implementing specific form fields to show/hide other location or video URL field based on the type of event, so for now this issue / user story could cover the form fields on the event creation page.

allella:
The idea on the [event's venue type UI element](https://github.com/freeCodeCamp/chapter/issues/378) would be to only display the relevant form fields.

- if Venue Type = Physical, then show address fields
- if Venue Type = Online, then show video URL field

allella:
@katien I'm coping the authorization notes for CRUD an event here since they are relevant to this user story.

(via Discord) I don't think we have the permissions logic entirely figured out yet, but the ability for a *user* to post *events* on 1 or more *chapters* will depend on their authorization. 

Authorization will come from either of
- **user_chapter_roles** with possible role values **_organizer_**, _**member**_. A _user_ can have 0 or more roles per _chapter_.
- **user_instance_roles** with possible role values _**owner**_, _**administrator**_

Lots of earlier discussion on this issue led to this summary https://github.com/freeCodeCamp/chapter/issues/296#issuecomment-585347150 which was then closed via PR: https://github.com/freeCodeCamp/chapter/pull/335

[...]

PR #311, opened 2019-12-25
List of events for admin dashboard

Zeko369:
This relates to #270 (before we can create we should first be able to show that we created)


client/store/actions/events.ts
```diff
@@ -0,0 +1,49 @@
...
+): eventsTypes.IEventActionTypes => {
+  return {
+    type: eventsTypes.FETCH_SUCCESS,
+    payload: { events, chapterId },
+  };
+};
+
+export const fetchFail = (error: string): eventsTypes.IEventActionTypes => {
+  return {
+    payload: error,
```
vaibhavsingh97: not an issue: just we should have consistency 😅in return statement

pages/dashboard/events.tsx
```diff
@@ -0,0 +1,64 @@
...
+    loading: state.events.loading,
+    events: state.events.events,
+  }));
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(eventActions.fetchEvents('1'));
+  }, []);
+
+  console.log(error, loading, events);
```
vaibhavsingh97: please remove `console.log` statement

pages/_app.tsx
```diff
@@ -24,18 +24,21 @@ export default class MyApp extends App {
...
-              <CssBaseline />
-              <PageLayout>
-                <Component {...pageProps} />
-              </PageLayout>
-            </>
-          </StyledComponentsThemeProvider>
-        </MaterialUIThemeProvider>
-      </Provider>
+      <>
+        <title>Chapter</title>
```
timmyichen: Why here instead of in the `<Head>` component?

client/store/actions/events.ts
Zeko369: So just change the order?

pages/_app.tsx
Zeko369: Good question, I'll move it to a Head component in here

timmyichen: It looks like it was already moved from the `_document.tsx` file which was in a `<Head>` component already O_o

Zeko369: Yeah, I moved it from there because of a Next.js error saying we could have a conflict and it would be unpredictable if we used `<Head><title></title></Head>` somewhere else
https://github.com/zeit/next.js/blob/master/errors/no-document-title.md

timmyichen: it seems like `<Head><title></title></Head>` is the pattern they want us to use, just not in `_document.tsx`, right?

Zeko369: Yeah, I forgot `<Head/>` here 🙈

client/store/actions/events.ts
vaibhavsingh97: yes, I mean order should be consistent. code looks good 😬

client/store/types/events.ts
```diff
@@ -0,0 +1,55 @@
+import { ThunkAction } from 'redux-thunk';
+import { AppStoreState } from '../reducers';
+
+export const FETCH_START = 'fcc/chapter/EVENTS_START';
```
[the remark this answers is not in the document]
Zeko369: Maybe better to move them to `store/actions/modelName.ts`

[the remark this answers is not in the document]
Zeko369: Or just create another file here?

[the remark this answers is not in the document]
vaibhavsingh97: yes, component-wise action types would be good, if you have any other suggestions, lemme know

Zeko369:
@timmyichen the design isn't anything special
image

client/store/types/events.ts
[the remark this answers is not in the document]
Zeko369: Is this (latest commit) what you were thinking of?

timmyichen:
thanks for the screenshot! they're nice to see regardless, i think - gives spectators some insight into how things are going 🙃

docker-compose.yml
```diff
@@ -27,11 +27,11 @@ services:
...
       - '8000:8000'
-  speccy:
-    image: node:10-alpine
-    command: 'npm run speccy:watch'
-    working_dir: /usr/chapter
-    volumes:
-      - .:/usr/chapter
-    ports:
-      - '8001:8001'
+  # speccy:
```
Kanapka: this looks like an accidental change, adding event list does not require turning speccy off

pages/dashboard/events.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <>
+      <Grid container spacing={2}>
+        <Grid item xs={12}>
+          <button>Add new</button>
+          {error ? (
+            <h1>😢Error</h1>
+          ) : (
+            events.map((event: IEventModal) => (
+              <Card style={{ marginTop: '12px' }} key={`event-${event.id}`}>
```
Kanapka: might be separated to an event component

Issue #296, opened 2019-12-15
Add user_roles table for organizers and admins


vkWeb:
Create a `user_roles` table with `id` and `type` as string and reference the `id` of `user_roles` via `users` table.

AryanJ-NYC:
Relates to https://github.com/freeCodeCamp/chapter/issues/16.

madaleneaza-design:
Based on the discussion in Discord, I have explained what I thought  each role (organiser and admin) could do:

**Admin**
Only the administrator is able to do create chapters, although they have full control of what is going on in the Chapter instance, they probably will have little involvement in specific chapters (such as CRUD events, managing chapter members unless the situation is high-level priority). Other than that, they aim to focus on admin tasks and keeping the Chapter sailing afloat.

**Organiser**
the chapter organiser focuses on managing and organising events for the chapter (and events) they are assigned to, that includes any kind of tasks relating to their chapter only. 

Also, this can be useful for chapters to have more than one organiser (or at least a co-organiser), as unforeseen circumstances may occur and a particular organiser may not be able to do their role, at least the co-organiser can fill in the gap (even if it is just temporarily).

vkWeb:
@madaleneaza-design @allella I think the organizers shouldn't be able to add other organizers to the chapter they are appointed to. Suppose fCC is the organization and Quincy is the admin. Now the admin appointed me as an organizer for fCC Delhi (a chapter under fCC). If I add Tim as an organizer of fCC Delhi then we will need to put there a feature to inform Quincy that Tim has been added as an organizer for fCC Delhi by me because in Quincy's mind I should be the only organizer. 

We can make this thing more feature-rich by giving admin an option to "allow the organizer to add more organizers" but I don't think this should be in our MVP. So only administrators should be able to add new organizers to keep things a bit straightforward for our MVP. If the admin needs help or is busy then he/she can add more admin(s).

So here's my take on the permissions of administrators and organizers: 

**Administrators**
- can add, remove admins
- can add, remove organizers for a chapter
- can CRUD chapters and its details
- can CRUD any chapters' event and its details
- can email, ban any chapter organizer 
- can email, ban users
  - can email or ban only when the user has attended an event under any of the chapters
  - can email only when the user has subscribed to one of the chapter
- has full control over the Chapter instance and site-wide settings

**Organizers**
- can CRUD events and its details only under the chapter they are appointed to
- can email, ban users under the chapter they are appointed to
[...]

PR #313, opened 2019-12-30
Update README Terminology to define and split out admin vs organizer terms

allella:

Closes Nothing.

[See my fork for a visual of the changes](https://github.com/allella/chapter/tree/docs/readme-admin-vs-organizer#terminology).

- The user roles were mixed into the terminology but not well defined.
- the use of the word "admin" in the _chapter_ row was in conflict with [the most recent confirmation of _administrator_ vs _organizer_ permissions](https://github.com/freeCodeCamp/chapter/issues/296).

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,13 +16,13 @@ To better communicate and more easily build an API and UI, the current contribut
 
 | Term          | Definition      | Example  |
 | ------------- | ------------- | ----- |
-| _instance_      | a web server deployment of this _Chapter_ application, managed by an organization. | a Docker container running on a web host |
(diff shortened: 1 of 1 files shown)


vkWeb:
I think we should confirm this with @QuincyLarson. If he agrees to the permissions we are giving to organizers and admin then I'll be happy to merge.

allella:
Using @QuincyLarson 's "prior art" of moving things along if they sit unanswered, I'm merging this in since it's more clear than what we have and it's only a README change.

We're also having a conversation on #296 which depends on setting the user_role types, which should correspond to this terminology.  We'll figure out if the terminology needs tweaking in the course of defining the user_role table.

QuincyLarson:
@allella I'm a bit late to the party, but I just want to confirm - these are nice, succinct definitions.

PR #316, opened 2020-01-27
Feature/remove styled components

Zeko369:
- [x] Remove old AddSponsor components
- [x] Refactor examples
- [x] Update README.md


allella:
@Zeko369  out of curiosity, is this essentially switching things to a Material UI methodology and its default styling?

Zeko369:
Yeah, because materialUI has built in CSS-in-JS styling for when we need to use custom styles

client/components/SomeComponent/index.tsx
```diff
@@ -1,32 +1,25 @@
 import * as React from 'react';
 import Button from '@material-ui/core/Button';
 import { Grid } from '@material-ui/core';
-import styled from 'styled-components';
-import { withTheme } from '@material-ui/core/styles';
+// import styled from 'styled-components';
```
ScottBrenner:
```suggestion
```

Remove commented-out code?

Zeko369: This PR should have been a draft, it's not done yet

README.md
```diff
@@ -34,7 +34,7 @@ We are planning to use the following tools:
 * [Postgres](https://www.postgresql.org) with [TypeORM](https://typeorm.io/#/)
 * [Next.js](https://nextjs.org/) for both client and server-side rendering of the frontend (NextJS is based on [React](https://reactjs.org))
   * [JavaScript/TypeScript](https://www.typescriptlang.org/index.html#download-links)
-  * [Styled Components](https://www.styled-components.com) for styling.
+  * [Material UI](https://material-ui.com/) for components and it's built in `makeStyles` hook and `styled` HOC for custom styling
```
ScottBrenner:
```suggestion
  * [Material UI](https://material-ui.com/) for components and its built-in `makeStyles` hook and `styled` HOC for custom styling
```
Also, what is an HOC?

Zeko369: Thx for the typo fix.
HOC or higher order components is a pattern where you wrap a component in a function

```jsx
const StyledGrid = styled(Grid)({
  bacgroundColor: "red"
});
```

I specifically mentioned that because I myself didn't know material UI supported this, and this is a main way you do styling with styled components.

client/components/AddSponsor.tsx
```diff
@@ -29,32 +36,43 @@ const AddSponsor: React.FC = () => {
 
   return (
     <>
-      {(responseMsg || errors) && <ResponseDiv>{responseMsg}</ResponseDiv>}
-      <Form onSubmit={handleSubmit(onSubmit)}>
-        <label>Sponsor Name: </label>
-        <Input
+      {(responseMsg || errors) && <div>{responseMsg}</div>}
```
timmyichen: was the `ResponseDiv` margin intentionally left out?

client/components/AddSponsor.tsx
```diff
@@ -29,32 +36,43 @@ const AddSponsor: React.FC = () => {
...
-        <select name="type" required ref={register}>
-          <option value="FOOD">Food</option>
-          <option value="BEVERAGE">Beverage</option>
-          <option value="OTHER">Other</option>
-        </select>
-        <SubmitBtn type="submit">Add Sponsor</SubmitBtn>
-      </Form>
+        <label className={styles.item}>
+          Sponsor Type:
+          <select name="type" required ref={register}>
```
timmyichen: any reason to not use the material-ui `Select` component here?

client/components/SomeComponent/__test__/Index.test.tsx
```diff
@@ -4,15 +4,15 @@
...
-import { themeObject } from 'styles/theme';
+import theme from 'styles/theme';
 
 describe('SomeComponent', () => {
   it('should render text', () => {
     const { container } = render(
-      <ThemeProvider theme={themeObject}>
-        <SomeComponent />
+      <ThemeProvider theme={theme}>
+        <SomeComponent />,
```
timmyichen: remove comma from jsx

Zeko369: Actually not, this is a param in the render() function. And we have `"trailingComma": "all"`, so prettier adds it automatically

client/components/AddSponsor.tsx
Zeko369: Great idea 🚀

client/components/SomeComponent/__test__/Index.test.tsx
timmyichen: That shouldn't be the case because ThemeProvider is still unclosed be, so it'll render in the dom unless I'm missing something? There's a comma on the next line for ThemeProvider which makes sense since it's the end of the jsx block but that shouldn't be the case here

client/components/AddSponsor.tsx
Zeko369: Completely forgot about that component, but since this is only a skeleton a simple div should be fine, don't you agree?

client/components/SomeComponent/__test__/Index.test.tsx
Zeko369: Crap, there are two commas, only noticed one in this small snippet 🙈

client/components/AddSponsor.tsx
timmyichen: I'm fine with it, though I don't know what it looks like at the moment :P

Though looking at this a second time I'm not sure what it's supposed to be doing here.  If `responseMsg` is falsy but `errors` is truthy, it'll just render an empty div, and if `responseMsg` is truthy then it doesn't matter what `errors` is because it's not used in the jsx?

Zeko369: Honestly I didn't really look at the logic, just changed from styled-components to use materialUI's native JSS styling (or in this case remove it). We should maybe do something like
```jsx
{error ? <h1>Error: {responseMsg}</h1> : <div>{responseMsg}</div>}
```

timmyichen: I'm fine with either of them but I think someone else will have better context (whoever implemented that piece)

Zeko369: @joelrozen wrote this, any comments, did we miss something?

joelrozen: Hmmm... I don't remember fully, but I think perhaps I was going to have `errors` be related to input / frontend errors, but then just kept it as the browser based validation. 
Would make sense, sense `errors` is completely undefined and my errorhandler in submitForm just sets `responseMsg` anyway, could basically just keep it as `responseMsg && <div>{responseMsg}</div>`

PR #324, opened 2020-01-30
Update packages to fix vulnerabilites

Zeko369:
We still have some potential vulnerabilities in speccy, but since it's only used in dev and it's soon going to be replaced with SwaggerUI we can ignore that

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7084,9 +7084,9 @@
       "dev": true
     },
     "handlebars": {
-      "version": "4.4.3",
-      "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.4.3.tgz",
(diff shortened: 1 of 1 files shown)


vkWeb:
@Zeko369 Merging this since Tim approved it. 

If I'm not wrong there is a mechanism via dependabot by which it automatically creates a PR to fix vulnerabilities. Should we use that feature in our repo?

Shub1427:
https://github.com/greenkeeperio/greenkeeper

Never used it, but heard good things about it. Also have seen good amount of Open Source projects using it 👍

Zeko369:
Github has dependabot, which is their own implementation of greenkeeper, maybe try using that first?

PR #330, opened 2020-02-12
cleanup imports; remove pre-push linting

timmyichen:

The PR:
* Cleans up factory imports to use aliases instead of `../`
* Removes husky pre-push linting.  It's pretty slow (takes a few minutes to run on my machine) and there's no need to do this prior to pushing.  Any linting errors will be caught in CI; or, if you're unsure, it can be run itself.


package.json
```diff
@@ -131,4 +130,4 @@
     "node": "10.x",
     "npm": ">= 6.0.0"
   }
-}
+}
```
Zeko369: Why did you remove the newline?

timmyichen: Accident, fixing!

Issue #54, opened 2019-10-16
Schema image / ER Diagram is out of date


bernhard-hofmann:
Whilst I like the image of the schema, it's not auto-generated on updates to the SQL script so it could be misleading.

The version there is generated by a tool called DataGrip from JetBrains which is [not free](https://www.jetbrains.com/datagrip/buy/#commercial?billing=yearly).

bernhard-hofmann:
My recommendation would be to (in order of preference):
1. generate it on updates to the SQL script (are there tools to do that, online maybe?)
2. remove it
3. give the filename a date and time to make it clear when it was generated

dmmulroy:
Also, given that Chapter has now been decided to be a self-hosted solution I'm not sure this schema makes 100% sense any more, especially if there is no "central" service for keeping track of all users/groups/events across all instances.

allella:
@kognise also has a [proposed "terminology" update to the readme](https://github.com/freeCodeCamp/chapter/pull/61#pullrequestreview-302721424) to call a "chapter" effectively what a "group" is in the schema.  Group is pretty vague and it would be nice to have some parity between the core item in the schema and what we're referring to in the business logic terminology.

allella:
@nik-john is changing "group(s)" to "chapter(s)", so when this get's updated then please reflect those changes in the schema image.
https://github.com/freeCodeCamp/chapter/issues/17#issuecomment-542814020

nik-john:
You can see the changes [here](https://github.com/freeCodeCamp/chapter/pull/46/commits/6d67bd27a0ad438ccbbc494772108e7efe372bd1)

bernhard-hofmann:
Sorry to ask, but can we bring this issue back on topic?

I'd really like a schema diagram; I find them very useful to understand the schema without having to create the image in my head from SQL text.

But how are we going to ensure it's kept up to date without relying on one person to re-create it whenever someone updates the SQL?

I'll have a look at DBeaver and PGAdmin, thanks @dmmulroy. Maybe we can include steps asking that PRs that update the SQL include an updated schema image?

dmmulroy:
I think requesting an updated schema image is a reasonable request, esp. if DBeaver and PGAdmin diagrams are acceptable.

allella:
[...]

PR #335, opened 2020-02-13
add tables for chapter/instance roles

### LINKED ISSUE #296: Add user_roles table for organizers and admins

timmyichen:

Closes #296

Adds the role tables for chapter roles (member, organizer) and instance roles (admin, owner).

### DIFF
--- a/server/models/UserChapterRole.ts
+++ b/server/models/UserChapterRole.ts
@@ -1,0 +1,37 @@
+import { Entity, JoinColumn, ManyToOne, PrimaryColumn } from 'typeorm';
+import { User } from './User';
+import { Chapter } from './Chapter';
+
+export type ChapterRoles = 'organizer' | 'member';
+
+@Entity({ name: 'user_chapter_roles' })
+export class UserChapterRole {
+  @PrimaryColumn()
+  user_id!: number;
+
+  @PrimaryColumn()
+  chapter_id!: number;
+
+  @PrimaryColumn()
+  role_name!: ChapterRoles;
+
+  @ManyToOne(_type => User)
+  @JoinColumn({ name: 'user_id' })
+  user!: User;
+
+  @ManyToOne(_type => User)
+  @JoinColumn({ name: 'chapter_id' })
+  chapter!: Chapter;
+
+  constructor(params: {
+    userId: number;
+    roleName: ChapterRoles;
+    chapterId: number;
+  }) {
+    if (params) {
+      this.user_id = params.userId;
+      this.role_name = params.roleName;
+      this.chapter_id = params.chapterId;
+    }
+  }
+}
--- a/server/models/UserInstanceRole.ts
+++ b/server/models/UserInstanceRole.ts
@@ -1,0 +1,24 @@
+import { Entity, JoinColumn, ManyToOne, PrimaryColumn } from 'typeorm';
+import { User } from './User';
+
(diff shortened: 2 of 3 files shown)


server/models/UserInstanceRole.ts
```diff
@@ -0,0 +1,24 @@
...
+  user_id!: number;
+
+  @PrimaryColumn()
+  role_name!: InstanceRoles;
+
+  @ManyToOne(_type => User)
+  @JoinColumn({ name: 'user_id' })
+  user!: User;
+
+  constructor(params: { userId: number; roleName: InstanceRoles }) {
```
Zeko369: This shouldn't be required if you have `if (params)`

server/models/UserChapterRole.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  @ManyToOne(_type => User)
+  @JoinColumn({ name: 'user_id' })
+  user!: User;
+
+  @ManyToOne(_type => User)
+  @JoinColumn({ name: 'chapter_id' })
+  chapter!: Chapter;
+
+  constructor(params: {
```
Zeko369: Same as the other one, make `params?:`

allella:
@timmyichen do we also need to generate a data/ddl.sql and the associated ER Diagram?

I installed DBeaver and it looks like you need a real database for it to create an ERD.  That is, it doesn't look like you can paste in the DDL and get a diagram.

I haven't ventured into installing Docker and I suspect it would take me awhile to get DBeaver talking to Postgres inside of the container, so I figured to ask.

We previously used an online tool for the ERD, but people liked DBeaver better and there was a bug in the tool we were using which didn't support one of the psql date types.  So, that's how we ended up on DBeaver.

timmyichen:
@allella we don't need to, but we can if you think it'll be helpful.  Either way I think it's best if not done in this PR as we don't want that to be blocking other people developing the other API routes.

allella:
@timmyichen  that's fine. I was somewhat wondering if the ddl.sql was generated automatically.  I think that would be nice and then someone can update the E-R Diagram since it's been a manual process.

server/models/UserInstanceRole.ts
timmyichen: The `if (params)` is required for the migration; errors without them (apparently the constructor is called with no arguments?).  Ideally we keep it so that the params is not optional so people writing actual TS code don't think it's optional when it's not.

vkWeb:
@timmyichen Thanks! Merging 🎉 

@allella IMO `ddl.sql` shouldn't be updated from now on to keep things simple. We have models, migrations, seeds in pace. We can now just keep on updating models as we progress. 

What do you say?

allella:
I found the ddl.sql useful and it was helpful to [create an ER-Diagram on dbdiagram.io](https://github.com/freeCodeCamp/chapter/issues/54#issuecomment-543486645) but we did move to DBeaver since there was a [timestamptz issue with dbdiagram.io](https://community.dbdiagram.io/t/timestamptz-not-null/354).

I've re-opened #54 because for documentation we need to at least update it.  If the ddl.sql can be updated automatically then I'd say to keep it around too.

I've pinged the original creators of both files on #54 
Thanks

PR #336, opened 2020-02-13
Bump node version

Zeko369:
Remove 1.0.0 from package.json, we're not there yet.
Upgrade node version to 10 -> 12 -> 13 (first commit was to test in CI)


package.json
```diff
@@ -125,7 +125,7 @@
     "singleQuote": true
   },
   "engines": {
-    "node": "10.x",
+    "node": "13.x",
```
timmyichen: should probably be more flexible with this and change to  >=

Zeko369: What do you mean by this? to do `>= 10.x` or something else?

timmyichen: Ah, nevermind, I misunderstood - this is why I shouldn't review PRs first thing after waking up :P LGTM!

PR #332, opened 2020-02-13
Add Swagger UI Express in place of speccy

AshNaz87:
- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `master` branch of Chapter.

Closes #325

PR #339, opened 2020-02-13
Standalone dev setup

Zeko369:
### Description
Some people (windows 10 32bit or Home, slower laptop running something other than linux) might not want to boot docker to run this. Also some dev already have `PSQL` installed in the background and running `ts-node index.ts` is a lot lighter than booting a linux vm to run another `PSQL` instance and node

THIS WON'T BE THE MAIN NOR THE RECOMMENDED WAY OF RUNNING CHAPTER,
but if you know what you're doing or have windows 10 32/home it's possibly a better dev experience compared.

### Todos
- [x] Add script to init the `.env` file or just yell at you if you don't have it
- [x] Add single `.env` file for centralized configs
- [x] Better example file
- [x] Remove `readENV.js`
- [x] Remove `console.log`s
- [x] Refactor package.json scripts
- [x] Remove commented out code
- [x] Add to docs
  - [x] `README.md`
- [x] Test on windows and linux
- [x] Config swaggerUIExpress
- [x] Remove speccy from docker-compose and package.json


Zeko369:
I also added postInstall script that copies the .env file so it doesn't break the env for people who didn't setup the .env yet
image
+ EMOJIS, crap I built yarn 😅

.env.example
```diff
@@ -0,0 +1,14 @@
+# If running a local setup make sure to set
+# If using docker this flag will cause it to crash
+# DB_LOCAL=true
+
+DB_USER=postgres
+DB_PASSWORD=password
+DB_NAME=chapter
+DB_URL=localhost
+DB_LOCAL=
```
vkWeb: We should replace `db_local` with `is_setup_docker` and the default value should be `true` since our default setup is docker. 

Then we should add a comment which will say `set this variable to false if you are running without docker`.

ormconfig.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const { config } = require('dotenv');
+
+config();
+
+const dbConfig = {
+  host: process.env.DB_URL,
+  port: process.env.DB_PORT || (process.env.DB_LOCAL !== '' ? 5432 : 54320),
```
vkWeb: If `is_setup_docker` is true use 54320 else 5432.

Zeko369: Yes, but I'd also keep the DB_PORT because I could have my local db on a different port or use a remote db. For example hosting it on a free online db as a service

.env.example
Zeko369: Agreed, that sounds better

Zeko369:
@timmyichen How would you implement this to work cross platform? My idea was running node scripts that would invoke docker-compose or just plain jest.
```json
{
  "test": "NODE_ENV=test docker-compose exec app npx jest --coverage --verbose",
  "test:ci": "NODE_ENV=test docker-compose exec -T app npx jest --coverage --verbose",
  "test:watch": "NODE_ENV=test docker-compose exec app npx jest --watchAll",
}
```

timmyichen:
`docker-compose` shouldn't be used in any tests, I'm not sure why I added that in haha.  We should just have it not use docker-compose and update the docs to instruct users to use docker-compose to run tests.

Zeko369:
Yeah, I wanted to say that, and since we're not using e2e we don't need it for now. But if we want to do e2e we'll need to setup it inside of docker (because of the DB requirement).

Also docker-compose tests (current) take about a minute and a half to run on my machine compared to 6seconds to run just the `npx jest...` 🤣

ormconfig.js
vkWeb: @Zeko369 Yes sure we should keep db_port.

server/index.ts
```diff
@@ -11,6 +12,15 @@ dotenv.config();
 
 const app: express.Application = express();
 
+// Make sure to kill the app if using non docker-compose setup and docker-compose
+if (isDocker && process.env.IS_DOCKER === '') {
+  console.error(
+    '\n\n\nUSING LOCAL DB BUT RUNNING IN DOCKER WILL CAUSE IT TO USE DOCKER-COMPOSE DB INSTAD OF LOCAL',
+  );
+  console.error('npm run typeorm WONT WORK PROPRELY\n\n\n');
```
ScottBrenner:
```suggestion
  console.error('npm run typeorm WON'T WORK PROPRELY\n\n\n');
```

ceciliaconsta3:
@Zeko369 Before merging this issue, can we resolve the remaining items on the checklist?
_WIP__Standalone_dev_setup_by_Zeko369_·_Pull_Request__339_·_freeCodeCamp_chapter

vkWeb:
@Zeko369 Are we ready to get this merged?

Zeko369:
Not yet, I'm still working the docs

vkWeb:
@Zeko369 From #332 we need to do:

- [ ] Configure `swagger-ui-express`

Zeko369:
@vkWeb I'll finish this in a couple hours, tnx for closing the other one.

Zeko369:
image

README.md
```diff
@@ -113,10 +136,7 @@ npm run test:watch
 
 We use [Open API 3.0](https://www.openapis.org/about) to define the API structure of the application. You can see the full documentation with:
```
timmyichen: nit: combine this line and the line below since we're no longer asking them to run a command

PR #340, opened 2020-02-15
Add locations CRUD api

Zeko369:
Add API for creating/reading/updating/deleting locations

- [x] Add filter by params
- [x] Add address to location modal
- [x] Update swagger docs

### DIFF
--- a/server/controllers/locationController.ts
+++ b/server/controllers/locationController.ts
@@ -1,0 +1,78 @@
+import { Request, Response } from 'express';
+import { Location } from 'server/models/Location';
+
+export default {
+  async index(_req: Request, res: Response) {
+    const locations = await Location.find();
+
+    res.json(locations);
+  },
+
+  async show(req: Request, res: Response) {
+    const { id } = req.params;
+
+    const location = await Location.findOne({ id: parseInt(id) });
+
+    if (location) {
+      res.json(location);
+    } else {
+      res.status(404).json({ error: "Can't find location" });
+    }
+  },
+
+  async create(req: Request, res: Response) {
+    const { country_code, city, region, postal_code } = req.body;
+
+    const location = new Location({
+      country_code,
+      city,
+      region,
+      postal_code,
+    });
+
+    try {
+      await location.save();
+      res.status(201).json(location);
+    } catch (e) {
+      res.status(500).json({ error: e });
+    }
+  },
+  async update(req: Request, res: Response) {
+    const { id } = req.params;
+    const { country_code, city, region, postal_code } = req.body;
+
+    const location = await Location.findOne({ id: parseInt(id) });
+
+    if (location) {
+      location.country_code = country_code ?? location.country_code;
+      location.city = city ?? location.city;
+      location.region = region ?? location.region;
+      location.postal_code = postal_code ?? location.postal_code;
+
+      try {
+        await location.save();
+        res.json(location);
+      } catch (e) {
+        res.status(500).json({ error: e });
+      }
+    } else {
+      res.status(404).json({ error: "Can't find location" });
+    }
+  },
+  async remove(req: Request, res: Response) {
+    const { id } = req.params;
+
+    const location = await Location.findOne({ id: parseInt(id) });
+
+    if (location) {
+      try {
+        await location.remove();
+        res.json({ id });
+      } catch (e) {
+        res.status(500).json({ error: e });
+      }
+    } else {
+      res.status(404).json({ error: "Can't find location" });
+    }
+  },
+};
--- a/server/controllers/locationController.ts
+++ b/server/controllers/locationController.ts
@@ -6,16 +6,19 @@ import { Location } from 'server/models/Location';
 
 export default {
   async index(req: Request, res: Response) {
-    const { country_code, city, region, postal_code, address } = req.query;
+    const fields = ['country_code', 'city', 'region', 'postal_code', 'address'];
+
+    const query = Object.keys(req.query)
+      .filter(key => fields.includes(key))
+      .filter(item => req.query[item].length > 0)
+      .map(item => {
+        return { [item]: Like(`%${req.query[item]}%`) };
+      });
+
+    console.log(query);
 
     const locations = await Location.find({
-      where: [
-        { country_code: Like(`%${country_code}%`) },
-        { city: Like(`%${city}%`) },
-        { region: Like(`%${region}%`) },
-        { postal_code: Like(`%${postal_code}%`) },
-        { address: Like(`%${address}%`) },
-      ],
+      where: query,
     });
 
     res.json(locations);
--- a/server/controllers/locationController.ts
+++ b/server/controllers/locationController.ts
@@ -21,13 +21,14 @@ export default {
   },
 
   async create(req: Request, res: Response) {
-    const { country_code, city, region, postal_code } = req.body;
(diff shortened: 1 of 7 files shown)


server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
+import { Request, Response } from 'express';
+import { Like } from 'typeorm';
+
+import { Location } from 'server/models/Location';
+
+export default {
+  async index(req: Request, res: Response) {
+    const { country_code, city, region, postal_code, address } = req.query;
```
timmyichen: We should validate these fields; e.g. if `city` is undefined it'll look for `%undefined%`.

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const locations = await Location.find({
+      where: [
+        { country_code: Like(`%${country_code}%`) },
```
timmyichen: Additionally, we should probably not have a `where` clause for any field that isn't included in the search query.

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+      ],
+    });
+
+    res.json(locations);
+  },
+
+  async show(req: Request, res: Response) {
+    const { id } = req.params;
+
+    const location = await Location.findOne({ id: parseInt(id) });
```
timmyichen: Are we not requiring radix for `parseInt` calls?

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    if (location) {
+      res.json(location);
+    } else {
+      res.status(404).json({ error: "Can't find location" });
+    }
+  },
+
+  async create(req: Request, res: Response) {
```
timmyichen: This should be a protected route (only users can create)

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  async show(req: Request, res: Response) {
+    const { id } = req.params;
+
+    const location = await Location.findOne({ id: parseInt(id) });
+
+    if (location) {
+      res.json(location);
+    } else {
+      res.status(404).json({ error: "Can't find location" });
```
timmyichen: Let's use `express-response-errors` instead of manually setting status here.

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+      region,
+      postal_code,
+      address,
+    });
+
+    try {
+      await location.save();
+      res.status(201).json(location);
+    } catch (e) {
+      res.status(500).json({ error: e });
```
timmyichen: Same re `express-response-errors`

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+  },
+  async update(req: Request, res: Response) {
```
timmyichen: Same re protected route.

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+    const { id } = req.params;
+    const { country_code, city, region, postal_code, address } = req.body;
+
+    const location = await Location.findOne({ id: parseInt(id) });
+
+    if (location) {
```
timmyichen: Instead of wrapping everything in this `if` block, it's cleaner to instead check `if (!location)` then return error, otherwise continue as normal.

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      if (address !== undefined) {
+        location.address = address;
+      }
+
+      try {
+        await location.save();
+        res.json(location);
+      } catch (e) {
+        res.status(500).json({ error: e });
```
timmyichen: `express-response-errors`

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+    } else {
+      res.status(404).json({ error: "Can't find location" });
```
timmyichen: `express-response-errors`

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+    const { id } = req.params;
+
+    const location = await Location.findOne({ id: parseInt(id) });
+
+    if (location) {
+      try {
+        await location.remove();
+        res.json({ id });
+      } catch (e) {
+        res.status(500).json({ error: e });
```
timmyichen: `express-response-errors`

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+    } else {
+      res.status(404).json({ error: "Can't find location" });
+    }
+  },
+  async remove(req: Request, res: Response) {
+    const { id } = req.params;
+
+    const location = await Location.findOne({ id: parseInt(id) });
+
+    if (location) {
```
timmyichen: same as above re ordering of logic

server/controllers/locationController.ts
```diff
@@ -0,0 +1,95 @@
...
+      try {
+        await location.remove();
+        res.json({ id });
+      } catch (e) {
+        res.status(500).json({ error: e });
+      }
+    } else {
+      res.status(404).json({ error: "Can't find location" });
```
timmyichen: `express-response-errors`

server/controllers/locationController.ts
Zeko369: Ok, I'll make it dynamic

server/controllers/locationController.ts
Zeko369: Do we really need it?

timmyichen: Not really, I'm more curious why it wasn't picked up by the linter (it wasn't excluded in the rules and I remember it being a default; maybe we're using a different base)

server/controllers/locationController.ts
Zeko369: We'll add this later when we create the auth strategy (cookies/jwt/etc...)

server/controllers/locationController.ts
Zeko369: I don't know lol, but I don't think we should add it here, because it's a kind of verbose and unnecessary

server/controllers/locationController.ts
timmyichen: Doesn't passport handle it all the same way anyway regardless of strategy? By adding `req.user` to the request?

Zeko369: Yes it does, but I don't think we should add passport in this PR, maybe better in a separate PR. Also we need to do the same for other API routes, so better to do it in a different PR

timmyichen: I agree that we shouldn't add passport in this PR, but it would still be good to implement.  I'll address changes in #344 and once that's merged we can implement it here and test it accordingly.  No point in omitting it since we're not building the frontend for it anyway.

In any case, we should probably prioritize getting passport in since so many things depend on it.

Zeko369: Yeah we are, I'm doing it parallel in #342, So I'd like to get this merge so I can work on that, #341 depends on this and #342 and I'd like to get this done before tomorrow's meeting

timmyichen: Ah didn't realize you were building it.  Fine by me to defer this then!

PR #342, opened 2020-02-16
Add locations CRUD UI

Zeko369:
Add UI for #340


client/store/actions/locations.ts
```diff
@@ -0,0 +1,213 @@
...
+export const remove: ActionCreator<locationsTypes.ThunkResult<
+  Promise<void>
+>> = id => async dispatch => {
+  dispatch(deleteStart());
+
+  const http = new HttpService<{ id: string }>();
+  try {
+    const resData = await http.delete(`/locations/${id}`, {}, {});
+
+    console.log(resData);
```
timmyichen: remove console log

client/store/actions/locations.ts
```diff
@@ -0,0 +1,213 @@
...
+};
+
+export const create: ActionCreator<locationsTypes.ThunkResult<
+  Promise<void>
+>> = data => async dispatch => {
+  dispatch(createStart());
+
+  const http = new HttpService<locationsTypes.ILocationModal>();
+  try {
+    console.log(data);
```
timmyichen: remove console log

client/store/types/actions/locations.ts
```diff
@@ -0,0 +1,19 @@
+export const FETCH_START = 'fcc/chapter/LOCATIONS_START';
```
timmyichen: these probably aren't necessary with typescript, but it doesn't matter for now

client/store/types/locations.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+interface ILocationDeleteStartAction {
+  type: typeof ACTIONS.DELETE_START;
+}
+
+interface ILocationDeleteSuccessAction {
+  type: typeof ACTIONS.DELETE_SUCCESS;
+  payload: {
+    id: string;
```
timmyichen: why is this ID a string whereas location IDs are numbers?

pages/dashboard/locations/[id]/edit.tsx
```diff
@@ -0,0 +1,63 @@
...
+    await dispatch(
+      locationActions.updateLocation(
+        parseInt(Array.isArray(id) ? id[0] : id),
+        data,
+      ),
+    );
+    router.replace('/dashboard/locations');
+  };
+
+  return loading ? (
```
Comment (timmyichen) [inline pages/dashboard/locations/[id]/edit.tsx]: let's avoid nested ternary statements; they're difficult to parse

pages/dashboard/locations/[id]/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  }));
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (id !== undefined) {
+      dispatch(locationActions.fetchOneLocation(id));
+    }
+  }, [id]);
+
+  return loading ? (
```
Comment (timmyichen) [inline pages/dashboard/locations/[id]/index.tsx]: same re nested ternary

pages/dashboard/locations/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }));
+
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(locationActions.fetchLocations());
+  }, []);
+
+  return (
+    <>
```
timmyichen: `<>` wrapper not necessary

pages/dashboard/locations/[id]/edit.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }, [id]);
+
+  const onSubmit = async data => {
+    await dispatch(
+      locationActions.updateLocation(
+        parseInt(Array.isArray(id) ? id[0] : id),
+        data,
+      ),
+    );
+    router.replace('/dashboard/locations');
```
Comment (timmyichen) [inline pages/dashboard/locations/[id]/edit.tsx]: why replace instead of push?

pages/dashboard/locations/new.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  const { error, loading } = useSelector((state: AppStoreState) => ({
+    error: state.locations.create.error,
+    loading: state.locations.create.loading,
+  }));
+  const dispatch = useDispatch();
+
+  const onSubmit = async data => {
+    await dispatch(locationActions.create(data));
+    router.replace('/dashboard/locations');
```
timmyichen: same re replace vs push.  push is generally preferred.

client/components/DashboardLocation.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const styles = useStyles();
+
+  const remove = () => {
+    dispatch(locationActions.remove(location.id));
+  };
+
+  return (
+    <Card style={{ marginTop: '12px' }}>
+      <ProgressCardContent loading={loading}>
+        <MULink gutterBottom variant="h5" component="h2">
```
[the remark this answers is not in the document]
Zeko369: OH CRAP, that could by why it was doing the stupid redirect 🤦‍♂

client/store/types/actions/locations.ts
Zeko369: Why not? They're simpler to use because you can easily autocomplete them + less to type

pages/dashboard/locations/[id]/edit.tsx
Comment (Zeko369) [reply pages/dashboard/locations/[id]/edit.tsx]: Hmm, here we could have a push. But this was copied from the create action to which you don't want to go back to, unless you want to create multiple of the same which you most likely don't want to do

pages/dashboard/locations/[id]/edit.tsx
Comment (Zeko369) [reply pages/dashboard/locations/[id]/edit.tsx]: I'll extract this into separate returns

pages/dashboard/locations/new.tsx
Zeko369: Yes, but here I'd argue that we should keep replace because when you create something you don't want to go back to the new page. Or do we maybe? (longer explanation in a comment before)

PR #343, opened 2020-02-17
Added Location, users, chapter CRUD endpoints with Tests

joshghent:

Closes #XXXXX


Zeko369:
HEY @joshghent welcome to the community. I saw your fork on the network tab and asked Quincy to email you. For next time, could you open the PR first earlier so I don't rework some of you code 🙈.

I'll be really happy to review you code in an couple of hours (I have some things to do at work), but wanted to personally welcome you, because of the work you did.

Also could we maybe first merge my locations API PR #340 , and then yours because I added the whole CRUD set, and use it for the frontend PR.

Some other things that I saw last time I checked out your PR branch.
- We don't need babel plugins for `??` and `?.`, because TS3.7+ does that for us
- Also maybe extract user controllers into a separate PR, because of the OAuth tokens and other things (at least the create route)
- Also just a nitpick but working on `master` isn't really recommended because then you can't really easily sync to `upstream/master`

PS, I can't seem to find you in discord?, could you change your username there if you've already joined, or join if you haven't?

.babelrc
```diff
@@ -3,6 +3,16 @@
     "next/babel"
   ],
   "plugins": [
+    [
```
Zeko369: As I mentioned before, TS3.7+ already has `??` and `.?`, so this isn't required

server/controllers/locationController.ts
```diff
@@ -0,0 +1,29 @@
...
+    });
+
+    try {
+      await location.save();
+      res.status(201).json(location);
+    } catch (e) {
+      res.status(500).json({ error: e });
+    }
+  },
+};
```
Zeko369: Let's first merge my PR because I already implemented the whole CRUD pattern #340

server/routes/tests/chapterRouter.test.ts
```diff
@@ -0,0 +1,62 @@
...
+describe('router: sampleRouter', () => {
+  const app = new App();
+
+  beforeAll(async () => {
+    await app.initialize({ withRouter: chapterRouter });
+  });
+
+  describe('/chapter', () => {
+    it('Creates a Chapter', async () => {
+      const location = await app.request.post('/locations').send({
```
Zeko369: We need to setup testing DB or maybe mock TypeORM for this. @timmyichen any ideas?

server/controllers/userChapterController.ts
```diff
@@ -0,0 +1,47 @@
+import { Request, Response } from 'express';
+import { Chapter, User } from 'server/models';
+import { UserChapter } from 'server/models/UserChapter';
+
+// The whole model is a json response, fix that if there's some sensitive data here
+
+export default {
+  async create(req: Request, res: Response) {
```
Zeko369: This is more of a `join` action, since here a user joins a chapter (event though we create a new userChapter object)

server/routes/tests/chapterRouter.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+
+    it('Adds a user to the chapter', async () => {
+      const res = await app.request.get('/chapters/1/join/1');
+
+      expect(res.status).toBe(201);
+    });
+
+    it('Bans a user from a chapter', async () => {
+      const res = await app.request.get('/chapters/1/ban/1');
```
Zeko369: This is a POST not a GET

server/routes/tests/chapterRouter.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        description: 'Test',
+        category: 'Test',
+        details: 'Hello World',
+        creator: 1,
+      });
+      expect(res.status).toBe(201);
+    });
+
+    it('Adds a user to the chapter', async () => {
+      const res = await app.request.get('/chapters/1/join/1');
```
Zeko369: This is a POST not a GET

server/routes/v1/chapter.ts
```diff
@@ -11,4 +12,13 @@ router.get('/chapters/:id', asyncHandler(chapterController.show));
 router.patch('/chapters/:id', asyncHandler(chapterController.update));
 router.delete('/chapters/:id', asyncHandler(chapterController.remove));
 
+router.post(
+  '/chapters/:id/join/:user_id',
+  asyncHandler(userChapterController.create),
+);
+router.post(
```
Zeko369: Maybe change this to DELETE, just for the semantics of REST

server/routes/tests/chapterRouter.test.ts
Zeko369: Also here you wouldn't put userId, rather you'd extract that from the request cookie

server/routes/tests/chapterRouter.test.ts
timmyichen: We should just have a bunch of helper functions (factories?) to aid with the creation of models we need.

Zeko369: Yeah, but there must be an function to remove data after a test

timmyichen: tests data should be spun up in a test database anyway (which is nuked/recreated when running tests), is this configurable by typeorm?

Zeko369: I was also thinking about (this)[https://gist.github.com/Zeko369/2e81e78d1eca03ca35a5a1fc7d63b222] this is a `rspec` test (for rails controller) where I can
1. easily create entities at the beginning (also there is a factory linked)
2. can check what changed in the db during a (request)[https://gist.github.com/Zeko369/2e81e78d1eca03ca35a5a1fc7d63b222]

I'll look into what can be done with `typeorm` in that regard

server/controllers/userController.ts
```diff
@@ -0,0 +1,27 @@
+import { Request, Response } from 'express';
```
Zeko369: I'll remove this for now because the actual user creation will work differently later

server/routes/tests/chapterRouter.test.ts
Zeko369: Resolving this, it'll be done in a different PR

server/routes/v1/chapter.ts
Zeko369: It's fine for now

Zeko369:
@timmyichen The Lib you suggested for HTTP errors returns HTML on that error, so I'll keep the current `res.status(404).json(...)` and later we can use a fancy lib for that

PR #344, opened 2020-02-18
testing utils for users

timmyichen:

This PR allows us to mock users for testing.  For example, a test might say:
```javascript
const app = new App();

// beforeAll
await app.initialize({ withRouter: routerWithAuthedRoutes });

// in test
app.login(someUser);
app.request.get('/blahblah');
expect('something').toBe('something');
app.logout(someUser);
```

### DIFF
--- a/types/.eslintrc
+++ b/types/.eslintrc
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "@typescript-eslint/interface-name-prefix": "off"
+  }
+}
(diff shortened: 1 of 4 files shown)


types/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "@typescript-eslint/interface-name-prefix": "off"
```
Zeko369: I would just disable it for the `expressUser.d.ts` not generally

Zeko369:
Seems good, but since we have example controllers/routers could you add a route that return 200/401/403 respectively?, so pretty much a bigger example of your PR description

This will be done in a separate PR

Issue #360, opened 2020-03-06
Documentation: Consolidate Technical Documentation


allella:
Quincy suggested all [technical docs go into CONTRIBUTING.md on the Jan 25 2020 meeting](https://github.com/freeCodeCamp/chapter/wiki/Meeting-Notes-January-25,-2020#re-orientation-and-design-overview).

- [x] 1. Move https://github.com/freeCodeCamp/chapter/tree/master/server/docs to CONTRIBUTING.md
- [x] 2. (edited) Move https://github.com/freeCodeCamp/chapter/blob/master/docs/how-to-open-a-pull-request.md into an collapsed / expandable section within [Step 4 of the CONTRIBUTING.md](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md#step-4-proposing-a-pull-request-pr) 
- [x] 3. We can delete /docs/ directories, except /images since it's only general stubbed out stuff from really early on.
- [x] 4. Is there a better place to store CONTRIBUTING.md screenshots than in /docs/images?  If not, then do we want to rename it.
- [x] 5. As with the "Step 4" note above, we can / should use expandable sections in CONTRIBUTING.md to make certain sections not take up as much space until someone clicks to expand.
- [x] 6. @Narutuffy was looking to get involved and I thought it could be an easy first contribution 
- [x] 7. has Docker setup and is more familiar with the stack than I am, so he may be able to add other tips or improvements to CONTRIBUTING.md on the technical side. - Closed via #396 
- [x] 8. This also begs the question of if some of the README.md technical documentation is better in CONTRIBUTING.md. These sections of the README may be better in CONTRIBUTING.md: [Schema](https://github.com/freeCodeCamp/chapter/blob/master/README.md#schema), [API Specs](https://github.com/freeCodeCamp/chapter/blob/master/README.md#api-specification
), [Testing](https://github.com/freeCodeCamp/chapter/blob/master/README.md#testing) - Closed via #397

allella:
The /docs/images and docs/how-to-open-a-pull-request.md are used or linked to from within the CONTRIBUTING.md, so we don't want to delete them without consolidating.

Quincy had suggesting making one long technical document and using the Wiki for user docs.

I figure we can use the collapsed sections in CONTRIBUTING.md to make it more usable despite the length.

allella:
This also begs the question of if some of the README.md technical documentation is better in CONTRIBUTING.md.

[...]

PR #361, opened 2020-03-12
remove user chapter, replace with user chapter role

timmyichen:

This PR:
* Removes the `UserChapter` model
* Adds factories/seeds/etc for `UserChapterRole` model

Strangely enough, I can't seem to get a migration running for this.  Any ideas?

```
tim@dropten:~/chapter$ npm run typeorm migration:generate -- --name=RemoveChapterUser

> chapter@0.0.1 typeorm /home/tim/chapter
> ts-node -P tsconfig.server.json -r tsconfig-paths/register node_modules/typeorm/cli.js "migration:generate" "--name=RemoveChapterUser"

No changes in database schema were found - cannot generate a migration. To create a new empty migration use "typeorm migration:create" command
```


Zeko369:
I think I know that the problem is. TypeORM is really finicky when it comes to changes relations. So when you change the migration table it doesn't actually change the migration.

Try creating a new migration for a slightly different column eg `users123`, run the migrations, and then manually change stuff around in the migration file.

allella:
To summarize for myself, and for anybody else watching, it sounds like these changes are furthering the work of #335 and #345 and endeavoring to create a migration to make the schema changes which were previously only part of the models? Is that correct?

Regarding automatically updating the schema image, we have that in progress on #54 and https://github.com/freeCodeCamp/chapter/issues/360#issuecomment-598486978

Zeko369:
Any progress with this @timmyichen?

server/models/UserChapterRole.ts
```diff
@@ -23,15 +24,25 @@ export class UserChapterRole {
   @JoinColumn({ name: 'chapter_id' })
   chapter!: Chapter;
 
+  /*
+    This indicates whether he use wants to receive notifications about this chapter.
```
Zeko369, line 28: Maybe rename this to notifications?
This sounds a lot like someone wanting to join a chapter

Zeko369: Honestly this is fine as is, we can change it later if others find it confusing

PR #402, opened 2020-09-14
remove yarn.lock; changed README.md to include `npm install` for client

paulywill:
- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `master` branch of Chapter.

----

Troubleshooting errors with `docker-compose up` on OSX (10.13.6)
someone in the Discord mentioned to resolve this you have to `cd client && npm install`

Added the missing instruction to the README.md

Instead of individual commands, I strung them together with `&&`.:

`npm install && cd client && npm install && cd ..`

Since quite of bit of the README kept indicating using `npm install`, I removed the yarn.lock file and instead included the package-lock.json

PR #403, opened 2020-09-15
Document: changed README.md to include npm install for client

paulywill:

Changed PR from @pdotsani suggestions in  #402 

This change includes an inline comment next to the `npm install` for a description.
Tested locally and inline comment is ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,9 +81,7 @@ cd chapter
 
 Install dependencies:
 ```
-npm install
-cd ./client
-npm install `#install client dependencies`
+npx recursive-install
 ```
 
 If you're using local setup (no docker), make sure you add your DB credentials to .env file
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ cd chapter
 Install dependencies:
 ```
 npm install
-cd client
+cd ./client
 npm install `#install client dependencies`
 ```
 
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ cd chapter
 Install dependencies:
 ```
 npm install
+cd client
+npm install `#install client dependencies`
 ```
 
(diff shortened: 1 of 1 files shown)


allella:
Thanks @paulywill is this change specific to any operating system or environment, or should we consider it an improvement for all situation?

paulywill:
@allella It did resolve to start `docker-compose up` on my OSX.

However, I suspect it would be an improvement in all situations.

allella:
@Zeko369 do you see any downside to this documentation change? 

We've had a couple contributors suggest a 

`cd client`
improved their installation.

README.md
```diff
@@ -82,6 +82,8 @@ cd chapter
 Install dependencies:
 ```
 npm install
+cd client
```
allella, line 85: Not sure it matters, but tomn on Discord used `cd ./client`

As far as I know they are the equivalent, but if there's an OS-specific reason to include the period, then we could consider it.

paulywill: Agree. Made the suggested change.

timmyichen (commented):
Thanks for adding this! Do you think it's more semantically clear if we just install things recursively, e.g. with:

```
npx recursive-install
```

Thoughts?

paulywill:
Wow! I had no idea that command existed.

paulywill:
@timmyichen I have tested locally and it appears to accomplish the same thing... always allows flexibility and not hardcoding to structure of directories (a minor concern but valid).

I will make the change to the PR.

timmyichen (approved):
:tada

allella:
Nice. Teamwork makes the dream work.

allella:
@all-contributors please add @paulywill for docs.

tomnoland:
Didn't know about that either. Looks good to me

PR #405, opened 2020-09-22
Created a Messages controller and a basic resolver for sending emails

megajon:

Closes #XXXXX

I created a new Messages controller and a basic resolver with a mutation for sending emails, but when I try to run the mutation in  the graphql playground I get an error "Cannot return null for non-nullable field Email.to." when I submit the recipient email as a string. When I submit the recipient email an array I get the error "String cannot represent a non string value". The server runs fine and there are no errors in the console. I've attached screenshots of both errors. Any advice on what I'm doing wrong would be appreciated. 

arrayError

stringError


server/controllers/Messages/inputs.ts
```diff
@@ -0,0 +1,13 @@
+import { InputType, Field } from 'type-graphql';
+
+@InputType()
+export class SendEmailInputs {
+  @Field(() => String)
```
Zeko369, line 5:
```suggestion
  @Field(() => [String])
```

Zeko369: You forgot an array here

server/models/Email.ts
```diff
@@ -0,0 +1,13 @@
...
+export class Email {
+  @Field(() => String)
+  to: string[];
+
+  @Field(() => String)
+  subject: string;
+
+  @Field(() => String)
+  html: string;
+}
```
Zeko369, line 13: Since this is only used in the mailer controller, I'd move it in the `server/controllers/Messages`

Zeko369 (changes requested):
I think that the problem was you forgot an array around [String] (yeah I know, the API isn't as simple as it seams at first)

server/models/Email.ts
megajon: As you suggested I changed the field type in Email.ts to an array of strings. I also moved the Email.ts file to the Messages folder. I was still getting a graphql error when I ran the mutation in the playground, but the messages were still being sent because I received all of them in my nodeMailer App. If I return only the subject from the mutation then it works fine.

emailSent

Zeko369: The problem is that unlike REST, Graphql does runtime type checking, so if you say that `to` is going to be an array of strings and returns null/undefind it's going to error out. Check what is really returned from that Mailer service and change the types accordingly.

megajon: I was able to successfully return the desired data by changing the names of the fields in Email.ts to match the properties returned by the MailerService. Not sure if that's the best way to do it, but it seemed simpler than changing anything in the MailerService.

successReturn

Zeko369:
@megajon  Why was this closed?

megajon:
@Zeko369 There was no response from my last comment, so I didn't see any reason to keep it open. Should I reopen the PR?

allella:
@megajon I suspect anything that works and/or keeps the conversation going is better than discarding work.

I'm not going to offer advise on the framework implementation, but I do have experience and opinions about using SMTP with the 3rd party services. I've shared those earlier comments, but I'm glad to chime in if there are questions on such things. 

For the most part, I suspect we'll only support the SMTP in nodemailer and this would be configurable in .env as part of mailer settings.  Most, if not all, 3rd party email services will support SMTP and relying on allows folks who have their own localhost relay to use their own services. All of this avoids having to enable and thus support other specific plugins for nodemailer, which will only seem to complicate things vs using just SMTP (at least for the MVP).

allella:
We don't need to hijack this PR with the .env conversation, but here are [nodemailer's SMTP](https://nodemailer.com/smtp/) options, which would mostly be set in the .env and stubbed out in .env.example.

allella:
Also, I see the code above using [sendmail](https://www.npmjs.com/package/sendmail), which looks like part of NodeJS, but not the [nodemailer](https://www.npmjs.com/package/nodemailer) package we were targeting for the MVP.

megajon:
@allella Which line of code are you specifically referring to.

allella:
@megajon I was looking at the screenshots and thought it was sendmail, but I see the method is sendEmail. Plus, there's an import nodemailer in the code, so looks like you're using that package.

I'm not a Node dev so sorry for going out of my realm there.  I do think the .env pattern is one we'd use for the mail configuration, so I think that part is still relevant for the SMTP hostname, port, and such. 

I'm not sure if we've talked about where we'd save the admin "To:" email address.  I suspect that could be the email address of the _organizer_ of the current _chapter_.  This is likely going to come from the [user_chapter_role table] (https://github.com/freeCodeCamp/chapter/blob/master/server/models/UserChapterRole.ts). 

The schema image is still out of date and [we've talked about automating it](#54 ), but for now that part of the database is the likely origin of a "To:" email address.

The "From:" email address may be something we set in the .env since there are email validation considerations for the from email and domain, such as SPF, DKIM, and DMARC validation. 

The "Reply To:" email may end up being a "no-reply@" or perhaps it's also the _organizer_'s email, but we'll figure that out with more context of the specific emails.

allella:
@megajon the Reply-to could potentially be the organizaer's email, but the from email shouldn't be the organizer.  This is due to modern email validation with SPF, DKIM, and DMARC.  If we're sending emails "From:" any random address the emails will end up in spam or even rejected (via DMARC) because there are no DNS policies setup to validate the Chapter's instance as a valid sender.

The only situation where this would work is if everyone using the instance had an email address at the primary sending domain, but that's something we can't enforce.  

We'll to assume emails from a validated domain, even if the email address at that domain is a no-reply@

allella:
I'm not even sure we'll be able to support localhost email relaying, so that's something we might also need to document or restrict.

I doubt the Docker instance / hosting server will be running a full mail relay daemon, like Postfix.  Plus, even if we installed one it's rather likely folks wouldn't setup the correct SPF and DKIM rules and emails originating from the _instance_ would be very likely to end up in spam.

I'm going to update the FAQs to indicate we don't suggest / support localhost relaying, even though nodemailer may allow it out of the box. https://github.com/freeCodeCamp/chapter/wiki/Owner-FAQs

allella:
Localhost will still be technically allowed, and it may be useful for some development environments, but there's a lot of variations in daemons and configurations for things like DKIM.

Anybody who wants to use localhost in production could figure it out, but most sane people use a remote mail relay or SMTP.

PR #407, opened 2020-09-24
Feat/ initalize client tests

paulywill:

Attempting to initialize and add some client tests.

Require some eye and expertise to resolve configuration setup.

Currently receiving the following error for a simple client test:

```
 ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/paulgamble/Coding/fcc/chapter/client/tests/PageLayout.test.ts:1
    import SomethingObject from '../src/components';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      at Runtime._execModule (node_modules/jest-runtime/build/index.js:1157:58)
```


allella:
@paizle is this issue relevant to your testing bug posted on the chat?

paizle:
I don't think so. My output looks like this (from the root Chapter dir):
```
$ npm run test

> npx jest --coverage --verbose

TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at exports.pathsToModuleNameMapper 
```

ojeytonwilliams:
Hi @paulywill, when I fixed the conflict, I tried to keep everything as close to how you had it as possible.  The only change being a slight tweak to `jest.config.js`.  I hope that's okay.

I've give this a proper review as soon as I can.

ojeytonwilliams:
After a bit of digging, I discovered that switching from `.babelrc` to `babel.config.js` magically fixed things.  I don't understand *why* this mattered, but it fixed the import error without the need for a second config file.

paulywill:
> After a bit of digging, I discovered that switching from `.babelrc` to `babel.config.js` magically fixed things. I don't understand _why_ this mattered, but it fixed the import error without the need for a second config file.

Just getting back into this.... for my clarity you removed `.babelrc`?

paulywill:
Will open another PR for the client tests.

Issue #349, opened 2020-02-23
As a user, I can search using the search bar in the navigation (implemented with PostgreSQL queries)


QuincyLarson:
As a user I should be able to search through through chapters.

- [ ] Mock-up search bar in Material UI (#414) as seen in [homepage mock-up](https://www.figma.com/proto/q7DikyL3N0c4CUWxHNa97i/Chapter-Prototype?node-id=1%3A2&scaling=scale-down)
- [ ] Implement results query or API call
- [ ] Implement results view with lazy load, or infinity scroll, or pagination

allella:
The backend logic of this has been discussed on #112

xarielx:
@Zeko369 please let me know if you need help setting this up, are you planning to use Axios? We can coordinate at any time after 5 pm eastern time if you need help. Thanks

Zeko369:
We're not using axios, we have our own server/client [implementation](https://github.com/freeCodeCamp/chapter/blob/master/client/services/http-service.ts) with `cross-fetch`.

This should be a part of `eventsQuery` on homepage, so you would actually want to create a redux store for it and everything.
But for simplicity you can just use `cross-fetch` and useState to query some events according to the search param for example `api/v1/events?serach=FooBar`

This is more of a API issue, I'll handle it in a today, but for now you can stub it.

xarielx:
That makes perfect sense. I was including useState into waft I was coding but wasn’t sure if we were tying it to the pre loaded events or all the ones saved on the back end

UF ID: 0903-9139<tel:0903-9139>


On Feb 27, 2020, at 3:55 AM, Fran Zekan <notifications@github.com> wrote:

﻿
[External Email]

We're not using axios, we have our own server/client implementation<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_freeCodeCamp_chapter_blob_master_client_services_http-2Dservice.ts&d=DwMCaQ&c=sJ6xIWYx-zLMB3EPkvcnVg&r=RjXmsQnc1Nqs3ON4MiC_jP_ja-8LDMlGR5rK5BQhg4k&m=XbcwwTc8atwTSlIW4x2GD9v8m-iwkMdkgKPSYMTIfwM&s=XRPAm78ol94L7RYfKOBW6cXx7-XfXqCdGsr8_N0XhRo&e=> with cross-fetch.

This should be a part of eventsQuery on homepage, so you would actually want to create a redux store for it and everything.
[...]

PR #410, opened 2020-10-11
Create dummy SERP.

Ryuno-Ki:
Signed-off-by: André Jaenisch <andre.jaenisch@posteo.de>

Closes #XXXXX

This is the next iteration after #409. The next PR will make the search bar go to the Search Result Page (a.k.a. SERP).
You can reach it via `/search-results` (we can make it `/search` if you want).
What's missing is the backend functionality (i.e. #349).

### DIFF
--- a/client/src/modules/home/shared/components/Layout.tsx
+++ b/client/src/modules/home/shared/components/Layout.tsx
@@ -1,0 +1,7 @@
+import React from 'react';
+
+const Layout: React.FC = ({ children }) => {
+  return <>{children}</>;
+};
+
+export default Layout;
--- a/client/src/modules/search-results/index.tsx
+++ b/client/src/modules/search-results/index.tsx
@@ -12,7 +12,7 @@ const SearchResults: React.FC = () => {
 
   return (
     <Layout>
-      <Typography variant="h1" component="h1">
+      <Typography variant="h4" component="h1">
         Search results for <q>{data?.searchTerm}</q>
       </Typography>
       <Grid container spacing={2}>
--- a/client/src/modules/search-results/index.tsx
+++ b/client/src/modules/search-results/index.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import { Card, Grid, Typography } from '@material-ui/core';
+import Layout from '../home/shared/components/Layout';
+
+const SearchResults: React.FC = () => {
+  // TODO: Replace with `useEventsQuery()`!
+  const data = {
+    searchResults: [],
+    searchTerm: 'London',
+  };
+
+  return (
+    <Layout>
+      <h1>
+        Search results for <q>{data?.searchTerm}</q>
+      </h1>
+      <Grid container spacing={2}>
+        <Grid item xs={10}>
+          {data?.searchResults.length === 0 && (
+            <Typography gutterButtom variant="p" component="p">
+              Nothing found.
+            </Typography>
+          )}
+          {data?.searchResults.map(searchResult => (
+            <Card>
+              <Typography gutterBottom variant="h5" component="h2">
+                {searchResult}
+              </Typography>
+            </Card>
+          ))}
+        </Grid>
+      </Grid>
+    </Layout>
+  );
+};
+
+export default SearchResults;
--- a/client/src/modules/search-results/index.tsx
+++ b/client/src/modules/search-results/index.tsx
@@ -4,6 +4,7 @@ import Layout from '../home/shared/components/Layout';
 
 const SearchResults: React.FC = () => {
   // TODO: Replace with `useEventsQuery()`!
+  // TODO: Add `key` prop to `<Card />`!
   const data = {
     searchResults: [],
     searchTerm: 'London',
@@ -11,13 +12,13 @@ const SearchResults: React.FC = () => {
 
   return (
     <Layout>
-      <h1>
+      <Typography variant="h1" component="h1">
         Search results for <q>{data?.searchTerm}</q>
-      </h1>
+      </Typography>
       <Grid container spacing={2}>
         <Grid item xs={10}>
           {data?.searchResults.length === 0 && (
-            <Typography gutterButtom variant="p" component="p">
+            <Typography variant="body1" component="p">
               Nothing found.
             </Typography>
           )}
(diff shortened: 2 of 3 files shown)


client/src/modules/search-results/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <Layout>
+      <h1>
+        Search results for <q>{data?.searchTerm}</q>
+      </h1>
+      <Grid container spacing={2}>
+        <Grid item xs={10}>
+          {data?.searchResults.length === 0 && (
+            <Typography gutterButtom variant="p" component="p">
```
vkWeb, line 20: @Ryuno-Ki Two warnings on build: 

1. `gutterButtom` (misspelled). We can use `paragraph` instead...?
2. `variant` doesn't accept `p` as value as per Material API. `component=p` alone might do the job. 

@timmyichen would you mind reviewing this...? 

I just did the build to see how it looks.

vkWeb (changes requested):
Warnings on build.

client/src/modules/search-results/index.tsx
Ryuno-Ki: Note to self: https://material-ui.com/components/typography/#changing-the-semantic-element

vkWeb: @Ryuno-Ki what's the point of assigning the variant to be a paragraph if the component is itself a paragraph...? 

According to my understanding of that doc, `variant` prop applies styling of the passed value. `component` prop makes the node of that kind. In our case variant and component are same. 

For situations where we need a `<h2>` and the styles of `<h1>` in `<h2>`, variant might be of help. But here simply passing `p` to component will do the job.

client/src/modules/home/shared/components/Layout.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const Layout: React.FC = ({ children }) => {
```
timmyichen, line 3: What is the purpose of this component?

client/src/modules/search-results/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+const SearchResults: React.FC = () => {
+  // TODO: Replace with `useEventsQuery()`!
+  const data = {
+    searchResults: [],
+    searchTerm: 'London',
+  };
+
+  return (
+    <Layout>
+      <h1>
```
timmyichen, line 14: If we're going with MUI components, this should be a Typography component with variant/component h1

client/src/modules/search-results/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+      </h1>
+      <Grid container spacing={2}>
+        <Grid item xs={10}>
+          {data?.searchResults.length === 0 && (
+            <Typography gutterButtom variant="p" component="p">
+              Nothing found.
+            </Typography>
+          )}
+          {data?.searchResults.map(searchResult => (
+            <Card>
```
timmyichen, line 25: We should have a `key` prop on this component

client/src/modules/search-results/index.tsx
timmyichen: List of variants: https://material-ui.com/customization/typography/#variants

timmyichen (commented):
Left a few comments/questions. Please ping me if anything isn't clear!

client/src/modules/home/shared/components/Layout.tsx
Ryuno-Ki: I was looking at how [`client/src/modules/dashboard/shared/components/Layout.tsx`](https://github.com/freeCodeCamp/chapter/blob/8f836a934ffccabb6837cd5f4989292d0687e104/client/src/modules/dashboard/shared/components/Layout.tsx) was rendered, but didn't want to have the Navbar.

client/src/modules/search-results/index.tsx
Ryuno-Ki: Will do.

client/src/modules/search-results/index.tsx
Ryuno-Ki: Do you have suggestions on its value? Right now, there aren't any real search results.
I'd need to wire it up with GraphQL (and then could look into adding a `key` prop).

client/src/modules/home/shared/components/Layout.tsx
timmyichen: If that's the case, we can do without this component. It can be added in when it's needed, otherwise right now it's not doing anything. Why do you not want the nav rendered though?

client/src/modules/search-results/index.tsx
timmyichen: Ah, just realized the result array is empty. I'd add a to-do here then to make sure it doesn't get left behind.

timmyichen:
On second thought, this should probably be on a feature branch if you're going to break it up into different PRs. I'll make one for you tonight.

Ryuno-Ki:
@timmyichen Remember https://github.com/freeCodeCamp/chapter/pull/409#issue-500453101 ?
It will be one or two more PRs until the search is implemented: Linking search bar to SERP + making a query to the backend and display results.

timmyichen:
Yes, I remember, but we shouldn't be merging half-completed features into master.  Ideally they should be merged into a feature branch first, then the feature branch can be merged into master once the feature is complete.

timmyichen:
Please rebase to branch `feat/search`!

Ryuno-Ki:
Rebased and addressed review comments.

vkWeb (approved):
Warnings are gone. Build OK.

Maybe we can use variant ~~`h4`~~ `h3` on `Search results for...` to make it look not too big.

client/src/modules/home/shared/components/Layout.tsx
timmyichen: Approving assuming this file will become something later on.

timmyichen (approved):

vkWeb:
Since Tim has approved. I'm merging this in. :tada:

allella:
@all-contributors please add @Ryuno-Ki for code

Ryuno-Ki:
Finally!

I think, I can work on the next iteration earliest in two days.

Issue #262, opened 2019-11-30
Documentation for Accounts and Notifications


QuincyLarson:
**Account & Notifications**
- [ ] Accessing an account
- [ ] Resetting your password
- [ ] Organizer subscriptions
- [ ] Notification settings

allella:
@chrismgonzalez 

- If the MVP is using Google for authentication, then is documenting an account page or password reset still applicable? I guess we could document the fact that your email and password are managed by Google, so visit https://myaccount.google.com or https://myaccount.google.com/security "Signing in to other sites" section? 
- The MVP hasn't mentioned an account page, but if we do have one then can we imagine is needed?  #274 and [the suggestions about subscribing and unsubscribing may avoid the need for an account page](https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-562374895) to managed subscriptions, at least for MVP.
- What are we thinking with "Organizer subscriptions"? If it's for the organizer, then we probably need to move that item to #260.  Or, was this intended to be "Organize Subscriptions"?

QuincyLarson:
@allella These are great points. I've updated the bullet points above to reflect this by removing the need for password management.

We should have a profile/settings page, and the ability to make it private. If you can document what specific attributes the page should have, we can ask @madaleneaza-design if she can create a layout for this.

allella:
I would agree that a basic profile management page is necessary IF any of the following are true:
- [we're not pulling a _user's_ name from Google's profile or we use the Google name, but allow the user to override their name](https://github.com/freeCodeCamp/chapter/issues/110#issuecomment-566292771)
- the user needs a public profile page. The alternative is we simply show the _user's_ name on the _RSVP_ list.
- privacy is set at the_user_ profile level.  If privacy is set at the _chapter_-level, then we'd need a privacy profile settings page.  If privacy is set at at the _event_ level only, then we'd don't necessarily need a _user_ level privacy setting.  

I've shared my opinion that a subscription management / settings page is not a necessary MVP feature for a [number or reasons outlined earlier].(https://github.com/freeCodeCamp/chapter/issues/274#issuecomment-565740617).

I've figured out that "Organizer subscriptions" in the checklist above was likely copied from [Meetup's settings page](https://www.meetup.com/account/subscription/). If that's the case then "Organizer Dashboard" and "Admin Dashboard" are more appropriate to #260 since everything is free and organizers don't pay a subscription.

allella:
[...]

PR #411, opened 2020-10-11
feat: Google Authentication via passportjs and one tap

### LINKED ISSUE #306: As a user I can log in or register via Google OAuth 2.0 authentication.
[The consensus was to split the authentication piece out](https://github.com/freeCodeCamp/chapter/issues/110#issuecomment-566281560) authenication from the RSVP issues(#110 and #278) 

The API / code issue was split to #305.

pdotsani:

Closes #306 

- Added server side authentication to google auth via passport js library and the auth20 api.
- Added client side to server side auth via google one tap javascript api
- Updated user schema to accommodate google profile attributes
- Added some UI to trigger login

This feature requires new dependencies and a few environment variables to run. You can create these credentials/keys at [console.cloud.google.com](https://console.cloud.google.com/) or ping me on discord.


client/src/components/PageLayout/Header.tsx
```diff
@@ -45,6 +47,43 @@ const useStyles = makeStyles({
...
+  useEffect(() => {
+    const script = document.createElement('script');
+
+    script.src = 'https://accounts.google.com/gsi/client';
+    script.async = true;
+    script.defer = true;
+
+    script.onload = () => {
+      const handleCredentialResponse = async (response: any) => {
+        console.log('handleCredentialResponse ', response);
```
timmyichen, line 59: console logs should be removed

client/.env.local.example
```diff
@@ -0,0 +1,5 @@
+NEXT_PUBLIC_GOOGLE_CLIENT_ID=<client id here>
+NEXT_PUBLIC_GOOGLE_API_KEY=<api key here>
+# will need to be assigned new values on deployment
+NEXT_PUBLIC_TOKEN_AUTH_URL=http://localhost:5000/auth/google/token
+NEXT_PUBLIC_IS_AUTH_URL=http://localhost:5000/auth/is-auth/
```
timmyichen, line 5: nit: newline

.env.example
```diff
@@ -11,3 +11,8 @@ CHAPTER_EMAIL=ourEmail@placeholder.place
 EMAIL_USERNAME=emailUsernamePlaceholder
 EMAIL_PASSWORD=emailPasswordPlaceholder
 EMAIL_SERVICE=emailServicePlaceholder
+
+GOOGLE_CLIENT_ID=id from applications dashboard
+GOOGLE_CLIENT_SECRET=secret from applicaitons dashboard
+# Will need to be changed for production
+GOOGLE_REDIRECT_URI=http://localhost:5000/auth/google/callback
```
timmyichen, line 18: nit: newline

client/.env.local.example
timmyichen, line 5: why do these need to be environment variables?

.env.example
timmyichen, line 18: also - why does this need to be an environment variable?

client/src/constants/Header.ts
```diff
@@ -9,4 +9,9 @@ export const headerLinks: IHeaderLink[] = [
   { name: 'chapters', label: 'Chapters', href: '/chapters' },
   { name: 'login', label: 'Login', href: '/login' },
   { name: 'register', label: 'Register', href: '/register' },
+  {
+    name: 'google',
+    label: 'google',
+    href: 'http://localhost:5000/auth/google/',
+  },
```
timmyichen, line 16: why does google have its own link in the header? shouldn't this just be part of logging in?

server/app.ts
```diff
@@ -27,7 +30,11 @@ const PORT = process.env.PORT || 5000;
 
 export const main = async (app: Express) => {
   await initDB();
+  app.use(session({ secret: 'secret' }));
```
timmyichen, line 33: this secret should be an environment variable

server/auth/google/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  )
+  .get(
+    '/callback',
+    passport.authenticate('google', {
+      failureRedirect: '/',
+      session: true,
+    }),
+    (req, res) => {
+      console.log('GOTHERE! ');
+      console.log('req.user ', req.user);
```
timmyichen, line 23: extraneous console logs

server/auth/google/index.ts
```diff
@@ -0,0 +1,32 @@
...
+      res.redirect('http://localhost:3000/');
```
timmyichen, line 24: shouldn't be hardcoding to localhost, can we just redirect to `/`?

server/auth/google/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    },
+  )
+  .post('/token', passport.authenticate('google-verify-token'), (req, res) => {
+    console.log(req.user);
```
timmyichen, line 28: extraneous console log

server/auth/google/passport.ts
```diff
@@ -0,0 +1,71 @@
...
+        callbackURL: process.env.GOOGLE_REDIRECT_URI,
+      },
+      async function(_accessToken, _refreshToken, profile, cb) {
+        const jsonProfile = profile._json;
+        const user = await User.findOne({ google_id: jsonProfile.sub });
+        // todo: add an update to get users most recent picture
+        if (user) {
+          return cb(null, user);
+        } else {
+          // Tmemporary untill we have way to create / add users to chapters
```
timmyichen, line 30: nit: spelling

```suggestion
          // Temporary until we have way to create / add users to chapters
```

server/auth/google/passport.ts
```diff
@@ -0,0 +1,71 @@
...
+  );
+
+  passport.use(
+    new GoogleTokenStrategy(
+      {
+        clientID: process.env.GOOGLE_CLIENT_ID,
+      },
+      async function(parsedToken, googleId, done) {
+        console.log('parsedToken ', parsedToken);
+        console.log('googleId ', googleId);
```
timmyichen, line 51: extraneous console logs

server/auth/index.ts
```diff
@@ -0,0 +1,23 @@
+import express from 'express';
+import Setup from './google/passport';
```
timmyichen, line 2: nit: `Setup` -> `setup`

server/auth/index.ts
```diff
@@ -0,0 +1,23 @@
...
+
+Setup();
+
+const router = express.Router();
+
+router.use(
+  '/is-auth', // add { credentials: 'include' } into fetch request for this endpoint to work
+  router.get('/', (req, res) => {
+    console.log('is there ', req.isAuthenticated());
+    console.log('is there ', req.user);
```
timmyichen, line 13: extraneous console.logs

server/auth/index.ts
```diff
@@ -0,0 +1,23 @@
...
+    if (req.isAuthenticated()) {
+      return res.send({ isAuth: true });
+    } else {
+      return res.send({ isAuth: false });
+    }
```
timmyichen, line 18:
```suggestion
    return res.send({ isAuth: req.isAuthenticated() });
```

server/auth/index.ts
```diff
@@ -0,0 +1,23 @@
+import express from 'express';
+import Setup from './google/passport';
+import googleRouter from './google';
+
+Setup();
+
+const router = express.Router();
+
+router.use(
+  '/is-auth', // add { credentials: 'include' } into fetch request for this endpoint to work
```
timmyichen, line 10: would prefer this to be spelled, or at least add 'ed',  out for clarity
```suggestion
  '/is-authed', // add { credentials: 'include' } into fetch request for this endpoint to work
```

timmyichen (commented):
Thanks for taking on this monster of a task!  I left a few comments/questions.  There are quite a few todos in comments - is this PR complete or did I jump the gun on reviewing?

Ryuno-Ki:
@timmyichen Do we happen to have a logging library already (I like [loglevel](https://www.npmjs.com/package/loglevel) personally), which should be used instead of `console.log()`?

timmyichen:
> @timmyichen Do we happen to have a logging library already (I like [loglevel](https://www.npmjs.com/package/loglevel) personally), which should be used instead of `console.log()`?

not yet i don't think, though i'm not sure if one is needed.  what use case were you thinking?  we can take this into Discord or into an issue if it's not relevant to this PR

Ryuno-Ki:
Let's move it to #412, since it is outside the scope of this PR.

client/.env.local.example
pdotsani: Depending on deployment. I wasn't sure if the url for the api would change in production. If not, then I guess we could hardcode them.

.env.example
pdotsani: See my comment above about the other env vars.

server/auth/index.ts
pdotsani: I'll take care of this locally, then push so I can update the call as well

client/.env.local.example
timmyichen: It should be relative regardless, right? Eg `/auth/google/token` should work regardless of what the host is

pdotsani:
> Thanks for taking on this monster of a task! I left a few comments/questions. There are quite a few todos in comments - is this PR complete or did I jump the gun on reviewing?

@timmyichen It's almost there! The ui still needs some logic to prevent the one tap from loading and the google login from showing when you're logged in. Also, a logout endpoint needs to be created and connected to the ui. Should I add those changes to this pr or leave it for another one?

timmyichen:
> > Thanks for taking on this monster of a task! I left a few comments/questions. There are quite a few todos in comments - is this PR complete or did I jump the gun on reviewing?
> 
> @timmyichen It's almost there! The ui still needs some logic to prevent the one tap from loading and the google login from showing when you're logged in. Also, a logout endpoint needs to be created and connected to the ui. Should I add those changes to this pr or leave it for another one?
> 
> Sorry about the extraneous logs btw. I'll handle those soon.

No worries! Same PR makes sense, given it's a related feature and logging out shouldn't be that complex using passport (i think it's just `req.logout()`). If you want to kick it back into draft mode until it's ready to be reviewed, you can do so under the reviewers - that way you don't get me commenting on a bunch of console logs 😉 

image

client/.env.local.example
pdotsani: Frontend is on 3000 and the server is on 5000, so I thought I needed to distinguish between the two.

timmyichen: ahhhhhhhhhhhhh right.  in that case, the client should just set like `baseUrl` to be whatever its meant to be, and then in the code it should just be ``${baseUrl}/auth/google/token``

although i don't know why they are separate (and don't agree with it), so i'll look to change that soon

client/src/components/PageLayout/Header.tsx
```diff
@@ -16,6 +18,66 @@ const useStyles = makeStyles({
...
+        `${process.env.NEXT_PUBLIC_CHAPTER_SERVER}/auth/is-authed`,
+        {
+          method: 'get',
+          credentials: 'include',
+        },
+      );
+      const responseBody = await response.json();
+      const { isAuth } = responseBody;
+
+      if (!isAuth) {
```
timmyichen, line 35: `isAuth` will always be `false` in this hook because it's initialized to be false, and the hook only runs once on mount.

client/src/components/PageLayout/Header.tsx
```diff
@@ -40,13 +102,43 @@ const Header: React.FC<{ classes: Record<string, any> }> = ({ classes }) => {
...
+                  </a>
+                </Grid>
+              );
+            } else if (headerLink.name === 'logout' && !isAuth) {
+              return null;
+            }
+
+            if (headerLink.name === 'login' && isAuth) {
+              return null;
+            }
```
timmyichen, line 132: This can probably be rewritten to be easier to understand - maybe something like the following, outside of the .map call:

```js
const links = headerLinks;

if (isAuth) {
  links.push(logout button config object);
} else {
  links.push(login button config object);
}

links.map(link => etc etc
```

client/src/constants/Header.ts
timmyichen: This comment still stands^

client/src/constants/Header.ts
```diff
@@ -9,4 +9,14 @@ export const headerLinks: IHeaderLink[] = [
   { name: 'chapters', label: 'Chapters', href: '/chapters' },
   { name: 'login', label: 'Login', href: '/login' },
   { name: 'register', label: 'Register', href: '/register' },
+  {
+    name: 'google',
+    label: 'google',
+    href: 'http://localhost:5000/auth/google/',
```
timmyichen, line 15: we shouldn't hardcode localhost:5000

db/migrations/1575817780103-User.ts
```diff
@@ -4,13 +4,16 @@ export class User1575817780103 implements MigrationInterface {
...
     await queryRunner.query(
       `CREATE TABLE "users" (
           "id" SERIAL NOT NULL,
           "first_name" character varying NOT NULL,
           "last_name" character varying NOT NULL,
           "email" character varying NOT NULL,
-          "password_digest" character varying NOT NULL,
+          "google_id" character varying DEFAULT NULL,
+          "google_picture" character varying DEFAULT NULL,
+          "password_digest" character varying,
```
timmyichen, line 16: we shouldn't edit existing migrations - if you're changing a model, create a new migration.

server/app.ts
timmyichen: Bumping this comment

server/auth/google/index.ts
```diff
@@ -0,0 +1,29 @@
+import express from 'express';
+import passport from 'passport';
+
+const router = express.Router();
+
+router
```
timmyichen, line 6: I think in other files we are not chaining multiple routes together - each get request here should be its own statement

server/auth/google/index.ts
timmyichen: bump

server/auth/google/passport.ts
```diff
@@ -0,0 +1,69 @@
...
+import { Strategy as GoogleStrategy } from 'passport-google-oauth20';
+import { Strategy as GoogleTokenStrategy } from 'passport-google-verify-token';
+import { User } from '../../models';
+
+export default function() {
+  passport.serializeUser((user: any, done) => {
+    done(null, user.id);
+  });
+
+  passport.deserializeUser(async (userId: any, done) => {
```
timmyichen, line 11: we know `userId` will be a number, so we can type it as such here

timmyichen (changes requested):
Thanks again for taking this on!  I left a few comments - there are a few older unaddressed comments as well.  Please ping me if you have any questions.

allella:
@all-contributors please add @pdotsani for code

client/src/components/PageLayout/Header.tsx
pdotsani: This `isAuth` is refering to the api call in the func.

allella:
[Today's meeting included a good deal of conversation about how Google Auth](https://github.com/freeCodeCamp/chapter/wiki/Meeting-Notes-November-7,-2020#authentication) would limit use cases in countries like China or even certain government agencies.

So, we left the conversation leaning very heavily to going back to an email solution.  There are some contributors like @timmyichen @vkWeb @AryanJ-NYC  from #306 who were part of previous authentication conversations who weren't part of today's conversation, so please see the [meeting notes](https://github.com/freeCodeCamp/chapter/wiki/Meeting-Notes-November-7,-2020#authentication). 

If Google Auth is off the table, we'd need to at least :
- close #411, #306 
- update #262 and https://github.com/freeCodeCamp/chapter/wiki/Owner-FAQs
- create a new user story with any prior conversations about using email authentication

allella:
@QuincyLarson @pdotsani @Zeko369 is my last comment and link to the meeting notes an accurate summary of today's conversation?

allella:
Closing since the direction changed to using email-based authentication on #437.

Issue #378, opened 2020-07-20
Schema: Add field(s) for video conference URL / Online Events


allella:
@Zeko369 @timmyichen @QuincyLarson @tomiiide 

One of actionable [next steps](https://github.com/freeCodeCamp/chapter/wiki/Meeting-Notes-July-18,-2020#next-steps) was to update the schema, models, and API to handle online / virtual events.

The [schema ER diagram](https://github.com/freeCodeCamp/chapter/blob/master/data/schema.png) may be a bit out of date, but I think it's current enough for this conversation.

### Rough Ideas to Seed the Conversation
- add an optional _events.video_url_ field specifically for a video conference / live stream URL.
- add a _events.url_ field specifically for a general event website / URL. In the application's logic, this could default to the _event_'s public _Chapter_ URL.  An _organizer_ could set / override this URL to point somewhere else. For instance, if the _chapter_ is hosting a hack-a-thon / large _event_, then they may prefer to direct API traffic to a more general outside_events.url_ landing page. 
- add the following fields to _venues_ (street_address, city / locality, postal_code, region, country, latitude, longitude). Otherwise, _events_ would be advertising to meet at a physical _venue_ but the _user_ may have no idea where to go.
- we'll want a way to filter between physical and online _events_. The easiest way to do this would be to have an _events.venue_type_ field with enumerated options {"physical", "online", "physical-and-online"}. It's possible we could infer whether an _event_ is physical, online, or both without this field, but it would be more complex and subject to errors.  Quincy mentioned a toggle in the UI for "Physical" vs "Online", but "Both" is just as likely to happen. We might as well ask the _organizer_ to specify when an _event_ is created and the UI could show/hide elements for a physical venue and/or online/video and we avoid the guesswork.
- remove postal_code from _locations_.  A _chapter_ like freeCodeCamp Greenville doesn't really have a "postal code". _chapters_ and non-profits typically serve a broader locality or region so a physical zip code for a chapter is probably not of much value to a _user_.  If we later develop a zip code radius search feature, then we could decide to add a postal code back, but I suspect we'd use a _venue_'s postal code and not that of a _chapter_. (edit: I proposed we [remove _locations_ in a comment below](https://github.com/freeCodeCamp/chapter/issues/378#issuecomment-665950015))

### Caveats

[...]

PR #423, opened 2020-10-25
feat(client, db): event url

### LINKED ISSUE #381: Add event URL for each event
Some bigger events might have landing pages so it would be cool if we could link to them.

Add url: string, optional column to Events table

For more info check #378

katien:

Closes #381

Added URL field to event model, created migration for new column to be added to event table, updated seeders to generate fake URLs, updated UI to allow read and update on URL.

While working on this I noticed that the form to edit events is creating new events.

### DIFF
--- a/client/src/generated/index.tsx
+++ b/client/src/generated/index.tsx
@@ -63,6 +63,7 @@ export type Event = {
   updated_at: Scalars['DateTime'];
   name: Scalars['String'];
   description: Scalars['String'];
+  url?: Maybe<Scalars['String']>;
   start_at: Scalars['DateTime'];
   ends_at: Scalars['DateTime'];
   canceled: Scalars['Boolean'];
@@ -280,6 +281,7 @@ export type UpdateVenueInputs = {
 export type CreateEventInputs = {
   name: Scalars['String'];
   description: Scalars['String'];
+  url: Scalars['String'];
   start_at: Scalars['DateTime'];
   ends_at: Scalars['DateTime'];
   capacity: Scalars['Float'];
@@ -290,6 +292,7 @@ export type CreateEventInputs = {
 export type UpdateEventInputs = {
   name?: Maybe<Scalars['String']>;
   description?: Maybe<Scalars['String']>;
+  url?: Maybe<Scalars['String']>;
   start_at?: Maybe<Scalars['DateTime']>;
   ends_at?: Maybe<Scalars['DateTime']>;
   capacity?: Maybe<Scalars['Float']>;
@@ -302,7 +305,7 @@ export type EventsQuery = { __typename?: 'Query' } & {
   events: Array<
     { __typename?: 'Event' } & Pick<
       Event,
-      'id' | 'name' | 'canceled' | 'description' | 'capacity'
+      'id' | 'name' | 'canceled' | 'description' | 'url' | 'capacity'
     > & {
         tags?: Maybe<Array<{ __typename?: 'Tag' } & Pick<Tag, 'id' | 'name'>>>;
       }
@@ -320,6 +323,7 @@ export type EventQuery = { __typename?: 'Query' } & {
       | 'id'
       | 'name'
       | 'description'
+      | 'url'
       | 'canceled'
       | 'capacity'
       | 'start_at'
@@ -356,7 +360,13 @@ export type EventVenuesQuery = { __typename?: 'Query' } & {
   event?: Maybe<
     { __typename?: 'Event' } & Pick<
       Event,
-      'id' | 'name' | 'description' | 'capacity' | 'start_at' | 'ends_at'
+      | 'id'
+      | 'name'
+      | 'description'
+      | 'url'
+      | 'capacity'
+      | 'start_at'
+      | 'ends_at'
     > & {
         tags?: Maybe<Array<{ __typename?: 'Tag' } & Pick<Tag, 'id' | 'name'>>>;
         venue: { __typename?: 'Venue' } & Pick<Venue, 'id'>;
@@ -372,7 +382,7 @@ export type CreateEventMutationVariables = Exact<{
 export type CreateEventMutation = { __typename?: 'Mutation' } & {
   createEvent: { __typename?: 'Event' } & Pick<
     Event,
-    'id' | 'name' | 'canceled' | 'description' | 'capacity'
+    'id' | 'name' | 'canceled' | 'description' | 'url' | 'capacity'
   > & {
       tags?: Maybe<Array<{ __typename?: 'Tag' } & Pick<Tag, 'id' | 'name'>>>;
     };
@@ -489,6 +499,7 @@ export const EventsDocument = gql`
       name
       canceled
       description
+      url
       capacity
       tags {
         id
@@ -547,6 +558,7 @@ export const EventDocument = gql`
       id
       name
       description
+      url
       canceled
       capacity
       start_at
@@ -627,6 +639,7 @@ export const EventVenuesDocument = gql`
       id
       name
       description
+      url
       capacity
       start_at
       ends_at
@@ -698,6 +711,7 @@ export const CreateEventDocument = gql`
       name
       canceled
       description
+      url
       capacity
       tags {
         id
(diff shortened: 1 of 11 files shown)


client/src/modules/dashboard/Events/pages/EventPage.tsx
```diff
@@ -24,7 +24,7 @@ export const EventPage: NextPage = () => {
             {loading
               ? 'Loading...'
               : !error
-              ? 'Cant find event :('
+              ? "Can't find event :("
```
timmyichen, line 27: nit: single quotes instead of double

server/controllers/Events/resolver.ts
```diff
@@ -47,6 +47,7 @@ export class EventResolver {
 
     event.name = data.name ?? event.name;
     event.description = data.description ?? event.description;
+    event.url = data.url ?? event.url;
```
timmyichen, line 50: I would recommend adding some validation before here to ensure that the url being added is indeed a URL, but this is more of a nice-to-have, and not a blocker.

timmyichen (commented):
Do any tests need updating? Do we even have tests for this? 🤔

client/src/modules/dashboard/Events/pages/EventPage.tsx
Ryuno-Ki: I'd go with a double quote here to avoid escaping the single quote in `Can't`.

server/controllers/Events/resolver.ts
Ryuno-Ki: Should be easy by passing it to `new URL()` though (in a try-catch block).

client/src/modules/dashboard/Events/pages/EventPage.tsx
katien: I was getting a lint error with single + the escaped char so I switched to double

timmyichen: Ah right thanks!

katien:
The only tests I saw were for the mailer

allella:
I'm no good on the technical end, but this does appear to add an _event.url_ as we discussed in https://github.com/freeCodeCamp/chapter/issues/378 and which became #381 so I'm good with it.

@katien it sounded like @timmyichen and @Ryuno-Ki were suggesting adding logic to validate the value passed in is actually a valid URL.  Could that be added to this PR?

katien:
Validation is in there :)

server/controllers/Events/resolver.ts
```diff
@@ -56,7 +58,7 @@ export class EventResolver {
       if (!venue) throw new Error('Cant find venue');
       event.venue = venue;
     }
-
+    if (event.url) this.validateEventUrl(event.url);
```
timmyichen, line 61: Nit: single line if statements are discouraged due to their brittleness.

```suggestion
    if (event.url) {
      this.validateEventUrl(event.url);
    }
```

server/controllers/Events/resolver.ts
```diff
@@ -80,4 +82,12 @@ export class EventResolver {
 
     return true;
   }
+
+  validateEventUrl(url) {
+    try {
+      new URL(url);
```
timmyichen, line 88: using `new URL()` is actually quite restrictive, and I think we want to be a little more permissive.  For example, the following will throw errors, despite being "fine" URLs:

```
new URL('google.com')
new URL('www.google.com')
```

Furthermore, valid URLs via `new URL()` include protocols we don't want to support, like `ftp` and `mailto`.

Suggestion: Before doing the try-catch, prepend it with `http://` if it isn't already, and if it passes, ensure the result's protocol is `http:` or `https:`  Let me know if this makes sense.

server/controllers/Events/resolver.ts
timmyichen: (same change to where this is done above)

server/controllers/Events/resolver.ts
katien: I don't know how much value this is going to add, if we prepend http to the input, `new URL` will accept just about anything. We could try regex-based validation, but do we really need to be restricting what users enter in here? I think we're more likely to end up preventing someone from entering something they want to have in there than helping them with this validation

timmyichen: Fair - I do think this is a nice to have and not at all a requirement, especially since we don't really validate any other fields.  I'd be fine with not adding any validation for now, and adding it later when we have better testing so we can also test it.

allella: If we're skipping the validation, is it safe to assume the framework handles sanitizing output so we're not subject to XSS on this field or any other user input for that matter?

server/controllers/Events/resolver.ts
katien: They're gone now that the validation is gone but I was following the convention already established in the file. If we don't want single life ifs maybe we should update the lint settings

allella (approved):
I'm good with this from the business logic.  If @timmyichen  and @Ryuno-Ki are good then holler and one of us can merge it.

timmyichen (approved):

allella:
@Ryuno-Ki if you give the formal approval, then we'll merge this in.

Ryuno-Ki:
Sorry, I was busy the last days. Lemme take a look!

client/src/generated/index.tsx
```diff
@@ -302,7 +305,7 @@ export type EventsQuery = { __typename?: 'Query' } & {
   events: Array<
     { __typename?: 'Event' } & Pick<
       Event,
-      'id' | 'name' | 'canceled' | 'description' | 'capacity'
+      'id' | 'name' | 'canceled' | 'description' | 'url' | 'capacity'
```
Ryuno-Ki, line 308: Would it make sense to have these ordered alphabetically?
(The OCD in me. Would not block a merge IMHO).

client/src/generated/index.tsx
```diff
@@ -320,6 +323,7 @@ export type EventQuery = { __typename?: 'Query' } & {
       | 'id'
       | 'name'
       | 'description'
+      | 'url'
```
Ryuno-Ki, line 326: Would it make sense to sort these alphabetically?
(Not blocking a merge IMHO).

client/src/modules/dashboard/Events/components/EventItem.tsx
```diff
@@ -37,6 +37,7 @@ const EventItem: React.FC<IEventItemProps> = ({ loading, event }) => {
         <Typography variant="body2" color="textSecondary" component="p">
           {event.description}
         </Typography>
+        {event.url && <a href={event.url}>{event.url}</a>}
```
Ryuno-Ki, line 40: Perhaps worth to wrap in a `p`, too?

client/src/modules/dashboard/Events/pages/EventPage.tsx
```diff
@@ -48,6 +48,7 @@ export const EventPage: NextPage = () => {
             <Typography variant="body2" color="textSecondary" component="p">
               {data.event.description}
             </Typography>
+            <a href={data.event.url}>{data.event.url}</a>
```
Ryuno-Ki, line 51: Perhaps worth to wrap in a `<p>`, too?

Ryuno-Ki (approved):
I'm worried about the lack of validation on URL input.
For example, this could lead to a persistent XSS via `javascript:` protocol.
But this could be also adressed in a followup-PR.

So I'm approving with comments.

katien:
@Ryuno-Ki agreed on sanitization being important, @timmyichen and I were going back and forth on how to handle this here since the suggested `new URL()` wasn't doing what we wanted (`new URL("javascript:alert('foo')")` would pass, for example). FYI URL validation has also been removed from my other PR with the assumption it would be handled in a later issue

Ryuno-Ki:
Could we at least have an issue for that?
It would be handy if it would link to those discussions here and your other PR.

allella:
We can continue the conversation on URL validation on #430.  If there are no final points, then we'll merge #423 and #424.

allella:
@Ryuno-Ki  agreed to proceed on Discord

> Looked at #423. Fine with me, but make sure to fill some issues / followup-PRs, @allella
> Yes, sanitizing is important. Input validation as well. Those are the two cornerstones of web security.

PR #428, opened 2020-10-28
feat(server): send email

megajon:

I'm trying to create some tests for the sendEmail resolver but when I run the "npm run test" command I get the error "TypeError: Cannot convert undefined or null to object".

jest_error


Ryuno-Ki:
I checked out this PR by running the following:

```
git remote add megajon https://github.com/megajon/chapter.git
git fetch
git fetch megajon feat/sendEmail:feat/sendEmail
```
(The last commit has this branch on the left hand of the colon and the local branch name on the right).

Ryuno-Ki:
Looking closely at the error report shows `jest.config.js:13:21` as the faulty one (line 13, column 21 of `jest.config.js`).
That line means

https://github.com/freeCodeCamp/chapter/blob/2ec3c0bbcbf0ad33b4d54f8a625600813b4c8163/jest.config.js#L13

So I looked into where `compilerOptions.paths` comes from. Logging it yields `undefined` (and thus the above error message).
The reason? No property `paths` in this object:

https://github.com/freeCodeCamp/chapter/blob/2ec3c0bbcbf0ad33b4d54f8a625600813b4c8163/tsconfig.json#L2-L29

That line in Jest config was added with https://github.com/freeCodeCamp/chapter/commit/292b1e54177d2d7e1069b0fc265760bcd0df933c as part of #326. So I'm pinging @phoenisx (creator) and @Zeko369 for more insights into the context of the change there.

ojeytonwilliams:
That's what I found, too.  Simply removing `moduleNameMapper` is enough to get the tests running.

Ryuno-Ki:
It's broken on `master`, too.

phoenisx:
I haven't worked on this project for a long time now, so have lost context.

Anyways, I do remember why I made those changes:

* Typeform used to throw module not found errors, for the versions installed at that time, as we were using custom aliases for `rootPath`s.
* To resolve custom aliases in Jest, `moduleNameMapper` was required, which was coded to take aliases from `tsconfig.json`.

Some recent changes:

* https://github.com/freeCodeCamp/chapter/pull/390/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519
* https://github.com/freeCodeCamp/chapter/pull/389/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0

have removed custom aliases from `chapter` project setup.

If they are desired changes, then `moduleNameMapper` can be removed as well 👍 

cc: @Zeko369

ojeytonwilliams:
Thanks for taking a look @phoenisx. Am I right in thinking that Jest would fail if the custom aliases were there, but `moduleNameMapper` was missing or not configured correctly?  If so, given that master passes with it removed, it's probably safe to take it out.

phoenisx:
@ojeytonwilliams I have to set up the codebase again (will do it sometime later).

Just by looking in the code, written test cases are very limited and present only for `server/*` root. Also, these test cases are using relative URLs, instead of absolute, except https://github.com/freeCodeCamp/chapter/blob/master/server/tests/Mailer.test.ts, where:

`import Utilities from 'server/util/Utilities';` is absolute, and might not work!!

If the above test is not failing after removing `moduleNameMapper`, then I am sure you can remove it for now, unless we require aliases mappers later on. 🎉

megajon:
I'm having trouble getting input validation to work. I've imported 'class-validator' and added a length restriction to the 'subject' field in server/controllers/Messages/inputs.ts, but when I run the 'sendEmail' resolver from the graphql playground with an empty string in the 'subject' field, the resolver runs successfully when there should be an input validation error.

input_validation

jest.config.js
```diff
@@ -9,10 +9,10 @@ module.exports = {
...
     '^.+\\.tsx?$': 'babel-jest',
   },
-  moduleDirectories: ['node_modules', 'server'],
-  moduleNameMapper: pathsToModuleNameMapper(compilerOptions.paths, {
-    prefix: '<rootDir>/',
-  }),
+  // moduleDirectories: ['node_modules', 'server'],
+  // moduleNameMapper: pathsToModuleNameMapper(compilerOptions.paths, {
+  //   prefix: '<rootDir>/',
+  // }),
```
Ryuno-Ki, line 15: From personal experience, I can tell, that it is better to use a block comment + add a line stating the reason, why code is commented (instead of deleted).

allella:
@megajon I have zero experience with any of this code stack, but there is an [example in the docs that includes MaxLength](https://typegraphql.com/docs/validation.html#how-to-use), so perhaps that's a missing piece of it?

server/controllers/Messages/inputs.ts
```diff
@@ -0,0 +1,15 @@
+import { Length } from 'class-validator';
+import { InputType, Field } from 'type-graphql';
+
+@InputType()
+export class SendEmailInputs {
+  @Field(() => [String])
+  to: String;
```
ojeytonwilliams, line 7:
```suggestion
  to: [string];
```
TypeScript was complaining with just `String`

server/controllers/Messages/resolver.ts
```diff
@@ -0,0 +1,19 @@
+import { Resolver, Mutation, Arg } from 'type-graphql';
+import MailerService from '../../services/MailerService';
+import { SendEmailInputs } from './inputs';
+import { Email } from './Email';
+
```
ojeytonwilliams, line 5:
```suggestion
import { validate } from 'class-validator';

```

server/controllers/Messages/resolver.ts
```diff
@@ -0,0 +1,19 @@
...
+    await email.sendEmail();
+    return {
+      ourEmail: email.ourEmail,
+      emailList: email.emailList,
+      subject: email.subject,
+      htmlEmail: email.htmlEmail,
+      backupText: email.backupText,
+    };
+  }
+}
```
ojeytonwilliams, line 19:
```suggestion
@Resolver()
export class EmailResolver {
  @Mutation(() => Email) async sendEmail(@Arg('data') data: SendEmailInputs) {
    const email = new MailerService(data.to, data.subject, data.html);
    const errors = await validate(data);

    if (errors.length > 0) {
      console.log('validation failed. errors: ', errors);
      // This object is returned just to pass the type checks.
      return {
        ourEmail: 'this failed', // Debugging entry, so you can see from the
        // graphql playground that the validation failed
        emailList: email.emailList,
        subject: email.subject,
        htmlEmail: email.htmlEmail,
        backupText: email.backupText,
      };
    } else {
      console.log('validation succeeded');
      await email.sendEmail();
      return {
        ourEmail: email.ourEmail,
        emailList: email.emailList,
        subject: email.subject,
        htmlEmail: email.htmlEmail,
        backupText: email.backupText,
      };
    }
  }
}
```
This (obviously) isn't a real solution, but I hope it helps.

ojeytonwilliams (commented):
Hey @megajon, I've added a sketch of a solution to the validation issue below.

server/controllers/Messages/resolver.ts
megajon: @ojeytonwilliams It appears that the input validation wasn't working because the 'validate: false' option was set in the buildSchema() function in server/app.ts. Once I remove the 'validate: false' option the built in validation works. I prefer the built in method over importing 'validate' from 'class-validator' and writing custom error handling logic in the resolver, but I suppose the option was set for a reason. @Zeko369 Is there a specific reason the 'validate: false' option is set in the buidSchema() function?

ojeytonwilliams: Ah, I assumed we needed to invoke `validate` directly.  I'm glad you spotted that.

From what I can gather from the [docs](https://typegraphql.com/docs/validation.html#how-to-use) you only use that option if you want to manually enable validation for specific arguments.  Given that `SendEmailInputs` is the only class being validated at the moment, it should be fine.

megajon: According to the docs input validation is enabled by default, all that is needed is to import class-validator and decorate the fields with the appropriate validation requirements. The 'validate: false' option is used to disable the default validation. At the moment the SendEmailInputs class is the only class validating input, but whichever method we use to validate input should be used in any resolver that accepts inputs, so we should decide early on which method we're going to use so we can be consistent. I prefer using the default validation, unless someone else disagrees.

ojeytonwilliams: It seems to me that custom handling approach would be much slower to write, so we might as well defer that effort until we're confident it's necessary.  I'd say go ahead with the default validation for now and we can always switch to the custom approach later if we have to.

megajon:
I've been able to get the input validation and testing to work for the sendEmail resolver. There is still some work to be done regarding sanitization and authentication, but I think its close to being ready to incorporate into other features. I created some custom validation to do some checks that don't appear to be built into the class-validator library, but one of the tests for the resolver is failing because there is no email service to return an email object after the resolver is successfully called. So far I've been using the NodemailerApp on my dev machine as a local email service, but that requires adding configuration in MailerService.ts that I don't want to commit. To make development smoother for everyone I think it might be a good idea to use ethereal as a fake email service for the time being, but I didn't want make those kinds of changes without getting other opinions. Anyone against using ethereal until we figure out what were going to use for a real email service?

allella:
Thanks for the update @megajon 

I believe we're still going to depend on the SMTP protocol and not installing custom Nodemailer plugins, so that give a good deal of flexibility for dev and production email configurations.

Ethereal and other "fake" services like Mailtrap should all support SMTP giving each developer a choice of what they want to use for capturing test emails.

I assume there's a sample developer configuration, so perhaps that can mention the options for development email, such as Ethereal.

megajon:
I was finally able to get the test for a successfully sent email to pass and the test emails are successfully sent to the nodeMailerApp. As I mentioned before there is still some work to be done regarding authentication middleware and input validation, but at this point I think the sendEmail resolver is ready to be merged for development purposes.

allella:
@Zeko369  I know you're not free for a few weeks, but if you have time in the interim, I think your review would help Jonathan move this along.

package.json
```diff
@@ -127,7 +127,7 @@
     }
   },
   "engines": {
-    "node": "13.x",
-    "npm": ">= 6.0.0"
+    "node": ">=14.15.0",
```
allella, line 130: @Zeko369  were you suggesting @megajon and @tomnoland remove this "engines" section entirely?

They both suggested not supporting Node 13, so would it be best to fix these to a Node 14 compatible version, or remove the section, [like you suggested on Discord](https://discord.com/channels/633401573002969128/633826607617933353/817832134387040318)?

megajon: I noticed that the master branch had the engine set to node 14 so I copied those settings to my sendEmail branch and I no longer get the error.

megajon: Another problem I'm having is that when I run the client side server from my sendEmail branch I get an error stating that a certain module can't be resolved. I noticed that my sendEmail branch is missing a few directories under client/src/ that are present in the master branch, which seems to work fine when I run both the client and backend servers. This will be my first time merging code into an open source project so I'm guessing the directories are missing from my sendEmail branch because they were created after I created the branch. I'm assuming everything will work properly once the code is merged, but I just wanted to check to make sure I'm not missing something.

megajon:
Another problem I'm having is that when I run the client side server from my sendEmail branch I get an error stating that a certain module can't be resolved. I noticed that my sendEmail branch is missing a few directories under client/src/ that are present in the master branch, which seems to work fine when I run both the client and backend servers. This will be my first time merging code into an open source project so I'm guessing the directories are missing from my sendEmail branch because they were created after I created the branch. I'm assuming everything will work properly once the code is merged, but I just wanted to check to make sure I'm not missing something.

allella:
@megajon Others may have a better suggestion, but I think your feat/sendEmail branch can be rebased to resolve conflicts with the master branch.   I don't use Github, so there may be an easier way.

allella:
@Ryuno-Ki and @timmyichen have rebased on this project before and can hopefully share tips in case you have an issue resolving the conflicts. A rebase may avoid merge conflicts and be easier than manually resolving.

megajon:
I tried to rebase the master branch to my sendEmail branch, but it ended up deleting all the code that I wrote, so I used merge instead and everything seems to be working the way it should.

server/controllers/Messages/resolver.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      subject
+      htmlEmail
+    }
+  }
+`;
+
+describe('Test sendEmail resolver', () => {
+  it('Should reject the mutation call when subject field has exceeded MaxLength', async () => {
+    const emailData = {
+      to: ['jonathan.seubert@megajon.com', 'finisher1017@gmail.com'],
```
allella, line 21: Was having these emails addresses in the code intentional?  Perhaps replace with a .env variable?

server/controllers/Messages/resolver.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    });
+
+    expect(email.errors[0].message).to.equal('Argument Validation Error');
+  });
+
+  it('Should reject the mutation call when there is an improperly formatted email in the list', async () => {
+    const emailData = {
+      to: [
+        'jonathan.seubert@megajon.com',
+        'finisher1017@gmail.com',
```
allella, line 57: Same email addresses here too.

server/controllers/Messages/resolver.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    });
+
+    expect(email.errors[0].message).to.equal('Argument Validation Error');
+  });
+
+  it('Should reject the mutation call when a duplicate email address is found', async () => {
+    const emailData = {
+      to: [
+        'jonathan.seubert@megajon.com',
+        'finisher1017@gmail.com',
```
allella, line 78: and here

ojeytonwilliams (changes requested):
Hey @megajon.  I've taken a good look and made some suggestions.  There are still a few more things I need to test, but I wanted to see what you thought.

server/controllers/Messages/inputs.ts
```diff
@@ -0,0 +1,26 @@
...
+  @Field(() => [String])
+  @IsListEmpty({ message: 'email list cannot be empty' })
+  @ValidateEmailList({ message: 'list contains invalid email' })
+  @FindDuplicateEmails({
+    message: 'list contains one or more duplicate emails',
+  })
+  to: [];
+
+  @Field()
+  @MaxLength(10)
```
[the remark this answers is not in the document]
megajon: I set it to 10 characters so I didn't have to write out a long string in the tests and I wasn't sure what the appropriate limit should be, so I just left it at 10 for now. Any ideas on what the appropriate limit should be?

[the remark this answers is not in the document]
allella: If we're talking about an email address, here's a quote from the standard.

> In addition to restrictions on syntax, there is a length limit on email addresses. That limit is a maximum of 64 characters (octets) in the "local part" (before the "@") and a maximum of 255 characters (octets) in the domain part (after the "@") for a total length of 320 characters.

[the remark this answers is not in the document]
ojeytonwilliams: I think it's the email subject and, according to https://stackoverflow.com/questions/1592291/what-is-the-email-subject-length-limit (and the RFC it references) the hard limit (per line) is 998 characters and the recommendation (again, per line) is 78.

I tested gmail and it has the limit of 998, so that might be a good idea.  Assuming that nodemailer can cope with that.

server/controllers/Messages/resolver.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    };
+
+    const email = await callSchema({
+      source: emailMutation,
+      variableValues: {
+        data: emailData,
+      },
+    });
+
+    expect(email.errors[0].message).to.equal('Argument Validation Error');
```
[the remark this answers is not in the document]
megajon: This is exactly what I was trying to do! For some reason neither my vscode nor the console.logs were showing all of the properties available on the email object that gets returned. I spent quite a bit of time trying to figure out how to access the specific error messages being returned, but came up short, which is why I wrote it the other way as a workaround. This is my first time working with graphql, so figuring out how to do proper error handling has been a challenge. I went ahead and made the changes you suggested. Thanks for the assist. Nice to have an extra set of eyes on this.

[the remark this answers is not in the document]
ojeytonwilliams: Yeah, this took some trial and error before I even found the properties.  Turns out `email.errors[0]` has a bunch of non-enumerable properties, including the ones we're interested in.  `console.log` does not always show them, so isn't reliable.  The easiest way I know to get all the info about an object is to use
```js
const { inspect } = require('util');
console.log(inspect(obj, true, null, true));
```
if you're in node and have access to `util`.  The options are, respectively, showHidden, depth and color.  That should show you everything might be interested in.

server/controllers/Messages/inputs.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@InputType()
+export class SendEmailInputs {
+  @Field(() => [String])
+  @IsListEmpty({ message: 'email list cannot be empty' })
+  @ValidateEmailList({ message: 'list contains invalid email' })
+  @FindDuplicateEmails({
+    message: 'list contains one or more duplicate emails',
+  })
+  to: [];
```
ojeytonwilliams, line 18:
```suggestion
  to: string[];
```
Shouldn't we specify the type since we know it?  Sorry about the original suggestion, I'm still learning TS and messed up the syntax.

server/controllers/Messages/inputs.ts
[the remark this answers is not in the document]
ojeytonwilliams: By the way, there's a neat function to generate a long string.

```
var lotsOfAs = 'a'.repeat(999);
```

ojeytonwilliams (approved):
LGTM :+1: 

I went ahead and reverted the `IS_DOCKER` change - I couldn't see a reason to change the default. Also, I removed the changes to package.json and package-lock.json.  While we do want to keep them updated, those changes felt out of scope for this PR.

@megajon if you're okay with those changes, I think we can get this in.

megajon:
@ojeytonwilliams The changes are fine with me. It would be a good idea to create some documentation on how to configure a service to receive the emails sent by the resolver, but I think the code is in good enough shape to be merged.

ojeytonwilliams:
Cool, I'll create a PR for the docs change and ping you for review.

ojeytonwilliams:
This is such a key feature, so thanks for all your hard work and patience.  I'm glad we're able to get this in!

allella:
@megajon will the configuration mostly be in the .env section?

If you want to share rough details I can work it into the documentation.

Thanks

allella:
@megajon Could you share an example of the variable names needed in the .env for the mailer service?

Nick (@nhcarrigan) was asking on the chat due to what seemed to be missing values. 

I don't recall if that was addressed after the last comment above. 

I think all that needs to be done is add sample key=value in the _.env.example_ file and any helpful comments in that file if needed.   I can also update the CONTRIBUTING.md if / as needed once we a _.env.example_ to work from.

Thanks

PR #463, opened 2021-03-26
merge fix(docs): add npm run postinstall 

### LINKED ISSUE #462: Add `npm run postinstall` to contributing.md
### Describe the bug:

The _.env_ file is mandatory for the project to build, however it is intentionally not provided as part of the project as it will contain the project secrets. The Chapter developer/installer is required to issue the proper commands to create and edit the _.env_ file.

A script has been provided to copy defaults values from the _.env.example_ into the _.env_ .

Put the following sentence into the CONTRIBUTING.md file around line 310 in
the section run the app step 1.

> Run `npm run postinstall` to copy the file _.env.example_ into the file _.env_.
> If the _.env_ file exists it will not be overwritten.

- Browser(s) name and version:  __terminal__
- Operating System: 
uname -ior
5.3.7-301.fc31.x86_64 x86_64 GNU/Linux

NorthDecoder:

Closes #462

The _.env_ file is mandatory but not supplied by the project.
Running the postinstall script command will copy the
_.env.example_ file into the _.env_ file. If the _.env_ exists,
the copy will not occur.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,19 @@
+**Table of Contents**
+
+* [Contribution Guidelines](#contribution-guidelines)
+* [Contributing Code](#contributing-code)
+* [Running the Application](#running-the-application)
+* [Frequently Asked Questions](#frequently-asked-questions)
+* [Server-side Technical Documentation](#server-side-technical-documentation)
+  * [API Specification](#api-specification)
+  * [Database](#database)
+    * [Schema](#schema)
+    * [Username and Password](#username-and-password)
+    * [Host and Port](#host-and-port)
+    * [Admin Tools](#admin-tools)
+    * [Using TypeORM and Yarn](#using-typeorm-and-yarn)
+* [Troubleshooting](#troubleshooting)
+    
 # Contribution Guidelines
 
 Hello :wave: and welcome to **_Chapter_**, a project of [freeCodeCamp](https://www.freecodecamp.org).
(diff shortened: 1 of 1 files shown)


allella:
@NorthDecoder I believe the "Docker" mode install will already run the post install script, so this may be best to add only in the "Manual Mode".  Would you agree we could move this under the Manual Mode heading of Step 2 of Running the Application?

Also, I suspect this could merge in fine, but the this PR is going master to master and the CONTRIBUTING.md suggests all PRs be off a branch like docs/name-of-your-branch.

Thanks,
Jim

allella:
Actually, I don't think it's Docker that creates the .env.

I just did a fresh install and ran the 
_npx recursive-install_
from Step 1 of Running the Application and it created a .env where one didn't exist before I ran it.  It must be the case that this command runs the postinstall hook of package.json

```
  "scripts": {
....
    "postinstall": "node scripts/postInstall.js",
```



That step is for Docker Mode or Manual Mode.  So, it would see we don't need to manually call 
_npm run postinstall_
as long as that step is run.

allella:
It looks like the ["post" prefix in package.json scripts creates a special lifecycle hook](https://docs.npmjs.com/cli/v7/using-npm/scripts) that fires when _npm install_, or seemingly the related _npx recursive-install_ are run.

So, that likely explains why running
`npx recursive-install`
executes _scripts/postInstall.sh_, which then creates a .env if it doesn't exist.

@NorthDecoder can you try a fresh test, like

```
mkdir chapter-test && cd chapter-test
git clone https://github.com/freeCodeCamp/chapter.git .
ls -la
npx recursive-install
ls -la

```
and see if the .env file gets created (via the output of the ls command before and after the npx)?

NorthDecoder:
I have removed project and started fresh again and found that the postinstall script was indeed run and the default .env file was created as expected. Therefore this pull request issue and the associated issue #462 can be closed, with understanding to continue review of wording in the contributing.md file.

PR #472, opened 2021-04-02
fix: repair broken imports

ojeytonwilliams:

In https://github.com/freeCodeCamp/chapter/pull/470 I removed `baseUrl` and this broke several absolute imports we had been using.  

This PR reverts that, adds some configuration to fix the broken imports and adds a lint rule that should check the imports for us.


ojeytonwilliams:
I'm still trying to understand exactly how the module resolution works. Everything seems fine, as far as VSCode's ts linter is concerned, but some absolute paths still fail and I'm not sure why.  Hence the blocked status.

ojeytonwilliams:
I went into the weeds a bit here.  End result being:

We can use 

```js
import Utilities from 'server/util/Utilities';
```
interchangeably with 
```js
import Utilities from '../util/Utilities';
```

whichever is easiest to read.

Some broken imports ('server/util/Utilities' amongst others) should now be fixed.

Eslint should warn us if any imports cannot be resolved.

If no-one wants absolute imports, we can simplify the config further.  Personally I quite like having the option, since `../../../../stuff` can be frustrating to deal with.

nhcarrigan:
Okay, giving this a pretty quick poke...

If I attempt to actually compile the TS with `tsc` on the `master` branch, I get all sorts of unresolved import errors. 

image

When I do the same on this PR, all of the unresolved imports are fixed (woohoo!). There's an error about `req.user` in the testUtils, but that's a separate issue I believe.

image

As far as I can tell, this PR takes care of the import issues.

ojeytonwilliams:
Thanks for taking a look, Nick, I've been staring at this so long, I don't trust myself.  

I agree that second issue is (probably) unrelated.

allella:
Sorry if I jumped the gun on the last merge.  If / when this is good just let me know when it's ready. Or, I assume one of you two has merge rights as well.

ojeytonwilliams:
> Sorry if I jumped the gun on the last merge.

I don't think so. It'll just take a few tries before all the kinks are ironed out.  Not helped by my inexperience with TS, but I'm getting there :)

ojeytonwilliams:
The reason for the block was that I was investigating `tsc` and the code it produces.  Turns out there's a neat package https://github.com/justkey007/tsc-alias that fixes all the absolute imports, once `outDir` is set in the tsconfig. 

For now, this is fine, but I think that's the way to go when we're closer to production.

nhcarrigan:
Hey Oliver,

The goal with this PR is to specifically resolve the import errors correct? When I run `tsc` locally I see a few other build errors, but all of the imports are resolved.

ojeytonwilliams:
Hey Nick,

That's exactly right.  I wanted to keep this focused on the import errors and https://github.com/freeCodeCamp/chapter/pull/540 should fix the rest.

nhcarrigan (approved):
This looks good to me, then - I don't see any import errors when I run the compiler.

ojeytonwilliams:
Thanks for going through this, Nick, much appreciated.

nhcarrigan:
Always glad to help 🙂

Issue #468, opened 2021-03-31
Lint eslint/interface-name-prefix


NorthDecoder:
### Describe the bug:

On the command line __`npm run lint`__ causes some errors.

The first error is
>  Interface name must be prefixed with "I"  @typescript-eslint/interface-name-prefix

Researching further results in finding that the current recommendation by the Typescipt team is to change the rule so that the interface names do __NOT__ begin with 'I'.

* [typescript-eslint/issues/374](https://github.com/typescript-eslint/typescript-eslint/issues/374)

* The code for [do-not-begin-with-an-i](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md#enforce-that-interface-names-do-not-begin-with-an-i)

* SO
  * [62915344](https://stackoverflow.com/questions/62915344/eslint-erro-when-adding-rule-typescript-eslint-interface-name-prefix)

Looking into `.eslintrc.json` to see the rule for discussion.

```json
"@typescript-eslint/interface-name-prefix": [
      2,
      { "prefixWithI": "always" }
    ]
```

We either need to fix the rule or fix the code :)

### Tell us about your browser and operating system:
- Browser(s) name and version: __terminal__
- Operating System: __5.3.7-301.fc31.x86_64 GNU/Linux__

allella:
@NorthDecoder I have not technical experience or opinion, but I scanned the comments and the final comment says they

> We will _not_ be removing interface-name-prefix from the recommended set

If that's the case, does that mean _"prefixWithI": "always"_ is still the default? If so, it sounds like "I" prefixes are still the official direction, yes / no?

allella:
Well, I see the other links suggest this may have been changed to _no I_, as you suggested.  If that's the official suggestion, then feel free to share a PR to fix it.

NorthDecoder:
My understanding is that they left the option available for those who want it one way or the other.  As you saw there was a long discussion by both sides of the opinion and that was the compromise.

I will change the setting to __'no I'__ to match the current status of the Chapter code.

ojeytonwilliams:
@NorthDecoder could we just drop the rule since it's deprecated?  I think it will default to **'no I'** if we do that.

NorthDecoder:
@timmyichen , in the blame for the file it looks like you added this rule. Do you have a strong preference for the 'I' before the interface names?

PR #473, opened 2021-04-02
feat: update and re-enable CI workflow

ojeytonwilliams:

The tests should fail initially - the lint step definitely has a few problems.  So, I will keep this as a draft at least until everything's green.

I think it would help a lot if we get these running again and hopefully catch a few bugs earlier.

### DIFF
--- a/.github/workflows/ci-workflow.yaml
+++ b/.github/workflows/ci-workflow.yaml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 jobs:
   build:
     name: build and test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


ojeytonwilliams:
Would it be useful to split this into a few workflows?  It might be frustrating to fix a lint error, try again, fix a broken test etc. etc.  Rather than just seeing all the problems up front. That said, fixing the lint error might fix the rest automatically, so I'm not sure which is best.

nhcarrigan:
> Error:   13:11  error  Interface name must be prefixed with "I"  @typescript-eslint/interface-name-prefix

This is a weird rule - I thought this fell out of practise?

allella:
@nhcarrigan #468 was talking about that too. @NorthDecoder said he'd remove the rule, but if you're in there feel free to do it.

ojeytonwilliams:
It seems renovate bot will not automerge PRs [if there are no checks](https://docs.renovatebot.com/automerge-configuration/#absence-of-tests).  So, to try and get that going, I'm merging this.

PR #474, opened 2021-04-02
fix(tools): remove interface rule

### LINKED ISSUE #468: Lint eslint/interface-name-prefix

nhcarrigan:
- [X] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md).
- X] My pull request has a descriptive title (not a vague title like `Update README.md`).
- X] My pull request targets the `master` branch of Chapter.

Closes #468 

This PR removes the linter rule that expects interface declaration names to be prefaced with `I`. Instead, it now asserts that they are *not* prefaced with `I`. 

If we would prefer the linter not care either way, I can push a commit to disable the rule instead of removing it.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -29,10 +29,6 @@
   "rules": {
     "@typescript-eslint/camelcase": 0,
     "@typescript-eslint/explicit-function-return-type": 0,
-    "@typescript-eslint/interface-name-prefix": [
-      2,
-      { "prefixWithI": "always" }
-    ],
     "@typescript-eslint/no-explicit-any": 0,
     "@typescript-eslint/no-unused-vars": [
       "error",


allella (approved):
Per the conversation in #468 we decided this was good, no "I" prefix.

nhcarrigan:
This will introduce more lint errors that we'll need to fix :D

NorthDecoder:
I have not tested yet, but it seems like removing the rule would remove the 'missing I' errors and allow various name choices for this particular rule that would not fail any either way.

ojeytonwilliams:
I just noticed we're too major versions behind the latest @typescript-eslint/eslint-plugin and parser.  How about we just move to the latest versions, discard any deprecated/removed rules and see how that looks?  I'm happy to go through and fix any errors it finds.

nhcarrigan:
> So we are really back to deciding 'I' or 'no I' and either way fixing the code to pass the lint tests.

We can also explicitly disable the rule, I believe, and allow either convention to pass. It looks like this is done in the `types` directory.

ojeytonwilliams:
I should have mentioned - we don't need to worry about this particular rule if we upgrade the plugin to v4.  Simply because they no-longer support it.

Unless there's a strong reason to stick with V2, I think upgrading should solve a few problems (this one, at least!).

nhcarrigan:
Upgrading it locally resolves that issue, but introduces the good ol' `missing return type on function` rule everywhere.

ojeytonwilliams:
> `missing return type on function` rule everywhere.

Is that related to this, https://github.com/facebook/create-react-app/pull/8177 ?

If so, it sounds like using React.FC isn't considered to be a good practice any more, so we should probably switch to something else.  What do you reckon?

nhcarrigan:
I don't think so - it's that the linter expects us to have explicit type definitions for functions.

```ts
const function = (arg: argtype): returntype => { code }
```

If we don't have the `returntype` definition, the linter gets mad.

ojeytonwilliams:
Oh, simpler than I thought. It seems like that's a reasonable request.  We can add those to my todo list.

NorthDecoder:
This commit will have the desired effect of conforming linting for interface "I" prefix to the more recent industry standard of "no I".  I have tested that the `docker- compose up` command will still continue to deploy the containers in there current condition. 

With the knowledge we have gained from the comments above, this commit can be merged knowing that linting will highlight the necessary changes.

ojeytonwilliams (approved):
LGTM 👍 

While this rule isn't in the latest typescript-eslint, I think we all agree that 'no I' is good rule. As @NorthDecoder found, the naming convention rule can enforce it. I'll use that when I update the linter.

Issue #445, opened 2021-03-08
Map out a "Golden Path" for a Demo Video


allella:
Per the [latest meeting](https://github.com/freeCodeCamp/chapter/wiki/Meeting-Notes-March-6,-2021#project-updates), Quincy suggested mapping out a "golden path" for a demo video.

@tomiiide previously [shared a work flow](https://www.figma.com/file/ehgBfxoLKrlSZH0uftD6dA/Chapter-Trial?node-id=0%3A1) which can largely describe critical paths for a _user_, _organizer_,  or _administrator_.

@QuincyLarson were you thinking we create a short video for each main role (user, organizer, administrator)?  Or, if we had to pick one would we focus on admin or organizer path, since they are more likely to be the consumer of a MVP demo video?

allella:
@QuincyLarson 

> were you thinking we create a short video for each main role (user, organizer, administrator)?

allella:
Once we narrow down which _user role(s)_ we can compare @tomiiide 's Figma roles workflow with comments from @Madalena-15 and others on https://github.com/freeCodeCamp/chapter/issues/320

That includes, among other things, [a user persona PDF for an _organizer_ named Julia(https://github.com/freeCodeCamp/chapter/issues/320#issuecomment-581653505).

Zeko369:
## Idea, for the whole flow, this should cover pretty much all the features we're planning to have in the MVP.

### User
- User comes to the page -> list events on homepage -> select -> click on RSVP -> modal opens to create an account -> crates an account -> recives an email -> joins -> has 2 buttons GCAL link / iCal download to import (also in email?)
- User -> looks at other events -> wants to RSVP but is invite only -> requests
Organizer -> accepts
- User -> looks at other events -> wants to RSVP but is full -> requests -> another user cancels -> get's email about joining

### Organizer
- gets email about new member of chapter and RSVP -> sends email to all RSVPs about a food change or something -> changes venue on event (generates email) -> {User gets both emails}

### Admin
- Admin -> logs in -> list of all chapters in this instance (one is private with a flag) -> invites a user to it (maybe better in organizer flow) -> invites a admin user to another instance

*Now we need to trim this down to get all the features we want to present in a 5min video*

megajon:
@Zeko369 If a user is a member of multiple chapters within the same organization, can they view the different chapters from a single interface, or do they have to manually change the url in order to view different chapters?

Zeko369:
[...]

PR #414, opened 2020-10-13
Revert "Mock search bar with Material UI. refs #349 (#409)"

timmyichen:
This reverts commit 942f802322d4e80b1ba90e41c4754145ee33e6d7.

The reverted commit is captured in the `feat/search` branch.  Once feature complete, the `feat/search` branch will then be merged to master.

PR #477, opened 2021-04-06
fix: Make the site/app header look more like the mockup

paizle:

Updated the site header to look more like the mockup [here](https://www.figma.com/proto/q7DikyL3N0c4CUWxHNa97i/Chapter-Prototype?node-id=332%3A355&scaling=scale-down).

Swapped the placeholder for the search box with the logo.  Replaced the "search" placeholder with a TextField pre-pended with a search icon. Adjusted the spacing a bit.

Relates to issue #445 and #414. Making some minor style changes to get the app looking more like the mockup for the "Golden Path".

### DIFF
--- a/client/src/components/PageLayout/Header.tsx
+++ b/client/src/components/PageLayout/Header.tsx
@@ -4,6 +4,10 @@ import { makeStyles, Grid } from '@material-ui/core';
 
 import { headerLinks } from '../../constants/Header';
 
+import TextField from '@material-ui/core/TextField';
+import InputAdornment from '@material-ui/core/InputAdornment';
+import Search from '@material-ui/icons/Search';
+
 const useStyles = makeStyles({
   link: {
     textDecoration: 'none',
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (changes requested):
Nice work @paizle. One minor issue, the 'Upcoming Events' heading is now a bit closer to the header than it was before:

image

In the figma mockup there's the same about of space above the headings as below.

paizle:
Ok, I found the problem with the spacing after the header.

Just to note, I had initially changed the spacing in the grid header with `spacing={2}` but this was causing a horizontal scroll bar on the page. My fix was to add `xs={12}` to the grid but this causes the spacing issue after the header as shown in the image:
image

I followed a thread [here](https://stackoverflow.com/questions/45519275/grid-in-material-ui-causes-horizontal-scroll-react) on StackOverflow that illustrates the issue and links to the MUI github with discussion around the same issue/fix. I ended up going with the accepted SO answer which seems to solve it well:

```
style={{
        margin: 0,
        width: '100%',
      }}
```

ojeytonwilliams (approved):
LGTM :+1: 

The style trick worked nicely.

PR #525, opened 2021-04-09
docs(contributing): add run lint and test

### LINKED ISSUE #524: docs(contributing) Add run lint wording and guideline
### Describe the bug:

Consider:
1. adding wording to CONTRIBUTING that suggests linting and testing prior to running the application.
2. adding a guideline regarding the merging of pull requests that must pass linting and tests.

Reference pull request #525
- Browser(s) name and version: n/a
- Operating System: n/a

NorthDecoder:

Closes #524

Suggests running lint and tests prior to starting the app, and then advises that maintainers will not accept pull requests with lint or test errors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -188,9 +188,11 @@ You are almost ready to make changes to files, but before that you should **alwa
 
 5. Test your code **Always!** 
 
+    * Before starting the application confirm that `npm run lint` and then `npm run test` pass.  Fix all errors that do not pass.
     * If you started the application using the _Docker Mode_, then tests are run using `NODE_ENV=test docker-compose exec app npm run test` OR if you want to use the "watch" mode run `NODE_ENV=test docker-compose exec app npm run test:watch`
     * If you started the application using the _Manual Mode_ (without Docker), then tests are run using `npm run test` OR if you want to use the "watch" mode run `npm run test:watch`
-
+    * Project maintainers will not merge pull requests that fail to pass linting or tests.
+ 
 6. Stage the changes and make a commit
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -188,9 +188,11 @@ You are almost ready to make changes to files, but before that you should **alwa
 
 5. Test your code **Always!** 
 
+    * Before starting the application confirm that `npm run lint` and then `npm run test` pass.  Fix all errors that do not pass.
```
allella, line 191: Would it be advisable to have a script command in package.json that runs all the necessary tests at once?  

@ojeytonwilliams is there a common pattern with running all tests at once, like
_npm run all-tests_
vs having people individually run each test?  

For instance, the _"both"_ script has an example of concurrently running multiple commands with a single script command. 

`    "both": "concurrently \"npm run dev\" \"cd client && npm run dev\"",
`

Also, since the two bullet point directly below this line are specific to Docker Mode and Manual Mode, how about we remove this new line and update those two bullets as needed?

ojeytonwilliams: Sure, we could do that.  I don't know that it's a common pattern, but it seems perfectly reasonable to me.  I'd call it `lint-and-test`, though, just because `test` is implicitly "run all tests".  At least that's how I interpret it.

> Also, since the two bullet point directly below this line are specific to Docker Mode and Manual Mode, how about we remove this new line and update those two bullets as needed?

Sounds good.

allella:
We already mention email services in the [_Owner_ FAQ](https://github.com/freeCodeCamp/chapter/wiki/Owner-FAQs), but I suspect we'll have a more concrete [step-by-step](https://documentation.divio.com/how-to-guides/) for the actual _Owner_ setup

PR #540, opened 2021-04-12
feat: compile TypeScript in CI

ojeytonwilliams:
- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `master` branch of Chapter.

This is a companion to https://github.com/freeCodeCamp/chapter/pull/472.  The idea is that issues like https://github.com/freeCodeCamp/chapter/issues/530 should be caught by automated testing.  

Some of the compilation errors are fixed in https://github.com/freeCodeCamp/chapter/pull/472, so if that gets merged first the updated CI should pass.

PR #572, opened 2021-04-16
fix: add iterator key for home page

ojeytonwilliams:

Fixes the error that @NorthDecoder encountered while reviewing https://github.com/freeCodeCamp/chapter/pull/540. `chapter.id` should be unique to each chapter which makes it a good candidate for a React key.


allella:
Seems like a simple change to me, even though I don't understand the full mechanics of it.

If there's no downside, and it fixes and existing bug, then I'd move to merge it in.

ojeytonwilliams:
> If there's no downside, and it fixes and existing bug, then I'd move to merge it in.

React needs the key so that it can be 100% sure which element is which when it re-renders the page.  It shouldn't have any side-effects.

PR #584, opened 2021-04-25
fix: allow docker build to work remotely

ojeytonwilliams:

Closes https://github.com/freeCodeCamp/chapter/issues/585

This lets you pass the Apollo server location to Docker.


NorthDecoder:
This commit was merged at 2 parents 99648a3 + b6f2a3b commit c0f73f00bef3545a6b56c5f5f6ebfa4313d06e20
and tested with `docker-compose up` to view the client to see that the CORS error persisted, so issue #585 was opened.

NorthDecoder:
Per the added hint in the *docker-compose.yml* file I changed the *.env* by adding 
```bash
NEXT_PUBLIC_APOLLO_SERVER="http://my.server.ip.address:5000"
```
in the client the error message reads `Error: JSON.parse: unexpected character at line 1 column 1 of the JSON data`
in the client web dev tools networking the error message reads
```bash
POST http://45.33.104.204:3000/%22http://45.33.104.204:5000%22/graphql
```


OR

```bash
# without the quotes
NEXT_PUBLIC_APOLLO_SERVER=http://my.server.ip.address:5000
```
in the client error message  reads `Error: relation "chapters" does not exist`
and the web dev tool networking shows
```
200 my.server.ip.address:5000
```
So it looks like the option without the quotes does send a query to the correct address, without the CORS error.

Searching the *README.md* and *CONTRIBUTING.md* for 'NEXT_PUBLIC_APOLLO_SERVER' I see there is no mention, so there is no clear way for the casual installer to know how to handle deployment to a server.

ojeytonwilliams:
@NorthDecoder I've updated the PR to include some instructions to help when running Chapter remotely, so the next person doesn't have to guess :)

@allella could you let me know what you think about the docs changes?  This seemed like the least disruptive way to add these instructions.

CONTRIBUTING.md
```diff
@@ -533,6 +541,10 @@ it should ouput something like
  [X] MigrationName1575633316367
 ```
 
+# Running Remotely
+
+When not running locally, the client needs to be passed the server's location.  This can be done by changing your _.env_ file to include `NEXT_PUBLIC_APOLLO_SERVER=<http://address.of.graphql.server>`.  For example, if you started **_Chapter_** with `npm run both` and hosted it on `http://demo.chapter.org` then the address will be `http://demo.chapter.org:5000`.
```
allella, line 546: @ojeytonwilliams this seems like a good start.

The only change I'd suggest is linking ".env file" to the .env section of the same file using the link to #env-configuration-file, like

[_.env_ file](#env-configuration-file)

allella (approved):
Approving the docs changes, with the one suggested tweak in the comments.

allella (changes requested):
@ojeytonwilliams one other change, we'll want the new NEXT_PUBLIC_APOLLO_SERVER variable defined in the .env.example

ojeytonwilliams:
Thanks for catching those @allella - I've made the changes.

NorthDecoder:
On a VM at 2 parents c0f73f0 + 0fd44b8 commit 6eea125741e9d255d1d70ba5979f9af1bf802f6a
I ran
```bash
rm .env
sudo rm -rf node_modules/; sudo rm -rf client/node_modules/
npx recursive-install
# edit the .env file per proposed CONTRIBUTING instruction  to add VM ip address
npm run lint # passed
npm run test # passed
docker-compose up
```
In the client the CORS error is gone, and it appears that the client is queriyng the graphql server.

allella:
@ojeytonwilliams if you confirm if the port needs to be part of the variable, then I can make the suggested changes as a test of committing to an open PR, either with the CLI tool or the manual way.

ojeytonwilliams:
@allella yep, the port is necessary.  Feel free to experiment as much as you like with this PR - everything is recoverable from!

FYI, this only affects Docker, because the next client doesn't care about .env files below it's root (/client).

allella (approved):
I submitted a manual commit to add the port to the .env.example comment.

allella:
@ojeytonwilliams I pushed a couple commits to make the comments of the NEXT_PUBLIC_APOLLO_SERVER consistent.

So, things are good as far as I'm concerned and I think @NorthDecoder's issue was resolved too.

CONTRIBUTING.md
```diff
@@ -533,6 +541,10 @@ it should ouput something like
  [X] MigrationName1575633316367
 ```
 
+# Running Remotely
+
+When not running locally, the client needs to be passed the server's location by changing your [_.env_](#env-configuration-file) file to include `NEXT_PUBLIC_APOLLO_SERVER=<https://address.of.graphql.server:port>`.  For example, if you started **_Chapter_** with `npm run both` and hosted it on `https://example.com` then the address will be `http://example.com:5000`.
```
ojeytonwilliams, line 546:
```suggestion
When not running locally, the client needs to be passed the server's location by changing your [_.env_](#env-configuration-file) file to include `NEXT_PUBLIC_APOLLO_SERVER=<https://address.of.graphql.server:port>`.  For example, if you started **_Chapter_** with `npm run both` and hosted it on `https://example.com` then the address will be `https://example.com:5000`.
```
The 's' is missing from the protocol.

allella:
Looks good to me. I'd say to merge this if there are no other concerns.

PR #597, opened 2021-04-27
fix: make automerges push directly, not create PRs

ojeytonwilliams:

Prior to this PR renovate would create PRs for all changes, including ones it would automatically merge.  Since dev input is not necessary for automerged PRs, this changes the config so that they will not be notified either.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -6,7 +6,8 @@
     {
       "matchUpdateTypes": ["minor", "patch", "pin", "digest"],
       "matchCurrentVersion": "!/^0/",
-      "automerge": true
+      "automerge": true,
+      "automergeType": "branch"
     }
   ],
   "rangeStrategy": "pin"


allella (approved):
I think tweaking is worth a shot.

I know Fran wants it off, and perhaps we go that route, but maybe some of these "noise" reduction steps will find a good middle-ground.

ojeytonwilliams:
Sure.  This should help a lot, I think.  Thanks for the suggestion, by the way.

allella:
Added the chat comments in the event others wonder what this was about, or if it comes back around at a later date.

> allella
> 9:54 PM
> ojeytonwilliams on the RenovateBot "noise" emails, would it help if we changed the rule and/or scheduling. For example, is it worth having "patch" in the rule (https://github.com/freeCodeCamp/chapter/blob/master/renovate.json)? Or, if patch were removed, then would it reduce updates to only happen on new "minor" versions?
> "matchUpdateTypes": ["minor", "patch", "pin", "digest"]
> Also, since we seem to be using automerging, there's a branch strategy that is said to reduce noise as well
> https://docs.renovatebot.com/noise-reduction/
> Scheduling was also part of the conversation at https://github.com/renovatebot/config-help/issues/88
> If you think any of this is practical, then I can create an issue based on what we discuss here.
> 
> 
> zeko369
> 6:26 AM
> Honestly don't see that much sense in using Renovate, I'd rather remove it for now and just update depends manually when/if we need to and enable this later when we hit a stable release to help with maintenance
> Also I don't like pinning depends for a non library type of project
> 
> 
> oliver
> 6:27 AM
> I don't see the problem with it, honestly. It lets us know exactly which change breaks things and removes a chore.
> Also, major version bumps (i.e. the only ones that will generate noise) give us a poke to see if we actually need the dependency or if it's just hanging around.
> 
> 
> zeko369
> 6:30 AM
> Let's keep on going with it for now, but I do think that upgrading depends on every change is not really necessary
> 
> 
> oliver
> 6:32 AM
> I know it's a bit of a hassle, but I'm happy to take care of it. I just may need a bit of input occasionally if I'm poking at code I only partly understand!
> 👍 1 
> Though I'm gradually wrapping my brain around things 🙂
> 
> 
> zeko369
> 6:32 AM
> Nice

ojeytonwilliams:
Thanks Jim, that context would definitely have got lost otherwise.

allella:
@all-contributors add @ojeytonwilliams for code, maintenance, tests, and bugs

Issue #444, opened 2021-03-08
As a user (authenticated or not), I can view a list of upcoming events on the homepage


allella:
- [x] API call to get the next 5 upcoming events 
- [x] Material UI card layout for rendering a single event
- [x] Render initial view of events
- [x] Logic to render additional events via lazy-load / infinite scroll

Example from the mock-up
https://www.figma.com/proto/q7DikyL3N0c4CUWxHNa97i/Chapter-Prototype?node-id=332%3A355&scaling=scale-down

allella:
@dtwilliams10 we decide if we're going to implement the "filter" shown in the mockup.

For MVP, I'm thinking the filter may not be necessary since / if the location of an event is likely the primary decision for a _user_.

allella:
@paizle perhaps you and @dtwilliams10 can move along this homepage and related search bar #349.

allella:
Actually, I see the MVP [homepage mock-up](https://www.figma.com/proto/q7DikyL3N0c4CUWxHNa97i/Chapter-Prototype?node-id=1%3A2&scaling=scale-down) doesn't include the filters and that seems reasonable since there will be a search bar and "Events" page that may offer more features.

dtwilliams10:
I was thinking to load eager load 5 events (need to figure out how to determine those 5) at page load into individual cards and then lazy load the rest in a paginated format.  Thoughts, ideas, suggestions?

allella:
I'd assume the homepage would be ordered upcoming event date.

Loading +/-5 and lazy loading the rest sounds like a fine start to me.

allella:
@dtwilliams10  if you have ideas on more specific "tasks" for the top of this issue.

We've learned that many contributors appreciate viewing the issue as pieces, while others like to maintain the user story approach.  

As a middle-ground, we're going to add a smaller task-based list at the top of the user story issues and we can link off to smaller task-based issues as needed.

This will keep the user story issues for a broader view of the [MVP project](https://github.com/freeCodeCamp/chapter/projects/1#card-56434681).

Do you have thoughts on additional tasks for this issue?

Thanks

paizle:
@dtwilliams10 It sounds like you've got a good idea to handle this so maybe I can just help support you by taking care of time consuming tasks while I get more familiar with everything.

allella:
@dtwilliams10  @paizle the merged in button in #413 created a few issues mentioned on #450

Those are creating an issue with the front-end.

[...]

PR #607, opened 2021-05-01
Implement lazy loading events on homepage

Zeko369:
It's limited to 2 events for now to be more obvious


Zeko369:
Closes #444 since all other things were already implemented

PR #610, opened 2021-05-04
Code Quality Challange (30 days)

Zeko369:
https://tuple.app/code-quality-challenge


allella:
@Zeko369 what's the plan / desire with the client/README.md?   Is this something other developers need to get the app running or is it a special case for internal purposes?

We've previously consolidated all of the technical notes into the CONTRIBUTING.md, per Quincy's suggestion to have one long document, so if this is something others will need to reference than adding to a new or existing section in CONTRIBUTING.md may be the best place for it.

Zeko369:
Well, didn't do this 🤦

PR #617, opened 2021-05-11
FIX: moved Resolver test into common server/tests directory

paulywill:

Looking at the current tests would suggest keeping all the tests together, i.e. `server/tests` and `client/tests`.
Although it did not take me long I had to dig around to locate the random test in  `server/controllers/Messages/resolver.test.ts`.

I realize this not moving the needle for the MVP other than agreement of placing all the tests in one location.

### DIFF
--- a/server/tests/Resolver.test.ts
+++ b/server/tests/Resolver.test.ts
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
 import faker from 'faker';
-import { callSchema } from '../../testUtils/callSchema';
-import longString from './longString';
+import { callSchema } from '../testUtils/callSchema';
+import longString from '../controllers/Messages/longString';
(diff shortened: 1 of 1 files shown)


ojeytonwilliams:
@paulywill what about going the other way and putting all the tests next to the files they're testing?  It has the fringe benefit that you can go from
```js
import longString from '../controllers/Messages/longString';
```
to
```js
import longString from './longString';
```

I'm fine with either direction, and I do agree that we should be consistent.

paulywill:
Let me do some research on this.

If there's something definitive like a best practice or something to measure.

ojeytonwilliams (approved):
I went ahead and checked a bunch of big, popular libraries and they all used either __tests__ or tests, so I guess that will take the fewest people by surprise.

LGTM 👍

Zeko369:
Merging to close for now, we can see what we'll do with all other tests later (i.e. cypress for client...)

PR #619, opened 2021-05-12
fix: event dashboard edit bugs

ojeytonwilliams:
The defaultValues get merged with the new form values, so cannot include
anything that is not expected by updateEvent

It's a shame that `data` can be `undefined`.  It would be nice if there were two forms (one for new events and one for updating them), so we could have type safe defaultValues, but this seems reasonable for now.


ojeytonwilliams:
@Zeko369 the last commit isn't the DRYest, but it seems to work.  Any ideas on how to refactor it would be very welcome.

Zeko369 (approved):

PR #631, opened 2021-06-03
feat(client): enable tests

ojeytonwilliams:

It took a bit of rearrangement to get the client tests to work (React and testing-library have to be both together and in the right dir, for example), but we can write them now.

Still to be fixed: module resolution for client tests.  In this PR you need to use relative paths for imports, but I plan to address that separately.


Zeko369:
Hey @ojeytonwilliams, let's fix this after 1.7. and make it a draft till then

ojeytonwilliams:
Sure.  My pre-demo plan was to get as much working as possible and my post-demo plan was circle back to writing a ton of tests.

PR #634, opened 2021-06-22
feat: email magic link in development

ojeytonwilliams:
- feat: send the magic link via email
- feat: wrap the link in an anchor tag

With https://github.com/freeCodeCamp/chapter/issues/445 in mind, this makes the login process a little more natural.

### DIFF
--- a/server/controllers/Auth/resolver.ts
+++ b/server/controllers/Auth/resolver.ts
@@ -11,7 +11,7 @@ import {
 import { getConfig, isDev } from 'server/config';
 import { authTokenService } from 'server/services/AuthToken';
 import { GQLCtx } from 'server/ts/gql';
-import { EmailResolver } from '../Messages/resolver';
+import MailerService from 'server/services/MailerService';
 
 type TokenResponseType = {
   email: string;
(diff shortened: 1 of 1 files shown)


ojeytonwilliams:
Just to clarify, this should make the dev experience a bit smoother and provide a fallback for the demo.  The plan is still to provide Google auth for the demo and a few more options for the MVP.

server/controllers/Auth/resolver.ts
```diff
@@ -52,6 +53,11 @@ export class AuthResolver {
       console.log(
         `Code: ${code}\nhttp://localhost:3000/auth/token?token=${token}`,
       );
+      new EmailResolver().sendEmail({
+        to: [data.email],
+        subject: 'Login to Chapter',
+        html: `<a href=http://localhost:3000/auth/token?token=${token}>Click here to log in to chapter</a>`,
+      });
```
Zeko369, line 60:
```suggestion
      await new MailerService(
        [data.email],
        'Login to Chapter',
        `<a href=http://localhost:3000/auth/token?token=${token}>Click here to log in to chapter</a>`,
      ).sendEmail();
```

Zeko369 (changes requested):
I wouldn't mix resolvers into resolvers, create a new service here

ojeytonwilliams:
> I wouldn't mix resolvers into resolvers, create a new service here

Thanks, I wasn't entirely confident this was right. Is the point that resolvers should just be concerned handling GraphQL queries/mutations?

Zeko369:
Yup, when you annotate a class with `@Resolver` use it only in the typeGraphql context, because other stuff might break (args, ctx, other decorators)

Zeko369 (approved):

PR #637, opened 2021-06-23
feat: add chapters to dashboard

ojeytonwilliams:
This is basically a copy-paste of the Venue tab, with a few tweaks to make it work. The /dashboard page is still rather basic.  I didn't try to implement any of @megajon's designs and focused on getting it up and running, rather than looking good.

Nonetheless, it's now possible to edit and update chapters from the dashboard view.  @allella I'd love to get your feedback, even though it's a bit rough.

I based it off of https://github.com/freeCodeCamp/chapter/pull/627, so that I could easily make use of those changes.  Might be easier if we merged that first, or we could just merge this and close that.

### DIFF
--- a/client/src/generated/graphql.tsx
+++ b/client/src/generated/graphql.tsx
@@ -440,7 +440,14 @@ export type ChapterQuery = { __typename?: 'Query' } & {
   chapter?: Maybe<
     { __typename?: 'Chapter' } & Pick<
       Chapter,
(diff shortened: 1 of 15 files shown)


allella:
@ojeytonwilliams I was able to add and edit an event via the dashboard.

I'd say we merge in what you have so others can potentially start looking into the styling.

PR #638, opened 2021-06-24
feat: make tags and events many to many

ojeytonwilliams:
This is still very WIP, and I probably won't have time to do more before the demo, but here's what I have so far.

The UI is just a placeholder (we don't want select, we want checkboxes or something analogous to how Trello handles labels).


allella:
Note that this is part of MVP issue #270

ojeytonwilliams:
Closing as stale.  Also, Tom's working on this.

PR #639, opened 2021-06-24
feat: sync db with models in development

ojeytonwilliams:
If NODE_ENV is 'development', reloading the app will automatically update the database to match the models

Manual syncing can be triggered by the db:sync command

I *think* we can delete the migrations, since db:sync should handle that.  At least, I couldn't see any changes locally when I did. @Zeko369 what do you reckon?

TODO:

- [x] update docs with the workflow Fran suggested in chat: https://chat.freecodecamp.org/channel/chapter?msg=aiHiM5KjuQpvLTCDQ


allella:
Are we saying the TypeORM [**Initializing the Database**](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md#initializing-the-database) section can be updated to add a note about the auto sync feature in development?

> If so would we want to remove the notes these notes below, or just leave them as a manual example?
> 
> The npm run db:reset command will do all three tasks: drop, migrate, and seed.
> 
> If you prefer to run some or all of the steps manually, then they are:
> 
> npm run db:drop
> npm run db:migrate
> npm run db:seed

ojeytonwilliams:
What I've been doing is relying on syncing up until I need to create the PR and then just running a single migration at the end.  I'll update the docs

> Are we saying the TypeORM Initializing the Database section can be updated to add a note about the auto sync feature in development?

Yep, I'll add that to this PR

> If so, would we want to remove the notes below, or just leave them as a manual example?

I think we can make that a little more concise.  The individual scripts are fairly self-explanatory after all.

ojeytonwilliams:
@allella in the end I only modified the wording a little bit.  I don't have a strong preference, though, so if you like the old wording that's cool.

allella:
@ojeytonwilliams I pushed a change to CONTRIBUTING.md to not dev vs production and link to TypeORM's docs on synchronization since I had to lookup that ormconfig.js is where the setting was being defined.

There was a merge conflict, so I also resolved it. I think :)

CONTRIBUTING.md
```diff
@@ -516,6 +512,11 @@ To keep everything DRY, add `extends BaseModel` to the class and import it from
 
 You could also run `npx typeorm` since here you're not actually loading any ts files, but because regular `npx typeorm` runs inside of node it import from _.ts_ files, so we run it with `ts-node` and our custom server config (check package.json)
 
+#### Syncing the Schema
+
+* For development environments, [TypeORM will kept the database migrations in sync](https://github.com/typeorm/typeorm/blob/master/README.md) with the definitions in _server/models_ when the server is started. Developers must still create and commit a [migration](#creating-a-migration) for any modifications to the ORM / database schema.
+* For production environments, or if a manually sync becomes necessary, run the `npm run db:sync` command.
```
ojeytonwilliams, line 518:
```suggestion
* If a manual sync becomes necessary, run the `npm run db:sync` command.
* For production environments, run the `npm run db:migrate` command.
```
I don't think `db:sync` should be run in production.  According to TypeORM's [migration docs](https://typeorm.io/#/migrations/how-migrations-work) you should use `typeorm migration:run` in production (i.e. our `db:migrate` command).

ojeytonwilliams (commented):
Cheers for the updates to the wording, they LGTM.

Weirdly GitHub doesn't let request changes on my own PR, but I have one:

ojeytonwilliams:
Hey @allella if you're happy with the latest changes, could you merge this in?  I've changed the wording a bit to better reflect how I see the dev workflow once the migrations are out of the picture.

When we're closer to the MVP I'll give some thought about (and testing of!) how to handle migrations properly in production, but for now we don't need them.

PR #374, opened 2020-03-16
feat: Add columns to control notification flow

davidals:
Needed in order to implements featured mentioned on issues #107, #276, #359

- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `master` branch of Chapter.

Closes #XXXXX

Issue #276, opened 2019-11-30
Text for automatic emails notifying participants of subsequent events


QuincyLarson:
Depends on #107

allella:
@Zeko369  @timmyichen @vaibhavsingh97 @jacobbogers @eolculnamo2   we'll get some text for this one and #275 but @emred2700 asked about an email template language and I wondered what direction we'd go on the local email templates.

We talked about using nodemailer over SMTP on #146 and #283, but I didn't see a specific template package and perhaps there's one already baked into Nodemailer.

A quick search suggestions you can select a Nodemailer template package like [email-templates](https://www.npmjs.com/package/email-templates).

The choice of email template will define the markup of the text we generate and the location of the local template files.

Is Node mailer definitely the package for sending?
Which templating package, if any, is needed or suggested?

Zeko369:
The issue of templates was already raised by me and we agreed [to use plain text emails in the MVP](https://github.com/freeCodeCamp/chapter/issues/290#issuecomment-565839659)

Zeko369:
To answer other question. Yes we're using `nodemailer` to send emails, we'll maybe add mailgun or mailchimp "config templates" for it, so it's easier to integrate with those services.

allella:
Thanks I forgot about #290. So, [just plain text in nodemailer](https://nodemailer.com/message/), but if we need to insert dynamic pieces, like unique hashes to opt out of emails, then are those going to be hard coded text strings as the message or would we still be using a template file with placeholders variables that will be merged in?

It seems any of these 3rd party mail services should take a plain-text message as part of their SMTP, which I think is part of the SMTP spec, so perhaps the config templates are a post-MVP feature.

@emred2700 if you have any old reminder emails from event services, like Meetup.com, Facebook, or whatever is popular where you are, then how about posting a screenshot of those to seed the conversation on text. I'll try to do the same.

Zeko369:
We can just use a template literal to create a email "function" instead of using HTML templates for now.

```js
const email = (user, link) => `
Hello ${user.name},

Go to ${link},

Bye
`
```

allella:
Thanks. I was wondering if there as a plain-text template since folks usually don't like to mix content and logic. Though, for MVP what you posted looks expedient.

Zeko369:
The `email-templates` lib you mentioned can do that, but if we're going to setup a templating library I think that we should go all the way and setup HTML rendering

Zeko369:
[...]

PR #640, opened 2021-06-24
feat: email all interested users on event creation

ojeytonwilliams:
The UI is a bit horrible, but this PR adds another button to /dashboard/events/[id] which sends emails to everyone in that chapter.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,3 +20,6 @@ EMAIL_HOST=localhost
 # change this variable to at least 32 random characters
 # to enforce 256-bit cryto security on Auth0's JsonWebToken 
 JWT_SECRET=SetThisTo32orMoreRandomCharacters
+
+# When running remotely, this should be the full url of the landing page
+CLIENT_LOCATION=http://localhost:3000
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -1803,6 +1803,39 @@
             "deprecationReason": null
           },
           {
+            "name": "sendEventInvite",
+            "description": null,
+            "args": [
+              {
+                "name": "id",
+                "description": null,
+                "type": {
+                  "kind": "NON_NULL",
+                  "name": null,
+                  "ofType": {
+                    "kind": "SCALAR",
+                    "name": "Int",
+                    "ofType": null
+                  }
+                },
+                "defaultValue": null,
+                "isDeprecated": false,
+                "deprecationReason": null
+              }
+            ],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Boolean",
+                "ofType": null
+              }
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
             "name": "sendEmail",
             "description": null,
             "args": [
--- a/client/src/generated/graphql.tsx
+++ b/client/src/generated/graphql.tsx
@@ -143,6 +143,7 @@ export type Mutation = {
   updateEvent: Event;
   cancelEvent: Event;
   deleteEvent: Scalars['Boolean'];
+  sendEventInvite: Scalars['Boolean'];
   sendEmail: Email;
   register: User;
   login: LoginType;
@@ -201,6 +202,10 @@ export type MutationCancelEventArgs = {
 };
 
 export type MutationDeleteEventArgs = {
+  id: Scalars['Int'];
+};
+
+export type MutationSendEventInviteArgs = {
   id: Scalars['Int'];
 };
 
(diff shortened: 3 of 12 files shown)


allella:
@ojeytonwilliams I suppose it depends on if this PR is targeting #107  or #99 

#107 seems like it would automatically have the system send emails to _members_ who have previously attended an _event_.

#99 would allow the _organizer_ to email _members_ of a _chapter_ manually and for any reason.

ojeytonwilliams:
True.  I think what I'll do is target #107, since this only allows you to email a canned message.  I'll need to change a few things, but it's not tricky.

It shouldn't take much effort (on the server side, at least) to handle #99, but the UI would definitely take some work.  That would make sense as a separate PR.

allella:
@ojeytonwilliams we have a couple open MVP issues for the text of emails.

#276 is for reminders to go out [2 days prior to the event](https://github.com/freeCodeCamp/chapter/issues/276). 

We have [sample text _event_](https://github.com/freeCodeCamp/chapter/issues/276#issuecomment-596913322) for that use case that can likely be used and tweaked for the initial RSVP and possibly for other messages

allella:
@ojeytonwilliams if /as you, or others are working on the email, then the previous conversations about following / unfollowing a Chapter and unsubscribing from emails should provide good context on the type of links and privacy settings we imagined to make for easy opt-in and opt-out.

That conversation also goes into interfaces on the website and name of buttons, type of buttons and such.

For the emails, we also touched slightly on the security aspects of allowing _members_ to unsubscribe or opt-out without needing to login. For that, something like a secure / unique hash was discussed. Whether that anonymous / hash ideas is the correct approach, or easy to do is up for debate, but I thought to mention it.

The core of those email and UI opt-in / out conversations were on #107 and I tried to summarize the conversation at a few points, like:

https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-596905797
https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-562374895
https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-562432174

ojeytonwilliams:
Thanks for the context, Jim, that's really helpful.  For this PR, I'll keep the scope very limited.  It's just going to send emails to users whenever an event is created if they have `user_chapter_roles.interested` set to TRUE for that chapter.

Does that sound reasonable?  As far as I could see, the discussion was around scheduled notifications, but I think that someone who has set "Unfollow {chapters.name} Notifications" to FALSE would assume that that chapter would stop notifying them entirely.

allella:
@ojeytonwilliams yes, we wanted to account for privacy on emails for an _rsvp_ separately from general updates for the _chapter_ itself, like a new _event_ or even a notice from the _organizer_. 

user_chapter_role.interested and rsvp.interested were added in #374.

[user_chapters was the original table for the _chapter_ level setting, but Timmy Chen removed user_chapters](https://github.com/freeCodeCamp/chapter/pull/335) since [user_chapter_roles](https://github.com/freeCodeCamp/chapter/pull/361) gave us more flexibility on permissions w

I still see [_user_chapters_ in the schema](https://freecodecamp.github.io/chapter/index.html), so I'll have to see if that goes away the next time it's run. I'm wondering if the migrations never deleted _user_chapters_.

ojeytonwilliams:
Okay, cool. I think this is behaving how we want, then.

Regarding *user_chapters*, yes, there's still a migration to create it.  One thing I was considering was that, since we're still in development, it might be worth periodically clearing out the migrations once we're happy that `db:sync` is working.  We've already accrued a fair chunk of migrations and at least one of them (db/migrations/1575817782077-UserChapter.ts) shouldn't exist.

allella:
Sounds good on consolidating the dev migrations. I don't know the best practices on that so I'll defer to others on if / when is a good time to consolidate or clean the migrations.

I believe it's safe to remove _db/migrations/1575817782077-UserChapter.ts_ since there are no models and per #335.  Do you want to do that as part of this PR, or in a new one?

ojeytonwilliams:
> I don't know the best practices on that so I'll defer to others on if / when is a good time to consolidate or clean the migrations.

Neither do I, to be honest.  @Zeko369 do you know what's a reasonable practice here?  I know you pointed out that creating migrations so that we can validate them is a good idea.  As I think about this, I suspect that periodically running `db:drop` and `db:sync` and comparing that with the result of `db:drop` and `db:migrate` could be helpful. That should help us make sure that the migrations are all doing what we want.

Just for the sake of keeping this in scope, I'll create another PR to remove that migration.

ojeytonwilliams:
@allella to keep the number of PRs manageable, I'm going to merge this.  While it doesn't satisfy the user story, I think it's a step in the right direction.

PR #641, opened 2021-06-25
feat: email organizer on RSVP

ojeytonwilliams:
Before the organizer could be emailed, they needed to be added to the Event, so I migrated the DB to add a User to Event.

Since I had to modify `createEvent` I went ahead and restricted it to `organizers`. I'll do the same for the other mutations in another PR.


allella:
@ojeytonwilliams we have the [user_chapter_roles](https://freecodecamp.github.io/chapter/tables/user_chapter_roles.html) table

Would that be sufficient to infer the _organizer's_ relationship to the _chapter_ where the events don't need a specific organizer?  Also, user_chapter_roles would allow for co-_organizers_ where linking an _organizer_ to an _event_ limits to having just one organizer for an event.

I could see a situation where there's a 1-to-1 between organizer and event, but I think historically with systems like Meetup, the events are managed by any and all co-organizers.

allella:
For instance, we have three co-organizers for _Code For Greenville_ and it's inferred that we're all attending and managing any new events.

allella:
https://www.meetup.com/Code-for-Greenville/events/qwpbksycckbjb/

ojeytonwilliams:
Thanks for the feedback, @allella.  To be honest, I wasn't 100% sure what the best relationship would be, and what you're saying makes sense.  I'll give it some thought and try to re-organise things to use that table.

ojeytonwilliams:
Hey @allella, after a bit of reflection, I think we need to have at least one organiser per event (and, yeah, definitely the option to have more than one).  If one isn't provided, one could definitely be inferred from the Chapter's *chapter_roles*.

With that in mind, I think there are two reasonable options:

1. Create a custom relationship like for RSVP
2. Create a many-to-many relationship between Users and Events

I don't like the idea of modifying UserChapterRole to include Events, because I think it's useful to separate the concerns: 

UserChapterRole would only deal with the relationship between Users, Chapters and their roles within them.
Users would then have an *events_organized* relationship to Events and Events would have an *organizers* relationship to Users.  Since all we're trying to represent is the organisers for events and the events someone organises, I think a straightforward many-to-many relationship would suffice.

Let me know what you think, but I reckon option 2. should be fine.

allella:
@ojeytonwilliams I'm going to think through this a bit more, but I think this is a good catch that having event roles will be necessary.

If you wanted to do something quick and simple for the demo, then a boolean flag in _rsvp.is_organizer_ (or whatever field name would be consistent with existing booleans) would probably save time.

Though, to take your #2 a bit further, we could have an even more flexible _user_event_roles_ to mirror the pattern of _user_chapter_roles_.

By not limiting the table to just _organizers_ we'd gain the ability to define other _event_ roles without needing more tables down the road.

**Potential Event Roles**
- a _user_ is an organizer_
- a _user_ is a presenter / speaker
- a user is representing a sponsor - I started to think we could get rid of _event_sponsors_ table in the later case, but _sponsors_ in that context are not users and have other non-user fields like logos and websites.
- This is a new one on me, but I see [Drupal events](https://www.drupal.org/community/events/drupal-camp-asheville-2021-2021-07-08) even have a "Code of Conduct Contact", so we'd be able to deal with the miscellaneous event roles.
- a _user_ who is a _member_ of a _chapter_ may have some future application in this table

**_user_event_roles_** table

- id
- event_id
- user_id
- role_name
- interested - this field exists in _user_chapter_roles_, so it got me thinking if we'd have it in this table. A  _user_  with a role of _member_ currently indicates interest in an _event_ through _rsvps.interested_, which seems a bit strange to create an _rsvp_ record just for interest.  Though, if we track interest here, then once a _member_ RSVPs you'd have records about the member's relationship to the event in two places.  Worth more thought, but for now perhaps we leave this field out.
- other accessory fields similar to [_user_chapter_roles_](https://freecodecamp.github.io/chapter/tables/user_chapter_roles.html)

ojeytonwilliams:
@allella I like the idea for this table.  So, just to make sure we're on the same page, the idea would be to:

Create a many-to-many relationship between _users_ and _events_ analogous to the existing relationship between users and chapters.  This relationship would be captured by a _user_event_roles_ table that has foreign keys to _users_ and _events_.  

Presumably, if we want a user to have multiple roles for each event, then we'd need _role_name_ to be part of the primary key, too.  Otherwise we could end up with someone organising the same event twice.

However, wouldn't including _interested_ mean that each _user_, _role_name_ combination could have a different _interested_ value? Do we want it to be possible for, say, a speaker who is also an organiser to opt in to speaker emails and out of organiser emails?  If so, no worries, I just wanted to check.

allella:
@ojeytonwilliams that sounds good.

I think we can leave the _interested_ field of this new table, at least for now.

If we find a need for an opt-in / opt-out field for special event roles, like a sponsor or speaker, then we might choose a different field name based on the context.

The word "Interested" makes some sense for _user_chapter_roles_ and _rsvps_, but I'd also get on board with changing those fields names. I think we called them "subscribed" in an earlier iteration.

ojeytonwilliams:
Yep, I agree, "subscribed" seems reasonable given how we're using it. "interested" is a bit vague, but if I subscribe to an event I'm explicitly asking for notifications.

I'll create the _user_event_roles_ table (without _interested_) in this PR, but renaming _interested_ feels a bit out of scope, so I'll do that separately.

nhcarrigan (approved):
This is looking good.

The `db:seed` error is fixed, and when I RSVP to an event Mailhog captures an email:

image

ojeytonwilliams:
Thanks for the review, Nick. Let's get this in.

PR #642, opened 2021-06-25
feat: email users on venue change

### LINKED ISSUE #663: As a user, when an organizer changes an event's venue, I will receive an email if I have RSVP'd

ojeytonwilliams:
This emails the users that have RVSP'd whenever the venue is changed.

Currently the tests only check that the venue change happens and that emails about it are sent.  It does not, yet, confirm that only the RVSPs get emailed, but I plan to circle back to that test later.

Closes #663

### DIFF
--- a/client/src/modules/dashboard/Events/pages/EditEventPage.tsx
+++ b/client/src/modules/dashboard/Events/pages/EditEventPage.tsx
@@ -37,6 +37,7 @@ export const EditEventPage: NextPage = () => {
       const eventData = {
         ...data,
         capacity: parseInt(String(data.capacity)),
+        venueId: parseInt(String(data.venueId)),
         start_at: new Date(data.start_at).toISOString(),
         ends_at: new Date(data.ends_at).toISOString(),
         tags: undefined,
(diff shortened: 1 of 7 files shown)


allella:
@ojeytonwilliams this seems like a smart idea. I'm not sure we have an [MVP issue](https://github.com/freeCodeCamp/chapter/issues?page=1&q=is%3Aissue+is%3Aopen+label%3AMVP) for it, but if you already have it worked up, then we can create a user story if needed.

Also, I think it would be helpful to link the PRs to the MVP issue / story if it's not too much trouble.

Thanks and nice work.

ojeytonwilliams:
Sounds good.  I've been focusing on the features that @Zeko369 outlined for the demo video, since they seemed both urgent and good things to have in the MVP.

And, yes, I agree these should all be linked to MVP issues and stories.  I'll add all of those for the PRs I've recently created.

Zeko369 (approved):

ojeytonwilliams:
Given that this adds a single `organizer` to the db, I think this should wait until we've decided on and implemented the ideas discussed here: https://github.com/freeCodeCamp/chapter/pull/641#issuecomment-870761351

ojeytonwilliams:
There are no longer any changes to the db, since the recently added _user_event_roles_ table captures the relationship I was trying to add with the _organizer_.

PR #648, opened 2021-06-29
fix: remove redundant user_chapter table

ojeytonwilliams:
There is no longer a many to many relationship between *users* and *chapters* since *user_chapter_role* exists to cover this relationship.

Recent discussion: https://github.com/freeCodeCamp/chapter/pull/640#issuecomment-870267360


allella:
Approve / agree.

PR #651, opened 2021-06-30
Create event images

ahmadabdolsaheb:
- [] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md).
- [] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [] My pull request targets the `master` branch of Chapter.

Closes #XXXXX

PR #652, opened 2021-06-30
feat: add event images

ojeytonwilliams:
Recreated after I accidentally merged the last one: https://github.com/freeCodeCamp/chapter/pull/651

### DIFF
--- a/client/src/modules/dashboard/Events/components/EventFormUtils.ts
+++ b/client/src/modules/dashboard/Events/components/EventFormUtils.ts
@@ -22,6 +22,12 @@ export const fields: Field[] = [
     placeholder: '',
   },
   {
+    key: 'image',
+    type: 'text',
+    label: 'Event title',
+    placeholder: 'Foo and the Bars',
+  },
+  {
     key: 'url',
     type: 'text',
     label: 'Url',
@@ -65,6 +71,7 @@ export interface EventFormData {
   name: string;
   description: string;
   url?: string | null;
+  image: string;
   video_url?: string | null;
   capacity: number;
   tags: string;
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -3,6 +3,7 @@ import { NextPage } from 'next';
 import {
   Heading,
   VStack,
+  Image,
   Text,
   Button,
   useToast,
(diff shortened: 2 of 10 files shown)


client/src/modules/events/pages/eventPage.tsx
```diff
@@ -100,6 +101,14 @@ export const EventPage: NextPage = () => {
         userIds={data?.event?.rsvps.map((r) => r.user.id) || []}
         modalProps={modalProps}
       />
+      <Image
+        boxSize="100%"
+        maxH="300px"
+        src={data.event.image}
+        alt="The freeCodeCamp logo"
```
ojeytonwilliams, line 108: Can we add the alt text when adding the image and then use it here?

client/src/modules/dashboard/Events/components/EventFormUtils.ts
```diff
@@ -21,6 +21,12 @@ export const fields: Field[] = [
     label: 'Description',
     placeholder: '',
   },
+  {
+    key: 'image',
+    type: 'text',
+    label: 'Event title',
+    placeholder: 'Foo and the Bars',
```
ojeytonwilliams, line 28:
```suggestion
    label: 'Event image',
    placeholder: 'https://www.example.image/url',
```

ojeytonwilliams (commented):
@ahmadabdolsaheb this is technically my PR now, but what do think to the suggestions?

client/src/modules/events/pages/eventPage.tsx
ahmadabdolsaheb: Sure a field called description should do it.

ojeytonwilliams: Thinking about it, this is purely decorative. We can just require `alt=""` since screenreaders can safely ignore it.

server/db/migrations/1632132998306-RenameImageToImageURL.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
ahmadabdolsaheb, line 1: Not sure, but since we are introducing "image_url" for first time here we would not need a  migration  file?

ojeytonwilliams: Nope, we're not using explicit migration files now, since typeorm can set up the db for us without them.  I'll remove this file while I sort out the conflicts.

PR #650, opened 2021-06-30
Add chapter and event images

ahmadabdolsaheb:
- [] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md).
- [] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [] My pull request targets the `master` branch of Chapter.

Closes #XXXXX

PR #662, opened 2021-07-05
feat/add images to cards

ahmadabdolsaheb:

We would still need to make the chapter cards more readable and bring in the attendee and member info to the cards, but this is something to start from.

Screen Shot 2021-07-05 at 11 31 11 PM
Screen Shot 2021-07-05 at 11 35 26 PM

### DIFF
--- a/client/src/components/ChapterCard.tsx
+++ b/client/src/components/ChapterCard.tsx
@@ -1,30 +1,58 @@
 import React from 'react';
-import { Heading, Tag, Text, VStack } from '@chakra-ui/react';
+import {
+  Stack,
+  Heading,
+  Box,
+  Center,
+  Image,
+  Text,
+  useColorModeValue,
+} from '@chakra-ui/react';
 import { Link } from 'chakra-next-link';
 import { Chapter } from 'generated/graphql';
-
-import { Card } from './Card';
-import { truncate } from '../helpers/truncate';
 
 type ChapterCardProps = {
   chapter: Pick<
     Chapter,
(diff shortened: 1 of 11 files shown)


ojeytonwilliams:
Should be merged after https://github.com/freeCodeCamp/chapter/pull/650 and https://github.com/freeCodeCamp/chapter/pull/652

client/src/components/ChapterCard.tsx
```diff
@@ -1,30 +1,57 @@
...
+    <Center m={2}>
+      <Link href={`/chapters/${chapter.id}`} _hover={{}}>
+        <Box
+          w={'full'}
+          bg={useColorModeValue('white', 'gray.800')}
+          rounded={'md'}
+          overflow={'hidden'}
+        >
+          <Image
+            h={'168px'}
```
ojeytonwilliams, line 32: Is there a good reason to use `{'168px'}` over `"168px"`?  To me, the former suggests "thing to be evaluated" and the latter just "thing".

Obviously it's not a big deal, but just quotes seems preferable and is consistent with the rest of the codebase.

allella: Double-quote was the common style in earlier days. It seems single quote became popular with HTML 5, so it could be an old habit for someone.

ojeytonwilliams: I'm not quite sure *why*, but eslint is enforcing the double quotes... but only for react :man_shrugging:

I've been meaning to separate eslint and prettier.  When I do that, I'll try and figure out why it's being inconsistent with quotes.

PR #669, opened 2021-07-28
fix: standardizes installation to just use npm i

ojeytonwilliams:

Rather than using `npx recursive-install` I changed the postInstall script to install the client.  That way we get a friendlier output (Welcome to Chapter, etc) and sufficient to use `npm i` to get up and running.

### DIFF
--- a/.github/workflows/ci-workflow.yaml
+++ b/.github/workflows/ci-workflow.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Update npm
         run: npm i -g npm@7
       - name: Install dependencies
-        run: npx recursive-install
+        run: npm i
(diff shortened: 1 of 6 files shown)


allella:
@timmyichen had suggested _recursive-install_ in a [previous comment(https://github.com/freeCodeCamp/chapter/pull/403#pullrequestreview-488333433).

It sounds like there was a minor improvement with directory structure, but I'm not sure if it's worth keeping?

timmyichen:
OP's solution is much more clean and easier for everyone to use, I'd just use that :)

ojeytonwilliams:
Thanks @timmyichen, @allella. Let's get this in, then.

PR #678, opened 2021-07-31
fix: check IS_DOCKER against 'true' rather than ''

ojeytonwilliams:
It seems more intuitive to have IS_DOCKER be either 'true' or 'false' than 'true' or ''


allella:
Looks good to me.

PR #684, opened 2021-08-06
chore: move all server code into separate dir

ojeytonwilliams:
- chore: move db inside server
- chore: move typeorm inside server
- chore: move server deps into its package.json
- chore: fix ts-node/ts-node-dev issues
- chore: move tsconfig into server
- fix: update docker config


ojeytonwilliams:
There's an argument for keeping the db dir in root, since it contains dev scripts and isn't technically part of the server.  The rationale for moving it is that typeorm *is* part of the server and so it makes sense (and simplifies the config) if all the db-related things are in the same place.

ojeytonwilliams:
The plan is to make one more change - pulling all the server source into a `src` directory to mirror the client.  Then it should be straightforward to compile the TS files into a `dest` dir and use that for production.  I think this PR does enough, though.

allella:
Above my pay-grade, but double checking the docs changes.

Is the following part of CONTRIBUTING.md impacted by these changes?

> * The server will automatically restart anytime you save a _.ts_ or _.js_ file within the _server/_ directory.

ojeytonwilliams:
@allella it shouldn't, but I have broken the Docker setup by moving things around.  Since I don't use it myself, I'm likely to keep breaking it.  

With that in mind, I'll add a parallel CI job to run the Cypress tests against a container.  This is probably excessive, but it's an easy way to check it's working.

.github/workflows/cypress-docker.yaml
```diff
@@ -0,0 +1,57 @@
...
+          browser: ${{ matrix.browsers }}
+
+      - name: Check running containers
+        run: docker ps -a
+      - name: Check logs
+        run: | 
+          docker logs chapter_app_1
+          docker logs chapter_client_1
+
+// TODO: it's waiting for ports that never open.
```
ojeytonwilliams, line 57:
```suggestion
```

ojeytonwilliams:
Closing in favour of https://github.com/freeCodeCamp/chapter/pull/685

PR #685, opened 2021-08-10
feat: compile ts incrementally

ojeytonwilliams:
This supersedes https://github.com/freeCodeCamp/chapter/pull/684 because I realised that in order to test the docker build in CI, the tests needed to be much faster.

**What this does**

In a nutshell, it replaces ts-node with node.  Instead of compiling TypeScript on the fly and on each call to `npm run db:reset` it compiles it once and only re-compiles when a source file has changed.  This makes the dev experience (and tests) a lot snappier.  `db:reset` is roughly 10 times quicker now, for instance.

Also, this makes the production environment a little quicker, too, since we can compile the TS before starting the server.  In principle ts-node can do this, too, but I'd rather cut out the intermediary and just use js directly.

**Docker changes**

The dev experience should be unchanged, but now we have a production-like version, and Cypress tests to run against it.  It would be nice to check the dev-mode in CI, but it's painfully slow to the point that Cypress would need a much larger timeout to pass.  As such, it just tests the production-like version.  The main difference between that and dev-mode is that dev-mode uses local volumes and `npm run dev` so that code changes are watched.

**Reorganisation**

The changes in https://github.com/freeCodeCamp/chapter/pull/684 are included.


CONTRIBUTING.md
```diff
@@ -512,7 +512,7 @@ If you prefer to run some or all of the steps manually, then they are:
 
 This would create _ModelName.ts_ in _server/models_
 
-To keep everything DRY, add `extends BaseModel` to the class and import it from 'server/models/BaseModel' to no repeat id, createdAt, and updatedAt fields on every single model
+To keep everything DRY, add `extends BaseModel` to the class and import it from 'models/BaseModel' to no repeat id, createdAt, and updatedAt fields on every single model
 
 You could also run `npx typeorm` since here you're not actually loading any ts files, but because regular `npx typeorm` runs inside of node it import from _.ts_ files, so we run it with `ts-node` and our custom server config (check package.json)
```
allella, line 517: @ojeytonwilliams there's a reference here to ts-node. Your comment mentioned replacing ts-node with node, so I thought to ask if this needs updating.

allella: @ojeytonwilliams  Also,  just above this section is `npm run typeorm entity:create -- --name=ModelName` and it looks like _typeorm_ has been removed from package.json, so I figure this whole section may need updating but it's beyond my understanding to properly make those updates.

ojeytonwilliams: Thanks @allella, those did need changing.  I've simplified the scripts a bit and recreated the `typeorm` script so that that command still works.

I also made a few more docs tweaks while I was in there.

ojeytonwilliams: The commit messages are intended to explain the changes. If they aren't clear, I can try and justify them a bit better.

PR #690, opened 2021-08-18
docs: add a basic guide to the tech stack

ojeytonwilliams:
The hope is that this should help with onboarding new contributors as well as reminding longer standing ones.  One of the biggest barriers I had to contributing was figuring out what the various parts of the tech stack were responsible for and how they interacted.

This should help people get up to speed quickly so they can actually make stuff.

If https://github.com/freeCodeCamp/chapter/pull/685 goes in, the directories will need updating to reflect those changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -414,7 +414,7 @@ The Chapter client uses the React framework [Next.js](https://nextjs.org/) with 
 
 The database and GraphQL schema are defined by files in _server/models_
 
-The resolvers for the GraphQL queries are defined in _server/controllers_
+The resolvers for the GraphQL queries are defined in _server/src/controllers_
 
 The client accesses the data via hooks defined in _client/src/generated/generated.tsx_
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,8 @@
 - [Running the Application](#running-the-application)
   - [Docker Mode](#running-the-application)
   - [Manual Mode](#running-the-application)
+- [Adding a New Feature](#adding-a-new-feature)
+  - [Where to find the code](#where-to-find-the-code)
 - [Frequently Asked Questions](#frequently-asked-questions)
 - [Server-side Technical Documentation](#server-side-technical-documentation)
   - [API Specification](#api-specification)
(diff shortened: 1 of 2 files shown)


ojeytonwilliams:
@allella let me know if this reads okay and you think it's a worthwhile addition.  I was thinking it would help, particularly since I've highlighted some beginner friendly issues with https://github.com/freeCodeCamp/chapter/issues/691

CONTRIBUTING.md
```diff
@@ -400,6 +402,28 @@ Once the app has started you should be able to pull up these URLs in your web br
...
+
+The Chapter client uses the React framework [Next.js](https://nextjs.org/) with [Apollo Client](https://www.apollographql.com/docs/react/) for data fetching.  Since we are generating a GraphQL schema we can use [GraphQL Code Generator](https://www.graphql-code-generator.com/) to convert the schema into a set of TypeScript types and, more importantly, functions to get the data from the server.  As a result we know exactly what we're allowed to request from the server and the shape of the data it returns.
+</details>
+
+## Where to find the code
+
+The database and GraphQL schema are defined by files in _server/models_
+
+The resolvers for the GraphQL queries are defined in _server/controllers_
```
ojeytonwilliams, line 420:
```suggestion
The resolvers for the GraphQL queries are defined in _server/src/controllers_
```

allella:
@ojeytonwilliams I think this is a fine addition.  I may tweak some of the content a bit, but nothing to hold up merging it in.

My main question is if we should consolidate the [Tech Stack section from README](https://github.com/freeCodeCamp/chapter#tech-stack) and/or link and this new section of CONTRIBUTING.md.

I don't think we want to maintain the exact same content in two places since it's likely we'll forget to update both.  Though, having a very high-level set of technology bullet points, like we have in the README, and then linking from that section directly to this more verbose section of the CONTRIBUTING.md could be a middle-ground that mostly avoids duplication but still gives a highlight in the README.

Thoughts?

CONTRIBUTING.md
```diff
@@ -400,6 +402,28 @@ Once the app has started you should be able to pull up these URLs in your web br
 
 </details>
 
+# Adding a New Feature
+
+In order to understand where to start, it may help to familiarise yourself with our tech stack.  The key technologies we use are: PostgreSQL, TypeORM, Apollo, Express, React and Next.js.  For more details:
```
ojeytonwilliams, line 407:
```suggestion
In order to understand where to start, it may help to familiarize yourself with our [tech stack](https://github.com/freeCodeCamp/chapter/blob/main/README.md#tech-stack). For more details:
```
How about this?  Does it seem reasonable to have this link and a link back from the README?

allella: Sure, if you want to keep the bullet list for the Tech Stack in the README, then linking to/from each would be good.

ojeytonwilliams:
@allella I think this is fine, so I'm merging, but if I missed something just let me know and I'll fix it in another PR.

PR #709, opened 2021-09-17
docs: Update CONTRIBUTING.md to require npm 7

### LINKED ISSUE #705: fsevents Breaks scripts/postInstall.js When Using NPM Version 6 on Linux or Windows
### Problem
The inclusion of _fsevents_, which is a Mac specific tool, is breaking the _scripts/postInstall.js_ when npm i is run after the initial clone when npm version is 6.  npm version 7 doesn't have the issue.

image

I suspect the root of the problem could be related to the lock file being on version 2, [even though version 1 and 2 are supposed to be compatible](https://github.com/npm/cli/issues/724).

_lockfileVersion 1_ is used for npm 6 and _lockfileVersion 2_ is for npm 7.

The [renovateBot is generating lock files with _"lockfileVersion": 2_](https://github.com/freeCodeCamp/chapter/blob/main/server/package-lock.json) and I'm using npm 6.

**Environment**
Operating System:  Linux and Windows
NPM Version: 6

[Confirmed by @allella (Linux)and Ravi (Windows) in the chat](https://chat.freecodecamp.org/channel/chapter?msg=hD2ACS8yC7SPNGJY2)

### Workaround
rm server/package-lock.json
npm i

### Possible Solution
Require NPM version 7
Remove fsevents
Patch or otherwise prevent fsevents compatibility checking from breaking in npm 6 (when _lockfileVersion 1_)
Change RenovateBot to use npm 6

allella:

Update a line in the docs that says npm 6 is required.

We now require npm 7 to work around a lockVersion 1 bug mentioned in https://github.com/freeCodeCamp/chapter/issues/705

Closes #705 

@ojeytonwilliams I know you updated another spot in the docs, but this was a different spot.  Also, I believe you made changes to encourage / force npm 7, so I think that will workaround #705.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -322,7 +322,7 @@ You have successfully created a PR. Congratulations! :tada:
 Ensure the Node.js tools are installed:
 
 * Node.js 14 or greater - `node --version` and the output should be like **v14**.16.0
-* npm 6 or greater - `npm --version` and the output should be like **6**.14.11
+* npm 7 or greater - `npm --version` and the output should be like **7**.23.0
 
 Run `npm i` to install all of the necessary dependencies.
 


ojeytonwilliams:
Ooops, good catch.  Thanks @allella

PR #713, opened 2021-09-24
chore: minor tweaks and fixes

ojeytonwilliams:

The only moderately non-trivial change is forbidding implicit `any`, but it turns out we did not do that very much.  I'd like to forbid explicit `any`, too, but there are a lot more of those!


ojeytonwilliams:
There are a few more implicit `any`s than I thought!  Blocked until I figure out if there's a quick fix and/or if it's worth it.

PR #598, opened 2021-04-27
chore(client): Introduces the CodeSee devtool into the dev environment

jleven:
- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/master/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `master` branch of Chapter.

Introduces the CodeSee devtool into the dev environment. (@ojeytonwilliams is awaiting this PR.)

This involves adding two dev dependencies (@codesee/babel-plugin-instrument and @codesee/tracker), and adding the babel plugin to the client babel configuration.

This change should have no effect on the production build or production running application.

--- 

CodeSee helps you visually understand how your codebase works, document it, and communicate about it with others.

When running the app in development, developers can make and view recordings of the running application, then explore visualizations of those recordings, seeing every line of code that ran, as well as the associated runtime data. They can also add notes directly linked to the code. More details at [docs.codesee.io](docs.codesee.io).

---

CodeSee is currently in closed beta, but they will provide accounts to anyone on this project asap upon request. Sign up for the beta at [codesee.io](codesee.io) or email support@codesee.io for an account.

PR #714, opened 2021-09-30
Introduces the CodeSee devtool into the dev environment

ojeytonwilliams:
CodeSee helps you visually understand how your codebase
works, document it, and communicate about it with others.

When running the app in development, developers can make
and view recordings of the running application, then
explore visualizations of those recordings, seeing every
line of code that ran, as well as the associated runtime
data. They can also add notes directly linked to the code

Developers can make recordings by clicking the floating
"eye" to start and stop a recording.

### DIFF
--- a/client/babel.config.js
+++ b/client/babel.config.js
@@ -1,0 +1,10 @@
+var config = {
+  presets: ['next/babel'],
+  plugins: [],
+};
+
+if (process.env.NODE_ENV === 'development') {
+  config.plugins.push(['@codesee/instrument', { hosted: true }]);
+}
+
+module.exports = config;
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -28,6 +28,8 @@
         "react-hook-form": "7.16.2"
       },
       "devDependencies": {
+        "@codesee/babel-plugin-instrument": "^0.101.0",
+        "@codesee/tracker": "^0.101.0",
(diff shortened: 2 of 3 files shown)


ojeytonwilliams:
For some reason I couldn't rebase and push https://github.com/freeCodeCamp/chapter/pull/598 so I recreated it :man_shrugging:

ojeytonwilliams:
@allella I've been using this a fair bit on the freeCodeCamp repo, and I think it would help us a bit.

I added a (very) brief intro in our docs, mostly just linking to their docs.

CONTRIBUTING.md
```diff
@@ -5,8 +5,8 @@
   - [Using GitPod](#using-gitpod)
   - [Using a Traditional Dev Environment](#using-a-traditional-dev-environment)
 - [Running the Application](#running-the-application)
-  - [Docker Mode](#running-the-application)
```
allella, line 8: The #docker-mode and #manual-mode don't work as in-page anchors since they are inside of a collapsible block.

It may be possible to rework the collapsible headings with HTML or something, but I messed with it a bit and having it jump to the parent section was the best quick fix I could come up with.

In short, leaving these as #running-the-application may be best for now.

ojeytonwilliams: Ooops.  VSCode keeps 'fixing' that when I save. I'll revert it.

PR #718, opened 2021-10-02
Created a setUserInterestForChapter mutation and Added placeholder for unsubscribe

### LINKED ISSUE #107: As a user / future participant after the event, I will automatically get emails notifying me of subsequent events.

Ravichandra-C:

Closes #107 

Features Added:
1. Created a setUserInterestForChapter Mutation and called it when user rsvp to an event
2. Added place holder for unsubscribe link in all the event related emails sent from the chapter.

File wise changes:

1. **server/src/controllers/UserChapterRole/resolver.ts** : Created for setUserInterestForChapter Mutation.
2. **client/src/generated/graphql.tsx**: Created the Type definitions of the setUserInterestForChapter Mutation .
3. **client/src/modules/dashboard/Events/graphql/queries.ts** : Added the client side mutation code. 
4. **client/src/modules/events/pages/eventPage.tsx** : Called the mutation when user rsvp to the Event.

### DIFF
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -19,7 +19,11 @@ import React, { useMemo } from 'react';
 import { LoginRegisterModal } from '../../../components/LoginRegisterModal';
 import { useAuth } from '../../auth/store';
 import { EVENT } from '../../dashboard/Events/graphql/queries';
-import { useEventQuery, useRsvpToEventMutation } from 'generated/graphql';
+import {
+  useEventQuery,
+  useRsvpToEventMutation,
+  useSetUserInterestForChapter,
+} from 'generated/graphql';
 import { useParam } from 'hooks/useParam';
 
 export const EventPage: NextPage = () => {
@@ -27,6 +31,7 @@ export const EventPage: NextPage = () => {
   const { user } = useAuth();
 
   const [rsvpToEvent] = useRsvpToEventMutation();
+  const [setUserInterestForChapter] = useSetUserInterestForChapter();
   const { loading, error, data } = useEventQuery({
     variables: { id: id || -1 },
   });
@@ -80,6 +85,11 @@ export const EventPage: NextPage = () => {
               }
             : { title: 'You canceled your RSVP 👋', status: 'error' },
         );
+        if (add) {
+          await setUserInterestForChapter({
+            variables: { event_id: 3 },
+          });
+        }
       } catch (err) {
         toast({ title: 'Something went wrong', status: 'error' });
         console.error(err);
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -87,7 +87,7 @@ export const EventPage: NextPage = () => {
         );
         if (add) {
           await initUserInterestForChapter({
-            variables: { event_id: 3 },
+            variables: { event_id: id },
           });
         }
       } catch (err) {
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -22,7 +22,7 @@ import { EVENT } from '../../dashboard/Events/graphql/queries';
 import {
   useEventQuery,
   useRsvpToEventMutation,
-  useSetUserInterestForChapter,
+  useinitUserInterestForChapter,
 } from 'generated/graphql';
 import { useParam } from 'hooks/useParam';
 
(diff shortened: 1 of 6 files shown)


server/src/controllers/UserChapterRole/resolver.ts
```diff
@@ -0,0 +1,37 @@
+import { Arg, Ctx, Int, Mutation, Resolver } from 'type-graphql';
+import { GQLCtx } from 'src/common-types/gql';
+import { Event, UserChapterRole } from 'src/models';
+
+@Resolver()
+export class UserChapterRoleResolver {
+  @Mutation(() => UserChapterRole)
```
ojeytonwilliams, line 7: Just checking, we're returning this so we can display a _interested_ toggle, is that right?

server/src/controllers/UserChapterRole/resolver.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (!event.chapter) {
+      throw Error('Cannot find the chapter of the event with id ' + event_id);
+    }
+
+    const userChapterRole = await UserChapterRole.findOne({
+      where: { user_id: ctx.user.id, chapter_id: event.chapter.id },
+    });
+
+    if (!userChapterRole) {
+      return await new UserChapterRole({
```
ojeytonwilliams, line 28:
```suggestion
      return new UserChapterRole({
```
Nitpick: it should not be necessary to `await` a promise

ojeytonwilliams (changes requested):
Nice work as usual, @Ravichandra-C, just a few minor things:

server/src/controllers/UserChapterRole/resolver.ts
```diff
@@ -0,0 +1,37 @@
+import { Arg, Ctx, Int, Mutation, Resolver } from 'type-graphql';
+import { GQLCtx } from 'src/common-types/gql';
+import { Event, UserChapterRole } from 'src/models';
+
+@Resolver()
+export class UserChapterRoleResolver {
+  @Mutation(() => UserChapterRole)
+  async setUserInterestForChapter(
```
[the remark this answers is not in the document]
Ravichandra-C: @ojeytonwilliams, I named it setUserInterestForChapter , so that in future when we implement the _interested_ toggle functionality, if we want to set interested either to false or true, we could just modify his Mutation to handle that .

server/src/controllers/UserChapterRole/resolver.ts
Ravichandra-C: To be honest, I returned to show that data is updated, but may be we can use it to display the interested toggle .

server/src/controllers/UserChapterRole/resolver.ts
Ravichandra-C: @ojeytonwilliams , Thanks a lot for the suggestions, I'll remove the await and merge the changes.

server/src/controllers/UserChapterRole/resolver.ts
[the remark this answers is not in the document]
ojeytonwilliams: If we have two functions, `set` and `init`, then someone that just reads the function names should realise that `init` is just used to start things up and `set` will modify something.  That way no one using them should be surprised by their behaviour and I think that's quite valuable.

It will make the code a little less DRY, but if we do it right it shouldn't bloat the file too much.

Does that make sense?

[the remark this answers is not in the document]
Ravichandra-C: Yeah, It make sense @ojeytonwilliams . I'll update the name and merge the changes

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -80,6 +85,11 @@ export const EventPage: NextPage = () => {
               }
             : { title: 'You canceled your RSVP 👋', status: 'error' },
         );
+        if (add) {
+          await initUserInterestForChapter({
+            variables: { event_id: 3 },
```
ojeytonwilliams, line 90:
```suggestion
            variables: { event_id: id },
```
Sorry, I missed this the first time!

ojeytonwilliams (changes requested):

client/src/modules/events/pages/eventPage.tsx
Ravichandra-C: @ojeytonwilliams , Oops!, Sorry, I missed it too. I'll update and merge the changes.

ojeytonwilliams (approved):
LGTM :+1: 

Nice work, @Ravichandra-C, I'll get this in once the CI is all green.

Ravichandra-C:
@ojeytonwilliams, Thanks a lot!

PR #726, opened 2021-10-08
adding Form fields for sponsor input

### LINKED ISSUE #96: As an organizer I can add a sponsor to the event with a link to their website as a way of thanking them.
Implementation details:

- The DB already has _event_sponsors_ and _sponsors_ tables
- We need a resolver and client GraphQL queries to access the data
- On /events/[id] we'd need basic UI which just has a link on the event page.  That's enough to move forward
- /dashboard/events/[id]/edit would need an input element to select a sponsor

Since there's been a fair amount of discussion surrounding the sponsor types, we should track those details in a separate issue before closing this.  However, I do think the above implementation would be a solid start.

Ravichandra-C:

Closes #96 

1. Adding Select fields for the Event Form for user to select the sponsor details.
2. Based on the type of the Sponsor the Sponsor Name select form Options will render
3. Added code to Save the Sponsor details to the Database as part of CreateEvent and updateEvent Mutation
4. Added code to Show the Sponsor details when the user edits the Event
5. Added isRequired field for all types in EventFormUtils .tsx
6. Made the tags input field non Mandatory field.
7. changed the default value of image_url while form editing
8. Created a new query to fetch the sponsor details.
9. Modified the cypress test to fetch only the correct select under venue


ojeytonwilliams:
@Ravichandra-C give me a shout if you have any questions or want any feedback, but if you're just tinkering with this in draft mode, that's totally cool.

Ravichandra-C:
@ojeytonwilliams , I'm almost done with the feature. There are a few type checking issues that I'm trying to resolve. Could you please review and let me know if you can find the issue?

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -108,6 +136,88 @@ const EventForm: React.FC<EventFormProps> = (props) => {
...
+                      <h5>loading sponsors</h5>
+                    ) : errorSponsor || !sponsorData ? (
+                      <h5> Error loading sponsors</h5>
+                    ) : (
+                      <Select
+                        defaultValue={getValues(`sponsors.${index}.id`)}
+                        {...register(`sponsors.${index}.id` as const, {
+                          required: true,
+                        })}
+                      >
```
ojeytonwilliams, line 184: I've encountered a problem when running this locally.  I haven't been able to pin down exactly how to reproduce this, but sometimes it tries to send 0 as a sponsor id, which the DB rejects.

Also, this is a bit odd:
SponsorsShifting
 
That's on /dashboard/events/[id]/edit

Finally, if the updates are successfully submitted, you have to reload the page before the sponsor changes appear in the form. Other fields seem to be fine.

ojeytonwilliams (changes requested):
Hey @Ravichandra-C, I found a few oddities with how the sponsors are displaying and updating.  

If you've got an idea what's going wrong, awesome.  If not, it might help to simplify the form and not filter for already selected sponsor names.  We can always revisit that in another PR if the core behaviour is working.

client/src/modules/dashboard/Events/components/EventForm.tsx
Ravichandra-C: >Finally, if the updates are successfully submitted, you have to reload the page before the sponsor changes appear in the form. Other fields seem to be fine.

@ojeytonwilliams , How can we overcome this problem?

Ravichandra-C: >If you've got an idea what's going wrong, awesome. If not, it might help to simplify the form and not filter for already selected sponsor names. We can always revisit that in another PR if the core behaviour is working.

@ojeytonwilliams , I've removed the code for filtering out the selected options.

>but sometimes it tries to send 0 as a sponsor id, which the DB rejects.

I too can't reproduce it. but my best guess is that, when we click on the Add button, we are by default setting the id to 0. and if user Click Save button then, id might be sent as zero. After removing the filter code , I'm not able to reproduce it my local.

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -42,14 +55,29 @@ const EventForm: React.FC<EventFormProps> = (props) => {
...
+  const {
+    fields: sponsors,
+    append,
+    remove,
+  } = useFieldArray({
+    name: 'sponsors',
+    keyName: 'key',
+    control,
   });
+
```
ojeytonwilliams, line 79:
```suggestion
  const firstSponsorType = sponsorData?.sponsors[0]?.type as string | undefined;
  const firstSponsorId = sponsorData?.sponsors[0]?.id as number | undefined;
```
We can use these (see below) to make sure we always add a valid type of sponsor.  That way if a user just clicks `Add` once then `Save Event Changes` it should be valid.  It doesn't handle duplicate sponsors, but one step at a time!

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -108,6 +136,72 @@ const EventForm: React.FC<EventFormProps> = (props) => {
...
 
+        <FormControl id="first-name" isRequired>
+          <Box display="flex" alignItems="end" m="1">
+            <FormLabel> Sponsers</FormLabel>
+            <Spacer />
+            <Button
+              onClick={() => {
+                append({
+                  type: 'FOOD',
+                  id: 0,
```
ojeytonwilliams, line 147:
```suggestion
                  type: firstSponsorType,
                  id: firstSponsorId,
```
This way we know both `type` and `id` are valid in isolation.

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -108,6 +136,72 @@ const EventForm: React.FC<EventFormProps> = (props) => {
...
+                });
+              }}
+            >
+              Add
+            </Button>
+          </Box>
+          {sponsors.map((sponsor, index) => {
+            return (
```
ojeytonwilliams, line 155:
```suggestion
            const registeredSponsor = register(
              `sponsors.${index}.type` as const,
              {
                required: true,
              },
            );
            return (
```
Used below for a custom `onChange`

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -108,6 +136,72 @@ const EventForm: React.FC<EventFormProps> = (props) => {
...
+              <Flex key={sponsor.key} borderWidth="1px" p="5" mb="5">
+                <Box display="flex" flexGrow={1}>
+                  <FormControl m="1">
+                    <FormLabel>Sponsor Type</FormLabel>
+                    <Select
+                      defaultValue={getValues(`sponsors.${index}.type`)}
+                      {...register(`sponsors.${index}.type` as const, {
+                        required: true,
+                      })}
```
ojeytonwilliams, line 164:
```suggestion
                     {...registeredSponsor}
                      onChange={(e) => {
                        registeredSponsor.onChange(e);
                        const sponsorsForThisType =
                          sponsorData?.sponsors.filter(
                            (s) => s.type === e.target.value,
                          ) ?? [];
                        setValue(
                          `sponsors.${index}.id`,
                          sponsorsForThisType[0]?.id,
                        );
                      }}
```
I noticed an issue that if you *just* changed the type and didn't click on the name, the form would not register the name correctly.

ojeytonwilliams (changes requested):
Hey Ravi, I've been doing my best to understand the mysteries of `react-hook-form` and I think I've finally got something useful to tell you.  

See below for some tentative solutions:

client/src/modules/dashboard/Events/components/EventForm.tsx
ojeytonwilliams: @Ravichandra-C I have not had chance to look into the re-fetching issue, but I'll let you know when I have something!

client/src/modules/dashboard/Events/components/EventForm.tsx
Ravichandra-C: @ojeytonwilliams , Thank you! I've modified the code to default to first sponsor type and id.

ojeytonwilliams (approved):
Thanks @Ravichandra-C, it's working really nicely now!

I finally figured out what was going wrong with the refetching: it needs to get that `EVENT` as well as the list of `EVENTS`.

Since Github wouldn't let me make that suggestion, I pushed up a quick commit.  I hope that's okay.

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -108,6 +137,88 @@ const EventForm: React.FC<EventFormProps> = (props) => {
           </FormControl>
         )}
 
+        <FormControl id="first-name" isRequired>
+          <Box display="flex" alignItems="end" m="1">
+            <FormLabel> Sponsers</FormLabel>
```
ojeytonwilliams:
```suggestion
            <FormLabel> Sponsors</FormLabel>
```

ojeytonwilliams (approved):
Re-approving since I pushed a typo fix.

ojeytonwilliams:
@Ravichandra-C let me know if you're happy with the changes and I'll get this in.  The problems we discussed re: duplicate sponsors are tracked here: https://github.com/freeCodeCamp/chapter/issues/781

Ravichandra-C:
Awesome , Thanks a lot for this @ojeytonwilliams !

PR #735, opened 2021-10-14
refactor: move interested flag to UserEventRole

ojeytonwilliams:
Instead of RSVP.interest, this makes it the responsibility of
UserEventRole

This makes RSVP responsible for information about the event (is a user
going, canceled etc.), but leaves information about the app (roles and
settings) in UserEventRole

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -2489,22 +2489,6 @@
             "deprecationReason": null
           },
           {
-            "name": "interested",
-            "description": null,
-            "args": [],
-            "type": {
-              "kind": "NON_NULL",
-              "name": null,
-              "ofType": {
-                "kind": "SCALAR",
-                "name": "Boolean",
-                "ofType": null
-              }
-            },
-            "isDeprecated": false,
-            "deprecationReason": null
-          },
-          {
             "name": "on_waitlist",
             "description": null,
             "args": [],
@@ -3834,6 +3818,22 @@
               "ofType": {
                 "kind": "SCALAR",
                 "name": "String",
+                "ofType": null
+              }
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
+            "name": "subscribed",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Boolean",
                 "ofType": null
               }
             },
--- a/client/src/generated/graphql.tsx
+++ b/client/src/generated/graphql.tsx
@@ -284,7 +284,6 @@ export type Rsvp = {
   date: Scalars['DateTime'];
   event: Event;
   id: Scalars['Int'];
-  interested: Scalars['Boolean'];
   on_waitlist: Scalars['Boolean'];
   updated_at: Scalars['DateTime'];
   user: User;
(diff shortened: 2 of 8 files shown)


allella:
The separation of roles sounds wise.  I was initially thinking we could append secondary fields to keep from overlapping table relationships, but it's true if we add one "minor" field now we'll be setting things up for more security concerns and rework.

If we're going to make a hard line in the sand on the use of the word _role_ for permissions / authorization, then how about we define it in the [terminology in the README](https://github.com/freeCodeCamp/chapter/#terminology).

I like what you suggested with a couple thoughts and questions to share.

Post-MVP, I believe an _organizer_ may want to give special _users_ attending an _event_ a title / position / (or some better synonym) to acknowledge a person for something that's unrelated to extra permissions and authorization.  I think this is in line with your comment above about 

> "The former being mostly informative, but the latter would control permissions."

For instance, if you're looking at an interface with an _event's_ list of attendees, certain people may have get an extra line of text below their name with a title like, "Presenter", "Host", "Pizza Sponsor".  Having an optional _position_title_ (or something better) text field on the newly renamed _rsvps_ table could give a good deal of flexibility without much complication or considerations of needing to define custom roles for "Beer Sponsor" vs "Food Sponsor". It would also not carry any additional concerns about defining permissions for a position / title that needs nothing more than recognition.

I'm curious if we'd treat users who are related to a _sponsors_ as another _role_ or if the aforementioned _position_title_ would suffice for recognizing them. Regardless of if sponsors can do anything special, we may need at least one additional table to account for eventual sponsorship of the _chapter_. The current issues like #96 focus on an event sponsor for the MVP, but we should be prepared to have _chapter_ sponsors at a later phase.  We have an _event_sponsors_ table, but we'd likely need a _chapter_sponsors_ at some point.

Renaming _rsvps_ to be more generic sounds reasonable. I wonder if _user_event_details_ would be more inline with the current naming pattern.  I mention this in part because if we're kicking the _user_chapter_roles_.interested_ field out of that table, then it needs a new home, like a _user_chapter_details.interested_ (or renamed as subscribed, as we've discussed)  to hold a _user's_ details as they relate to a _chapter_.

I may be over or under thinking this, but I wanted to put some things down and get feedback before diving any deeper.

ojeytonwilliams:
> If we're going to make a hard line in the sand on the use of the word role for permissions / authorization, then how about we define it in the terminology in the README.

I like that idea, it should definitely help avoid confusion.

> For instance, if you're looking at an interface with an event's list of attendees, certain people may have get an extra line of text below their name with a title like, "Presenter", "Host", "Pizza Sponsor". Having an optional position_title (or something better) text field on the newly renamed rsvps table could give a good deal of flexibility without much complication or considerations of needing to define custom roles for "Beer Sponsor" vs "Food Sponsor". It would also not carry any additional concerns about defining permissions for a position / title that needs nothing more than recognition.

Sounds good to me!  I was thinking about _attendee_status.capacity_ for this, (i.e. they're there in the _capacity_ of a Presenter but that's probably too cute.  _position_title_ seems fine for now.

> I'm curious if we'd treat users who are related to a sponsors as another role or if the aforementioned position_title would suffice for recognizing them. 

Interesting question.  If we do want to link them to a sponsor, I guess we can add a foreign key for that.  Until we're sure what we want to show, it's hard to say.

>  Regardless of if sponsors can do anything special, we may need at least one additional table to account for eventual sponsorship of the chapter.

I think so, too.  We might as well create that now, with a comment explaining why it's there unused (just to stop my future self deleting it!)

> user_event_details would be more inline with the current naming pattern

Sure, that works.

If you're happy with this direction, can you merge this?  Then I'll do the renaming and table creation in another PR.

PR #744, opened 2021-10-19
Docs: Define "role" in the README terminology table

allella:

Defining the term "_role_" to give more explicit purpose to tables like _user_chapter_role_ and _user_event_role_ and _user_instance_role_

This should help avoid mixing authorization and non-authorization fields within these tables.

For example, we want to avoid having _user_ defined details, like their _event_ or _chapter_ subscription preferences, mixed in with "_role_" tables because it could lead to a situation where CRUD grant on a table is intended to be limited to a specific field, but is then manipulated to escalate _roles_.

Closes nothing.


allella:
This is per the earlier conversation at https://github.com/freeCodeCamp/chapter/pull/735#issuecomment-944167783

PR #764, opened 2021-10-30
Docs: Add a note about the local Mailhog localhost URL to CONTRIBUTING.md

allella:

Closes nothing

Adding a note about the localhost URL where test emails can be viewed since it's not obvious what happens to emails in a dev environment.

This is important since magic links for logging in are captured by Mailhog and necessary to login to the client in dev mode.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -397,7 +397,9 @@ Once the app has started you should be able to pull up these URLs in your web br
 
 * Main Client Website - http://localhost:3000
 * GraphQL Playground - http://localhost:5000/graphql
-* Mailhog - http://localhost:8025/ - all dev environment emails are captured and viewable through this Mailhog interface 
+* MailHog - http://localhost:8025/ - all dev environment emails are captured and viewable through this Mailhog interface 
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -397,21 +397,21 @@ Once the app has started you should be able to pull up these URLs in your web br
 
 * Main Client Website - http://localhost:3000
 * GraphQL Playground - http://localhost:5000/graphql
-
+* Mailhog - http://localhost:8025/ - all dev environment emails are captured and viewable through this Mailhog interface
```
ojeytonwilliams, line 400:
```suggestion
* MailHog - http://localhost:8025/ - all dev environment emails are captured and viewable through this Mailhog interface 

Note, MailHog is not started automatically.  The easiest way to do that is via Docker: `docker run --rm --network host mailhog/mailhog`, but if you prefer to install it manually, instructions are on their [repository](https://github.com/mailhog/MailHog) 
```
It think this worth mentioning, since we don't introduce MailHog anywhere else.

ojeytonwilliams (changes requested):
Thanks, Jim.  I've added a suggestion with the easiest way to fire up MailHog:

CONTRIBUTING.md
allella: @ojeytonwilliams I don't recall having to start Mailhog in the past. Is that a new thing, or only for Docker?

PR #770, opened 2021-11-02
fix: handle expired and invalid session JWTs

ojeytonwilliams:

Instead of just reporting there's an error, this clears the JWT (since it was rejected by the server).  If the JWT has expired, the user is redirected to the login page.  This isn't the best UX, so we should investigate refresh tokens or some other way of silently authenticating.

### DIFF
--- a/client/src/pages/_app.tsx
+++ b/client/src/pages/_app.tsx
@@ -3,8 +3,12 @@ import {
   InMemoryCache,
   ApolloProvider,
   createHttpLink,
+  from,
+  ServerError,
+  ServerParseError,
 } from '@apollo/client';
 import { setContext } from '@apollo/client/link/context';
+import { onError } from '@apollo/client/link/error';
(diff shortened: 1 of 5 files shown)


moT01 (commented):
So, for the most part I think this looks good - this is what I came across when testing:

So, I would log in, then manually change the token in storage - I tried a few things, just changing it a little, removing parts/all of it - and I could still navigate around as if I'm authenticated. It wasn't until a hard refresh that the token was cleared. Is this just how it's supposed to work - or how static sites work - or maybe cause something is different locally? Seems like every request should be authenticated. Perhaps that's an issue and something for another PR - not sure.

I would be fine merging this, seems like it does what it's supposed to - curious about that behavior though.

moT01 (approved):
👍 💯 🥇 🚀

ojeytonwilliams:
Tom and I discussed this in chat.  In short: it's fine, you only see that there's a problem when you do something that requires authentication.

PR #804, opened 2021-11-16
refactor: TypeORM -> Prisma

ojeytonwilliams:
### What has changed in this PR

- TypeORM has been replaced with Prisma.  
- Some minor changes to the db schema in preparation for the rework.

#### Architectural

There are definite pros to using TypeORM alongside TypeGraphQL, since both use decorators. Also, the same files and classes can be used to define both the db tables and the GraphQL object types.  However, this makes those classes a confusing mixture of the two with no clear separation of concerns.  Fundamentally, the data source and the api specification should be allowed to vary separately.  While that is possible, doing so makes the classes even harder to understand.

Finally, it encourages developers to make everything in a given model *also* a property of the GraphQL object, making for needlessly bloated objects and types.  As a trivial example, every model inherited from https://github.com/freeCodeCamp/chapter/blob/7fa2471afae341658c29a7f5a47fefde86a52166/server/src/models/BaseModel.ts#L15-L21 and so those became part of the type, only to be removed on the client side: https://github.com/freeCodeCamp/chapter/blob/7fa2471afae341658c29a7f5a47fefde86a52166/client/src/modules/dashboard/Events/components/EventFormUtils.ts#L105

#### Type-safety

TypeORM is not truly type-safe.  It does automatically give you type information, but it can be misleading.  For instance, an Event has https://github.com/freeCodeCamp/chapter/blob/7fa2471afae341658c29a7f5a47fefde86a52166/server/src/models/Event.ts#L81-L83

if you then 

```
Event.findOne()
```
what you get is an instance of `Event`, claiming to have a reference to `rsvps`.  If you inspect that instance at runtime, you will see that `rsvps` are missing.  In other words, you get the illusion of type-safety, but have to check everything at runtime anyway.  This is a recipe for confusing bugs.

#### Bugs

When I dropped the id from the _rsvps_ table, TypeORM was unable to `save()` `Event`s any more.  There will be workarounds, but after several hours of looking, we (@moT01  and I) could not find them. In short, it's difficult to work with nested objects.  With Prisma the problem disappeared - you just give the `update` command the id and data and that's it.


ojeytonwilliams:
I think the tests are failing because the db reset takes slightly longer.

moT01 (approved):
LGTM 🎉

ojeytonwilliams:
Thanks, @moT01. Cheers for the quick review and merge.

PR #814, opened 2021-11-19
feat: authorization

ojeytonwilliams:
_Very_ preliminary draft of the authorization implementation.

Key point: all authorization logic is the responsibility of `authorizationChecker`.

This determines if a given user has a right to see or interact with a given field or resolver.  To my knowledge the main weakness of this approach is that it cannot filter a resolver's output.  So, if we wanted a public event feed and a private one, those would have to be implemented as separate resolvers.  

We cannot (well, should not) put any authorization logic in resolvers, but we can make different queries through the client based on the information about the user (are they logged in, what role do they have etc.).


ojeytonwilliams:
TODO

- [ ] Refactor the tests and fixtures (I think they're correct, but they're unreadable)

gikf:
A bit of side-question - is it planned to use this as a check if user is logged in? After passing any permission check it is implicit that user is logged in, so in theory some _logged-in_ permission could be used for authorizing access where no more specific permission exists.

ojeytonwilliams:
@gikf I hadn't planned to, but it's definitely possible if we want to.  

What you can do is simply add an `@Authorized()` decorator with no arguments.  The authorization logic would need tweaking slightly, but it's pretty straightforward.

Did you have something in mind where we'd want to do that?  If you do, I'll add that logic in.

ojeytonwilliams:
Also, I might need to drop the variable validation.  I thought it would be useful to make sure that a request comes with, say, `chapterId`, but if the request is malformed it will simply be denied.  As such, I don't think there are security implications.  However it is quite clunky, since we'd need to add every variable we use (limit, offet and so on).

gikf:
Nothing specific, just a general idea, that I'm now not sure if it's even needed...

ojeytonwilliams:
No worries. If we come up with a use case for it, it's easy enough to add in, but let's leave it until we do.

server/src/authorization/index.ts
```diff
@@ -0,0 +1,146 @@
...
+  const chapterPermissions = getChapterPermissionsById(user, chapterId);
+  return hasNecessaryPermission(roles, chapterPermissions);
+}
+
+async function isAllowedByInstanceRole(
+  user: UserWithRoles,
+  roles: string[],
+): Promise<boolean> {
+  const instancePermissions = getInstancePermissions(user);
+  return hasNecessaryPermission(roles, instancePermissions);
```
gikf, line 76:
```suggestion
  const userInstancePermissions = getUserPermissionsForInstance(user);
  return hasNecessaryPermission(allowingPermissions, userInstancePermissions);
```

server/src/authorization/index.ts
```diff
@@ -0,0 +1,146 @@
...
+  return null;
+}
+
+function isAllowedByEventRole(
+  user: UserWithRoles,
+  permissions: string[],
+  info: VariableValues,
+): boolean {
+  const eventPermissions = getEventPermissions(user, info);
+  return hasNecessaryPermission(permissions, eventPermissions);
```
gikf, line 105:
```suggestion
  const userEventPermissions = getUserPermissionsForEvent(user, info);
  return hasNecessaryPermission(allowingPermissions, userEventPermissions);
```

server/src/authorization/index.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function getInstancePermissions(user: UserWithRoles): string[] {
```
gikf, line 108:
```suggestion
function getUserPermissionsForInstance(user: UserWithRoles): string[] {
```

server/src/authorization/index.ts
```diff
@@ -0,0 +1,146 @@
...
+  return user.instance_role.instance_role_permissions.map(
+    (x) => x.instance_permission.name,
+  );
+}
+
+function getChapterPermissionsById(
```
gikf, line 114:
```suggestion
function getUserPermissionsForChapter(
```

server/src/authorization/index.ts
```diff
@@ -0,0 +1,146 @@
...
+): string[] {
+  const role = user.user_chapters.find((role) => role.chapter_id === chapterId);
+  return role
+    ? role.chapter_role.chapter_role_permissions.map(
+        (x) => x.chapter_permission.name,
+      )
+    : [];
+}
+
+function getEventPermissions(
```
gikf, line 126:
```suggestion
function getUserPermissionsForEvent(
```

gikf (commented):
Nitpicking inside, I've tried to unify names a bit and make them a bit more clear with not too much verbosity overhead. Feel free to ignore part or all of these.

ojeytonwilliams:
Thanks @gikf I like these changes a lot.  The only one I'd change is `allowingPermissions` to `requiredPermissions`.  `permissions` was definitely too vague, but `allowingPermissions` doesn't quite sound right to me.

What I'll do is commit all your changes and rename that separately.

PR #821, opened 2021-11-26
Show confirm dialog automatically if user is logged 

### LINKED ISSUE #796: Event invitations should take you directly to confirm/cancel
Right now, if I receive an invitation to an event, it has two links: "Event details" and "Cancel your RSVP".  These both link to the event page.

Instead, it would be a nicer user experience if, the first text was "RSVP to event" and immediately opened the confirm modal (rather than requiring the user to find and click on the RSVP button).  The same should be true for the cancel link.

Ravichandra-C:

Ref: #796 

1. Added useEffect hook to show the dialog if user is logged in and has not RSVP'd to the event

### DIFF
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -19,7 +19,8 @@ import {
 import { useConfirm } from 'chakra-confirm';
 import { Link } from 'chakra-next-link';
 import { NextPage } from 'next';
-import React, { useMemo } from 'react';
+import { useRouter } from 'next/router';
+import React, { useEffect, useMemo } from 'react';
 
 import { LoginRegisterModal } from '../../../components/LoginRegisterModal';
 import { useAuth } from '../../auth/store';
@@ -33,6 +34,7 @@ import { useParam } from 'hooks/useParam';
 
 export const EventPage: NextPage = () => {
   const id = useParam('eventId');
+  const router = useRouter();
   const { user } = useAuth();
 
   const [rsvpToEvent] = useRsvpToEventMutation();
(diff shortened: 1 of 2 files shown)


Ravichandra-C:
@ojeytonwilliams , For now Code only shows the confirm dialog if user is already logged into the chapter and has not rsvp'd to the event , 

I've also tried to show the confirm dialog if the user is not logged in, but the application was showing login in popup as our checkOnRsvp is only running after user is logged in. Please let me know if that is ok or if can do something else.

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -102,15 +85,41 @@ export const EventPage: NextPage = () => {
...
+    if (!rsvp) return null;
+    return rsvp.on_waitlist ? 'waitlist' : 'rsvp';
+  }, [data?.event]);
+
+  useEffect(() => {
+    if (router.query?.emaillink && user && !userRsvped) {
+      checkOnRsvp(true);
+      router.replace('/events/' + id, undefined, { shallow: true });
+    }
+  }, [userRsvped, router.query, user]);
```
ojeytonwilliams, line 106:
```suggestion
  const allDataLoaded = !loading && !!user;
  const canCheckRsvp = router.query?.emaillink && !userRsvped;

  useEffect(() => {
    if (allDataLoaded) {
      if (canCheckRsvp) {
        checkOnRsvp(true);
      }
      router.replace('/events/' + id, undefined, { shallow: true })
    }
  }, [allDataLoaded, canCheckRsvp]);
```
I think we need to go this way.  If we don't separate `checkOnRsvp` from `router.replace`, then if a user goes to http://localhost:3000/events/id?emaillink=true *after* confirming, then the query parameter will stay.

This means that clicking cancel will automatically trigger `checkOnRsvp` and ask if they want to confirm again.

ojeytonwilliams (changes requested):
> I've also tried to show the confirm dialog if the user is not logged in, but the application was showing login in popup as our onRsvp method which shows the confirm dialog is only running after user is logged in. Please let me know if that is ok or if we can do something else.

It would be nice to have it so that it goes through the login flow and then asks them to confirm.  We don't need to fix that here, though, but let's keep the original issue open to track it.  

Regardless, if you're happy with the suggested changes, I think we can merge this.

client/src/modules/events/pages/eventPage.tsx
Ravichandra-C: @ojeytonwilliams , I've updated the code to handle this. Could you please review onc?

Issue #100, opened 2019-10-20
As an owner or chapter admin I can cancel events and all participants who have RSVP'd will get an email notification.


QuincyLarson:
The UI exists and emails get sent, but we still need text for the email.

Depends on https://github.com/freeCodeCamp/chapter/pull/814

Once that is done, we should lock this down to the approved roles and confirm via Cypress tests

- [x] As an owner I can cancel any event
- [x] As an admin of a Chapter I can cancel any events in that Chapter
- [x] As any other role, I cannot cancel events 
- [x] Sensible email content
- [ ] update the test for it

The main thing to check is that requests from unauthorised users fail, but it's also helpful to confirm that the UI reflects their role. 

---

Test should account for developer typing logic, that stop canceling the event if there are no members in the event Reference https://github.com/freeCodeCamp/chapter/pull/1517#discussion_r1029353250
Test should account for the mutation deletion reference https://github.com/freeCodeCamp/chapter/pull/1517#discussion_r1029353250

allella:
Linking these for when someone starts working on email sending.

#283 #146 are related to the transactional email needed to send over SMTP.

Also, some questions which may be relevant were raised in https://github.com/freeCodeCamp/chapter/issues/320#issuecomment-581653505

allella:
This would apply to other bulk email sending as well, but the current default shows / exposes multiple recipients email addresses in the To: line in Mailhog.

We'd obviously want a BCC: for any bulk email sending.

It was suggested in https://github.com/freeCodeCamp/chapter/pull/909#issuecomment-1046295697 to 

> automatically handled by MailerService - depending if submitted email list has more than one entries. But either way that's for another PR.

Ismailtlem:
Hi,

Is this issue only about changing the body text of the email that get sent when cancelling an event?

allella:
I think @gikf has worked on a lot of the interface side, so I'm pinging and directing that question accordingly.

gikf:
Yep, from more immediate things - I believe so.

I think test for it might need changing later, once seeding will seed also unsubscribed users.

Sboonny:
Is there an email content decided? Or do I alter another email content for now, and we can think of one post MVP?

ojeytonwilliams:
@Sboonny is the test you mentioned updated?  If it is, could you close this?

Sboonny:
Nope, but I think we are done with the MVP part, we can open new issue for test.

ojeytonwilliams:
[...]

Issue #783, opened 2021-11-09
RSVP should track status with a single column


ojeytonwilliams:
Since a user can only be in a single state for a given event, the RSVP model should reflect this.  A user can either be invited, going, canceled or on the wait list and these states are mutually exclusive.  Since they can only be in one state, the model should represent this as a _status_ column, with an enum type.

In addition, to implement https://github.com/freeCodeCamp/chapter/issues/95 we will need a way of tracking how many users *rejected* an invite, so we should add an _invitee_ column that will just reflect if a user was asked to come.

ojeytonwilliams:
More discussion here: https://github.com/freeCodeCamp/chapter/issues/822#issuecomment-982813545

gikf:
Completed as a part of #965

PR #842, opened 2022-01-04
feat: send emails to rsvps on event cancel

gikf:

Still required:
- [ ] Email text.
- [x] Tests?
- [ ] Anything else?

Related to #100

---
Draft adding sending emails to rsvps when event is cancelled.
- Email is send to rsvps that are not cancelled. That means those confirmed or in waitlist. Note in generated data there might be generated rsvps with `cancelled: true`, but this doesn't seem to be displayed in the client right now.
- I've initially added permission check, similar to one in other resolver, based on `user.user_event_roles`, but as that's going to be handled by authorization I've removed it.
- Tested on local fork.

### DIFF
--- a/cypress/integration/dashboard/events.js
+++ b/cypress/integration/dashboard/events.js
@@ -219,4 +219,40 @@ describe('events dashboard', () => {
       });
     });
   });
+
+  it('emails not cancelled rsvps when event is cancelled', () => {
+    cy.visit('/dashboard/events');
+    cy.findAllByRole('row')
+      .not(':contains("canceled")')
+      .find('a')
+      .first('contains(/dashboard/events/\\d+$)')
+      .click()
+      .invoke('text')
+      .as('eventTitle');
+
+    cy.findByRole('button', { name: 'Cancel' }).click();
+    cy.findByRole('alertdialog').findByText('Confirm').click();
+
+    cy.waitUntilMail('allMail');
+    cy.get('@allMail').mhFirst().as('emails');
+
+    cy.url()
+      .then((url) => parseInt(url.match(/\d+$/)))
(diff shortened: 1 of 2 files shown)


ojeytonwilliams (commented):
Don't worry about the text for this PR, placeholders are fine for now.

Tests would be great, though!  

https://github.com/freeCodeCamp/chapter/blob/e0cac15bc75df2a3f1d8b971f4d00065690eaab8/cypress/integration/dashboard/events.js#L37

has some pretty similar logic, so might be worth a look.

One thing I was thinking about was changing RSVPs to be in a single 'state' (Going, OnWaitlist, Cancelled, etc), rather than having multiple boolean states.  Something like this: https://dbdiagram.io/d/61d5605b3205b45b73d7b04a Reason being, the RSVP can only be in a single state at once (you can't be on_waitlist and cancelled, say), so I think a single state should be easier to reason about. Since you've just been looking at rsvps, I wondered if you had any thoughts on this.

Sorry for the slight tangent and thanks for working on this!

gikf:
Thanks for the lead on tests, I'll take a look there.

I saw your issue about RSVPs status #783, I'd agree having status as a single field most likely will simplify some things and help avoid some awkwardness. Like determining if somebody is confirmed/going by checking both on_waitlist and cancelled.

ojeytonwilliams:
No problem, thanks for taking a look.  

> I saw your issue about RSVPs status

Cool, I wasn't sure if you'd have seen that, since there's been a lot of discussion. Anyways, it's not something to worry about here, but I appreciate the feedback.

ojeytonwilliams (approved):
Nice one, @gikf. This LGTM :+1:

PR #845, opened 2022-01-06
fix: prevent duplicating sponsors in ui

### LINKED ISSUE #781: Event form should not allow duplicate Sponsors
https://github.com/freeCodeCamp/chapter/pull/726 adds the UI for attaching Sponsors, but it's currently possible to add the exact same Sponsor twice. The database correctly rejects this update, but we should adjust the UI so a user never sees the option

The following should not be possible:

image

Instead, the lower sponsor name should not include any sponsors that have already been selected.  Concretely: in the above example, there are two venue sponsors, 'Batz, Williamson and Kreiger' and 'Rosenbaum Group'.  Since the first sponsor has 'Batz, Williamson and Kreiger' the second list should only include 'Rosenbaum Group'.

gikf:

Closes #781

---
- I tried to be more explicit about what is happening.
- I added `valueAsNumber: true` to registering `sponsors.${index}.id`, at some point `id`s of the sponsors in the `watchSponsorsArray` were starting to be as strings, what was messing up strict comparisons. I don't know why or how strings were making there.
- Default values for type/sponsor weren't touched (yet?), but sooner or later probably should.

### DIFF
--- a/client/src/modules/dashboard/Events/components/EventForm.tsx
+++ b/client/src/modules/dashboard/Events/components/EventForm.tsx
@@ -197,14 +197,24 @@ const EventForm: React.FC<EventFormProps> = (props) => {
                         defaultValue={getValues(`sponsors.${index}.id`)}
                         {...register(`sponsors.${index}.id` as const, {
                           required: true,
+                          valueAsNumber: true,
                         })}
                       >
                         {sponsorData.sponsors
-                          ?.filter(
-                            //Filtering out the options of only selected type
-                            (item) =>
-                              item.type === watchSponsorsArray[index]?.type,
-                          )
+                          ?.filter((sponsor) => {
+                            const isRightSponsorType =
+                              sponsor.type === watchSponsorsArray[index]?.type;
+                            if (!isRightSponsorType) {
+                              return false;
+                            }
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (approved):
Nice job, the UI is a lot friendlier now.  Thanks @gikf!

ojeytonwilliams:
> I don't know why or how strings were making there.

As I understand it, that's just the [input element spec](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) - all values are strings unless you specify otherwise.

> Default values for type/sponsor weren't touched (yet?), but sooner or later probably should.

Do you mean that it shouldn't generate empty fields if you keep clicking Add? Like this:

image

That's not *that* bad, but yeah, I'll track it in an issue.

gikf:
> Do you mean that it shouldn't generate empty fields if you keep clicking Add? Like this:

That too, but there's another case. When default type and default sponsor is already picked, but not all sponsors from default type are selected. After pressing `Add`, in new field leftover sponsor won't be selected automatically, but field will start as blank:
image

ojeytonwilliams:
Ah, yes, good point.  I'll add that to the issue.

PR #849, opened 2022-01-09
fix: sponsor fields should not be empty

### LINKED ISSUE #846: Default sponsor fields should not be empty
As an event organiser that's creating a new event, if I keep adding new sponsors they should not be empty

image
The above should never happen. 

Since each sponsor can only be added once, each new sponsor should be of a type that has un-selected sponsors.  When all sponsors have been added, the Add button should be disabled.

For example:

- If there is a single sponsor of the FOOD type, then the first sponsor added should be that and the button should immediately be disabled.

gikf:

Still required:
- [x] Clean up

Closes #846

---

- Disables `Add` button, when all existing sponsors are added.
- When adding new sponsor, default sponsor type and sponsor are selected from still available sponsors.
- Available sponsor types and sponsors are limited based on selections in other fields.
- Moves default sponsor types to utils.
- Moves functions getting allowed sponsors, sponsor types to utils.
- Any notes are appreciated.
- Tested on local fork.

### DIFF
--- a/client/src/modules/dashboard/Events/components/EventFormUtils.ts
+++ b/client/src/modules/dashboard/Events/components/EventFormUtils.ts
@@ -17,6 +17,26 @@ export type sponsorType = {
   id: number;
 };
 
+export interface EventSponsorTypeInput {
+  name: string;
+  type: string;
+}
+
+export const sponsorTypes: EventSponsorTypeInput[] = [
+  {
+    name: 'Food',
+    type: 'FOOD',
+  },
+  {
+    name: 'Venue',
+    type: 'VENUE',
+  },
+  {
+    name: 'Other',
+    type: 'OTHER',
+  },
+];
+
 export interface EventSponsorInput {
   id: number;
   type: string;
--- a/client/src/modules/dashboard/Events/components/EventFormUtils.ts
+++ b/client/src/modules/dashboard/Events/components/EventFormUtils.ts
@@ -200,7 +200,6 @@ export const getAllowedSponsors = (
       !isSponsorSelectedElsewhere(sponsor, watchSponsorsArray, sponsorFieldId),
   );
 
-
 const getSelectedFieldIdForSponsor = (
   sponsor: EventSponsorInput,
   watchSponsorsArray: EventSponsorInput[],
--- a/client/src/modules/dashboard/Events/components/EventFormUtils.ts
+++ b/client/src/modules/dashboard/Events/components/EventFormUtils.ts
@@ -1,4 +1,4 @@
-import { Event, Venue } from '../../../../generated/graphql';
+import { Event, Venue, SponsorsQuery } from '../../../../generated/graphql';
 
 export interface Field {
   key: keyof EventFormData;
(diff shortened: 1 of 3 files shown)


client/src/modules/dashboard/Events/components/EventFormUtils.ts
```diff
@@ -161,3 +161,83 @@ export const formatValue = (field: Field, store?: IEventData): any => {
...
+  sponsorFieldId: number,
+) =>
+  sponsorTypes.filter(({ type }) =>
+    hasTypeAllowedSponsors(
+      sponsorData,
+      type,
+      watchSponsorsArray,
+      sponsorFieldId,
+    ),
+  ) ?? [];
```
ojeytonwilliams, line 178:
```suggestion
  sponsorTypes.filter(
    ({ type }) =>
      getAllowedSponsorsForType(
        sponsorData,
        type,
        watchSponsorsArray,
        sponsorFieldId,
      ).length,
  );
```
It feels like overkill to create an extra function when `.length` does the same job.

client/src/modules/dashboard/Events/components/EventFormUtils.ts
```diff
@@ -161,3 +161,83 @@ export const formatValue = (field: Field, store?: IEventData): any => {
...
+
+export const getAllowedSponsorsForType = (
+  sponsorData: SponsorsQuery,
+  sponsorType: string,
+  watchSponsorsArray: EventSponsorInput[],
+  sponsorFieldId?: number,
+) =>
+  sponsorData.sponsors.filter(
+    (sponsor) =>
+      hasMatchingSponsorType(sponsor, sponsorType) &&
```
ojeytonwilliams, line 188:
```suggestion
      sponsor.type === sponsorType &&
```
I don't think we need any abstraction here. We only use `hasMatchingSponsorType` here and it's easy enough to understand.

client/src/modules/dashboard/Events/components/EventFormUtils.ts
```diff
@@ -161,3 +161,83 @@ export const formatValue = (field: Field, store?: IEventData): any => {
...
+) =>
+  watchSponsorsArray.findIndex(
+    (selectedSponsor) => selectedSponsor.id === sponsor.id,
+  );
+
+const hasMatchingSponsorType = (
+  sponsor: EventSponsorInput,
+  sponsorType: string,
+) => sponsor.type === sponsorType;
+
```
ojeytonwilliams, line 229:
```suggestion
```
Inlined above.

ojeytonwilliams (changes requested):
Nice one! The sponsor form is really smooth now.  I found a few places where I felt there was a little too much abstraction, but that was it - the code's solid.

client/src/modules/dashboard/Events/components/EventFormUtils.ts
gikf: This is lovely, I didn't realize `.length` on its own is already doing enough for `filter`.

gikf (commented):
Thanks for looking :+1:, suggestions are on-point :rocket:

client/src/modules/dashboard/Events/components/EventFormUtils.ts
ojeytonwilliams: Yeah, it works because 0 is the only falsy number. Which is pretty neat for things like this.

ojeytonwilliams (approved):
Cheers @gikf.  This LGTM :+1:

PR #858, opened 2022-01-16
feat: add css files

gikf:

---

- Adds `global.css` with header background styling.
- Moves inline css added as `style={}` to separate CSS modules.
- Removes inline css from `h1` element. And as it probably shouldn't be `h1`, changes it to `h3`.
- Note: There's also a lot of styling done by chakra with style props. I haven't touched those (except of the header color).
- Any notes and nitpicks are appreciated. 
- Tested on local fork.

### DIFF
--- a/client/src/styles/global.css
+++ b/client/src/styles/global.css
@@ -1,0 +1,3 @@
+header {
+    background: #1b1b32;
+}
--- a/client/src/components/PageLayout/Header.tsx
+++ b/client/src/components/PageLayout/Header.tsx
@@ -42,15 +42,7 @@ export const Header: React.FC = () => {
 
   return (
     <>
-      {/* TODO: Extract color into theme */}
-      <Grid
-        w="full"
-        bg="#1b1b32"
-        as="header"
-        templateColumns="repeat(3, 1fr)"
-        px="8"
-        py="4"
-      >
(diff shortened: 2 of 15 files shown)


client/src/styles/global.css
```diff
@@ -0,0 +1,3 @@
+header {
+    background: #1b1b32;
+}
```
ojeytonwilliams, line 3: Making this global creates a side-effect:
image

chakra-confirm uses `header` elements, so there's an interaction.

Is it possible to use css modules for everything, Header.tsx included?  At least until we've got a lot of repetition that might be better handling globally.

ojeytonwilliams (changes requested):
Aside from the one issue (below) I rather like this approach. For one thing, I like that Next doesn't let you import global css anywhere other than the _app file.  That seems sensible.

client/src/styles/global.css
gikf: Oops. Using css modules will work for header. Would you want to keep `global.css` empty for now, or remove it altogether?

ojeytonwilliams: Good question.  It doesn't matter a great deal, but I think we should remove it.  Reason being, it's trivial to add if needed and, until then, it's tidier without it.

ojeytonwilliams (approved):
I went through the site and everything looks how it should. Thanks @gikf :rocket:

PR #862, opened 2022-01-18
fix: bump postgres version used in Gitpod

ojeytonwilliams:


ojeytonwilliams:
Well, that went poorly.  I'll keep investigating.

ojeytonwilliams:
Closing for now since https://github.com/freeCodeCamp/chapter/pull/865 seems to have fixed the actual issue.

Even though Gitpod is using an earlier version of postgres, it doesn't seem to matter.  At least not yet.

PR #864, opened 2022-01-19
fix: check invite only checkbox when event is invite only

gikf:

---
- When event is invite only, displays checked checkbox in the edit form.
- Tested on local fork.


ojeytonwilliams (approved):
LGTM :+1: Nice catch, @gikf

PR #866, opened 2022-01-19
feat: update cache directly

gikf:

To do:
- [x] Add updating for fields referencing other objects - at least `sponsors` and `venue`.
- [ ] Add direct cache updating in other places?
- [ ] Clean up.
- [ ] Anything else?

---

- Updated values are written to two cache ids - `EventWithChapter:${eventId}` and `EventWithEverything:${eventId}` (if they exist already). First one is from events query on home page.
- Updates cache directly on event edit and event cancel.
- Any notes and nitpicks are appreciated.

### DIFF
--- a/client/src/modules/dashboard/Events/components/EventCancelButton.tsx
+++ b/client/src/modules/dashboard/Events/components/EventCancelButton.tsx
@@ -2,7 +2,7 @@ import { Button, ButtonProps } from '@chakra-ui/react';
 import { useConfirm } from 'chakra-confirm';
 import React from 'react';
 import { Event, useCancelEventMutation } from '../../../../generated/graphql';
-import { EVENT, EVENTS } from '../graphql/queries';
+import { EVENT_CANCEL_FRAGMENT } from '../graphql/queries';
 
 interface EventCancelButtonProps extends ButtonProps {
   isFullWidth?: boolean;
@@ -13,7 +13,30 @@ interface EventCancelButtonProps extends ButtonProps {
 const EventCancelButton: React.FC<EventCancelButtonProps> = (props) => {
   const { isFullWidth, buttonText, event, ...rest } = props;
 
-  const [cancel] = useCancelEventMutation();
+  const [cancel] = useCancelEventMutation({
+    update(cache, { data: canceledEvent }) {
+      cache.modify({
+        fields: {
+          events(existingData) {
+            cache.writeFragment({
+              id: `EventWithEverything:${canceledEvent?.cancelEvent.id}`,
+              data: canceledEvent?.cancelEvent,
+              fragment: EVENT_CANCEL_FRAGMENT,
+            });
+            return existingData;
+          },
+          paginatedEvents(existingData) {
+            cache.writeFragment({
+              id: `EventWithChapter:${canceledEvent?.cancelEvent.id}`,
+              data: canceledEvent?.cancelEvent,
+              fragment: EVENT_CANCEL_FRAGMENT,
+            });
+            return existingData;
+          },
+        },
+      });
+    },
+  });
 
   const confirmCancel = useConfirm({
     title: 'Are you sure you want to cancel this',
@@ -23,10 +46,6 @@ const EventCancelButton: React.FC<EventCancelButtonProps> = (props) => {
 
   const data = {
     variables: { id: event.id },
-    refetchQueries: [
-      { query: EVENT, variables: { id: event.id } },
-      { query: EVENTS },
-    ],
   };
 
   const clickCancel = async () => {
--- a/client/src/modules/dashboard/Events/graphql/queries.ts
+++ b/client/src/modules/dashboard/Events/graphql/queries.ts
@@ -129,11 +129,31 @@ export const updateEvent = gql`
       url
       streaming_url
       capacity
+      start_at
+      ends_at
+      image_url
+      invite_only
       tags {
         id
         name
       }
     }
+  }
+`;
+
+export const EVENT_FRAGMENT_UPDATE = gql`
+  fragment updateEventFragment on Event {
+    id
+    name
+    description
+    url
+    streaming_url
+    capacity
+    start_at
+    ends_at
+    capacity
+    image_url
+    invite_only
   }
 `;
 
--- a/client/src/modules/dashboard/Events/graphql/queries.ts
+++ b/client/src/modules/dashboard/Events/graphql/queries.ts
@@ -137,12 +137,34 @@ export const updateEvent = gql`
         id
         name
       }
+      sponsors {
+        sponsor {
+          name
+          website
+          logo_path
+          type
+          id
+        }
+      }
+      venue {
+        id
+        name
+        street_address
+        city
+        postal_code
+        region
+        country
+      }
+      chapter {
+        id
+        name
+      }
     }
   }
 `;
 
-export const EVENT_FRAGMENT_UPDATE = gql`
-  fragment updateEventFragment on Event {
+export const EVENT_WITH_CHAPTER_FRAGMENT_UPDATE = gql`
+  fragment updateFragmentEventWithChapter on EventWithChapter {
     id
     name
     description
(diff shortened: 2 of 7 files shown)


client/src/modules/dashboard/Events/graphql/queries.ts
```diff
@@ -129,14 +129,112 @@ export const updateEvent = gql`
...
+    image_url
+    invite_only
+    chapter {
+      id
+      name
+    }
+    tags {
+      id
+      name
+    }
```
ojeytonwilliams, line 210:
```suggestion
  fragment updateFragmentEventWithEverything on EventWithEverything {
    ...updateFragmentEventWithChapter
```
We can DRY this a fair bit.

client/src/modules/dashboard/Events/graphql/queries.ts
```diff
@@ -129,14 +129,112 @@ export const updateEvent = gql`
...
+    venue {
+      id
+      name
+      street_address
+      city
+      postal_code
+      region
+      country
+    }
+  }
```
ojeytonwilliams, line 229:
```suggestion
  }
  ${EVENT_WITH_CHAPTER_FRAGMENT_UPDATE}
```

server/src/controllers/Events/resolver.ts
```diff
@@ -373,7 +375,16 @@ ${unsubscribe}
...
-    return await prisma.events.update({ where: { id }, data: update });
+    return await prisma.events.update({
+      where: { id },
+      data: update,
+      include: {
+        sponsors: { include: { sponsor: true } },
+        venue: true,
+        chapter: true,
+        rsvps: { include: { user: true } },
+      },
```
ojeytonwilliams, line 386: Just to make sure I get what's going on: we have to include these now because we're using them to directly update the graphql cache?

client/src/modules/dashboard/Events/graphql/queries.ts
```diff
@@ -129,14 +129,112 @@ export const updateEvent = gql`
       url
       streaming_url
       capacity
+      start_at
```
ojeytonwilliams, line 132: I think the entire updateEvent mutation can be replaced by
```
  ...updateFragmentEventWithEverything
  canceled
```

ojeytonwilliams (changes requested):
Hey @gikf this looks nice.  I made some suggestions about how to DRY things out, but, while I think they're worth discussing here I'd rather they were done in a separate PR.

So, if you're happy with the general approach (where possible, composing queries out of fragments) I'll put together those changes after we merge this.

client/src/modules/dashboard/Events/components/EventCancelButton.tsx
```diff
@@ -13,7 +13,30 @@ interface EventCancelButtonProps extends ButtonProps {
...
+            });
+            return existingData;
+          },
+          paginatedEvents(existingData) {
+            cache.writeFragment({
+              id: `EventWithChapter:${canceledEvent?.cancelEvent.id}`,
+              data: canceledEvent?.cancelEvent,
+              fragment: EVENT_CANCEL_FRAGMENT,
+            });
+            return existingData;
```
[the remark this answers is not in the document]
gikf: I'm gonna write here to keep discussion complete, but example will be based on the case with editing event.

From my tests returning data is relevant in some way here.
- Load `http://localhost:3000`, press `F5` to clear cache. Two events on the main page will have `EventWithChapter:${id}` cache.
- Go to `Dashboard` -> `Events` -> one of the two first events, edit event name and submit.
- `EventWithChapter:${id}` key is not updated, what will be visible on the home page.

What I think is happening, without the returned data, is query loading events in the dashboard is run again after edit, what makes new data appear in dashboard.

Some factor might be here, that updates are via the query (?) fields - `events` and `paginatedEvents`.

server/src/controllers/Events/resolver.ts
gikf: Yes, that's how I understand it, I guess `rsvps` could be currently removed, as that's not updated after editing.

gikf:
Generally I'm all for removing unneeded repetition, and possibly making composing updates from more fragments. I was playing around a bit with it, and tried to compose updating events fragments from smaller fragments, with `sponsors`, `venue`, etc. as a separate fragments, however that was requiring increasingly more adjusts in other places, not related to this PR.

client/src/modules/dashboard/Events/components/EventCancelButton.tsx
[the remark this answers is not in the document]
ojeytonwilliams: I can't seem to reproduce the bug.  Here are the steps I took:

1. go to localhost:3000
2. press `F5`
3. navigate to dashboard/events by clicking Dashboard and then Events
4. click on the edit button of the top event (since that appears on the homepage)
5. change the name
6. click submit
7. see new name in dashboard/events
8. click the freecodecamp logo to go back to localhost:3000/
9. see new name

one thing I noticed was that the dashboard is updating regardless even if you delete the cache update logic. For some reason it's re-querying anyway, but the landing page doesn't update unless you actively modify the cache.

ojeytonwilliams:
> Generally I'm all for removing unneeded repetition, and possibly making composing updates from more fragments. I was playing around a bit with it, and tried to compose updating events fragments from smaller fragments, with sponsors, venue, etc. as a separate fragments, however that was requiring increasingly more adjusts in other places, not related to this PR.

Okay, that doesn't sound ideal.  I'll see what I can do, but if it ends up over-complicating things then I'll not bother.

client/src/modules/dashboard/Events/components/EventCancelButton.tsx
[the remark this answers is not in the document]
ojeytonwilliams: Oh, that's interesting.  I should play around with Apollo devtools more!

> Sorry, I might have described it incorrectly.

Nope, not at all. I was just confused that everything *appeared* to be working without the return, but I think I was just getting lucky.

Anyway, thanks for the detailed breakdown - your original approach definitely makes sense.

ojeytonwilliams (approved):

ojeytonwilliams:
@gikf I'm happy to merge this as is.  Is there anything else you wanted to change?

gikf:
@ojeytonwilliams I've fixed second update for editing event to make it work fully. Event cancelling updates are fine without it :man_shrugging:.

client/src/modules/dashboard/Events/components/EventCancelButton.tsx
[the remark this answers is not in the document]
gikf: I'm definitely still somewhat confused why it's working this way.

ojeytonwilliams:
Thanks @gikf. So, that last change made me think. When will I see errors if, say, I messed up the spelling of `EventWithChapter`? Or, more likely, I changed the name of the type and forgot to update `__typename`?

We want this to fail as early as possible because the bugs are likely to be subtle.

gikf:
We could add explicit interface with expected `__typename` values, that could give error, at least in editor. Or if this is subtle suggestion to add more tests around editing events, that's something to look at too.

ojeytonwilliams:
I'm not very subtle at the best of times!

My concern is that this kind of optimization bypasses type checking *and* can have bugs that don't appear on the page that causes them.  So, if we mess up, it's not obvious.

The options seem to be 

1. Add our own interface
2. Create tests for the expected behaviour (Cypress tests, probably)
3. Keep using `refetchQueries`, at least in the short term

While adding tests for this would be good, regardless, I suspect that 3. might be the way to go.  Reason being, that approach is both simple and flexible.  Simplicity helps us avoid bugs and flexibility is good because we're still building the MVP and a lot of Chapter is likely to change before we get there.

I think we've optimized prematurely and I'm sorry for both suggesting we do this (via TODOs) and only coming to this realization so late.

So, as for what to do: I think we should keep this PR around in draft form. That way, when optimization becomes a priority, we don't have to redo your work.

Does that seem reasonable?

gikf:
Sure, 3 is most convenient right now, I'm fine with whatever you decide. I got some deeper look into various internals, that's a plus regardless if this will be used right now.

ojeytonwilliams:
I'm afraid I'm only going to have time to review and merge PRs that're basically done. Unfortunately this has just got too stale.

PR #868, opened 2022-01-20
fix: use prisma's updateAt to handle updates

ojeytonwilliams:

I'm not 100% convinced we *need* `created_at` or `updated_at`, but this makes sure that `update_at` is automatically changed to the current time whenever a record is altered.

For future reference, it's not correct to use queries like `SELECT * FROM chapters WHERE created_at = updated_at;` to see if a record has been updated, since `created_at` and `updated_at` may differ by a few milliseconds even on a newly created record.  `SELECT * FROM chapters WHERE updated_at <= (created_at + '2 ms');` seems fine, though a wider interval might be wise.

### DIFF
--- a/server/prisma/schema.prisma
+++ b/server/prisma/schema.prisma
@@ -138,6 +138,8 @@ model user_event_roles {
 }
 
 model user_instance_roles {
+  created_at DateTime @default(now())
+  updated_at DateTime @updatedAt
   user_id   Int
   role_name String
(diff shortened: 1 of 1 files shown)


allella:
Looks like we have created_at and updated_at on most of the tables, right? Or is that just in [the latest diagram](https://dbdiagram.io/d/617ae3f8fa17df5ea673e3ea)?

ojeytonwilliams:
It was almost all of them.  I'm not sure why _user_instance_roles_ didn't have them, but your question made me check and it does now.

PR #872, opened 2022-01-21
refactor: create event in single call

gikf:

---

- Creates event in single call.
- `eventSponsorsData` and `userEventRoleData` objects can be moved directly into the `eventData`, if that won't get too noisy. With the `TODO` on the latter having them having them under separate variables might be a bit cleaner for now.
- To test manually, `hasPermission` check on line `261` might need to be commented out.


ojeytonwilliams (approved):
This achieves the same job with less code and fewer database calls :rocket:  

LGTM :+1:

PR #874, opened 2022-01-23
feat: return event on deletion

gikf:

---

- Fixes not deleting events due to data existing in other related tables.
- Batches deletion in transaction.
- Returns event (id) on event deletion - this will be useful in case of updating cache directly after deletion.

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -404,13 +404,7 @@ ${unsubscribe}
 
   @Mutation(() => Event)
   async deleteEvent(@Arg('id', () => Int) id: number): Promise<Event> {
-    const result = await prisma.$transaction([
-      prisma.event_sponsors.deleteMany({ where: { event_id: id } }),
-      prisma.rsvps.deleteMany({ where: { event_id: id } }),
-      prisma.user_event_roles.deleteMany({ where: { event_id: id } }),
-      prisma.events.delete({ where: { id } }),
-    ]);
-    return result[3];
+    return await prisma.events.delete({ where: { id } });
   }
 
   // TODO: This will need a real GraphQL return type (AFAIK you have to return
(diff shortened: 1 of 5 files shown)


server/src/controllers/Events/resolver.ts
```diff
@@ -402,12 +402,15 @@ ${unsubscribe}
...
-    await prisma.events.delete({ where: { id } });
-    return true;
+  @Mutation(() => Event)
+  async deleteEvent(@Arg('id', () => Int) id: number): Promise<Event> {
+    const result = await prisma.$transaction([
+      prisma.event_sponsors.deleteMany({ where: { event_id: id } }),
+      prisma.rsvps.deleteMany({ where: { event_id: id } }),
+      prisma.user_event_roles.deleteMany({ where: { event_id: id } }),
+      prisma.events.delete({ where: { id } }),
+    ]);
```
ojeytonwilliams, line 412: How about doing this via the db schema?  We can attach the `@relation(...onDelete: Cascade)` attribute to the event_sponsors, rsvps and user_event_roles models and let prisma handle it.

I think it should either all be handled in resolver logic or all handled at the DB level.  We could go either way, but data integrity seems more the database's responsibility than the server's.

gikf: Yep, that will make things easier in resolver. I've pushed changes with that.

ojeytonwilliams (approved):
LGTM :+1:

ojeytonwilliams:
I only just noticed that we have a resolver, but no way to reach that via the UI.  I think event deletion is *probably* MVP.  At the very least, I'll create an issue.

gikf:
@ojeytonwilliams events deletion is in UI in the event details on `dashboard/events/1`

ojeytonwilliams:
:man_facepalming: how did I forget that... I just looked on `dashboard/events/1/edit`, didn't see a delete button and didn't use my brain.

Anyway thanks for saving me from further foolishness.

Issue #891, opened 2022-01-29
Errors when starting docker-compose


gikf:
### Describe the bug:
When trying to run application with docker-compose I ended up today with two errors. Due to the timing I suspect it might need updating for npm workspaces, although I'm quite sure I was running it yesterday, after pulling new commits, npm workspaces included. It is possible though that docker was still using older cached images at that time.
```bash
app_1      | 
app_1      | > chapter-server@0.0.1 dev
app_1      | > ttsc && concurrently "ttsc --watch --preserveWatchOutput" "node-dev --notify=false src/app.js"
app_1      | 
app_1      | sh: ttsc: not found
chapter_app_1 exited with code 127

client_1   | 
client_1   | > chapter-client@1.0.0 dev
client_1   | > next
client_1   | 
client_1   | sh: next: not found
chapter_client_1 exited with code 127
```

After manually running `npm i` in `./client` and `./server`, client starts, while server gives another error:
```bash
app_1      | 
app_1      | > chapter-server@0.0.1 dev
app_1      | > ttsc && concurrently "ttsc --watch --preserveWatchOutput" "node-dev --notify=false src/app.js"
app_1      | 
app_1      | error TS2688: Cannot find type definition file for 'jest'.
app_1      |   The file is in the program because:
app_1      |     Entry point of type library 'jest' specified in compilerOptions
chapter_app_1 exited with code 2
```

Re-building images unfortunately doesn't help.

PR #892, opened 2022-01-29
fix: make root dir available in docker dev

ojeytonwilliams:

Closes https://github.com/freeCodeCamp/chapter/issues/891

@gikf you were right that this was a workspace issue.  Since a lot of dependencies were pulled up to `/` they weren't available to the containers (which had access to `/client` and `/server`). I'd like to keep using workspaces, but we can always revert (or switch to yarn) if the cons outweigh the pros.

Let me know if this sorts it out on your machine (it shouldn't be necessary to rebuild, just `docker-compose up`).

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -4,12 +4,12 @@ services:
   app:
     command: sh -c './wait-for db:5432 -- npm run dev'
     volumes:
-      - ./server:/usr/chapter/server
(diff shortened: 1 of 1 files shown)


gikf:
@ojeytonwilliams Thanks for fast fix, it works now :rocket:

ojeytonwilliams:
No problem, sorry I broke it!

PR #894, opened 2022-01-30
feat: throw by default on prisma findUnique and findFirst

gikf:

---

- Created as draft, as couple places relies a bit on the current behavior (returning `null` when record is not found). This might need some consideration, if this is indeed wanted change. More about this below.
- Sets Prisma to throw by default when `findUnique` or `findFrist` doesn't find record, instead of returning `null`. [(Reference)](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#rejectonnotfound). Typescript recognizes that, what makes half a dozen checks after the query obsolete.
- There's couple places where logic relies on the returned `null`. Currently if prisma would throw there when record is not found, then instead of simple check, we'd need `try/catch`. With ideally(?) making sure we are catching only not found errors. Because of that, I've added for these places optional `rejectOnNotFound` parameter to override default throwing, if needed this parameter also could be used for custom error message. Probably these could/should/would be refactored to rely more on the db constraints, but I haven't look at that too close in this PR. Couple examples: https://github.com/freeCodeCamp/chapter/blob/8287aca478355c39c0650a600cde217e9e751a53/server/src/controllers/Auth/resolver.ts#L33-L44
https://github.com/freeCodeCamp/chapter/blob/8287aca478355c39c0650a600cde217e9e751a53/server/src/controllers/Events/resolver.ts#L107-L142

### DIFF
--- a/server/src/controllers/Auth/middleware.ts
+++ b/server/src/controllers/Auth/middleware.ts
@@ -30,6 +30,7 @@ export const userMiddleware = (
     .findUnique({
       where: { id: value.id },
       include: { chapter_roles: true },
+      rejectOnNotFound: false,
     })
     .then((user) => {
       if (!user) {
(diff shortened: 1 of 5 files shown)


ojeytonwilliams (approved):
I rather like this as a default, to be honest.  If we had lots of situations where we're using `findUnique` speculatively, it would maybe be a pain, but we don't and I like the cleaner code this gives us.

PR #896, opened 2022-01-31
chore: remove ttypescript

ojeytonwilliams:
- chore: convert to relative links
- chore: remove ts-transform-paths and ttypescript

Absolute paths were quite nice, but they make it less obvious what's a node_module and what's source.  Also, supporting them requires both ttypescript and the plugin ts-transform-paths.  Finally, due to an incompatibility between ttypescript and recent versions of typescript, we've been stuck to version 4.4.4 of typescript.  Removing ttypescript unblocks updates.


ojeytonwilliams:
This should not cause any problems for development.  Just in case, though, I've added a simple 'clean' script that will force typescript to recompile everything.

PR #898, opened 2022-02-01
feat: send reminders before event

gikf:

---

- Very rough draft for something sending reminders about events to subscribed, confirmed (not cancelled and not on whitelist) rsvps.
- This draft is a bit to check whether this is appropriate direction and/or place for it. Putting it right now in `server/reminders` is mostly arbitrary.
- For testing purposes:
  - `daysForward` is set to `5` - how many days forward are events looked at.
  - Temporal `SEND_MAIL` can disable triggering sending emails.
- Is it intended for organizers to have two user event roles - `organizer` and `attendee`? 
- All notes and nitpicks are appreciated.


allella:
@gikf I saw the question in the initial post

> Is it intended for organizers to have two user event roles - organizer and attendee?

and wanted to use this as an opportunity get back into the conversation and connect the dots on previous direction.

Regarding roles @ojeytonwilliams and the related tables, [the last conversation on roles](https://github.com/freeCodeCamp/chapter/pull/735#issuecomment-943427625) is worth revisiting and I [summarized it while it was fresh in the brain here too](https://github.com/freeCodeCamp/chapter/issues/270#issuecomment-946363351).

Also, I believe the [target schema diagram](https://dbdiagram.io/d/617ae3f8fa17df5ea673e3ea), which I believe was made by Tom with fCC (not sure of username), is likely to answer many questions related to roles, users, and "details".  There may be some differences in the table names. For instance, the example schema shows _chapter_roles_, _event_roles_, and _instance_roles_ but previously had these tables prefixed them with _user__

The [current schema of the app](https://opensource.freecodecamp.org/chapter/) still shows _user_chapter_roles_, _user_event_roles_, and _user_instance_roles_ .  @ojeytonwilliams I'm having to get this all back in my brain. Did we decide to  keep the user_ prefix?

gikf:
I know current schema is expected to be changed, I'll clarify a bit why such question. When working on this draft, and querying user event roles I found organizers are getting two roles. That was a bit unexpected for me and lead me to the rsvps data generator doing just that. It's not a big deal, but if that's not intended (or it will not last) I might as well change the way it's generated when I'm poking around it here.

allella:
It's certainly the case that an organizer or administrator could managed an event, but not attend.  Or, they can manage the event and also attending the event.

So, your question is a good one that I don't think we've directly explored.  The _rsvp_ could go in the _rsvp_ table, so if an _organizer_ is attending the event we could have just one event role record for them for the event and then an _rsvp_ record to indicate their attendance. 

Having just one role may also make the permissions easier.  If the user had two roles for a single event, then you'd need some way for the permissions to have weight to know which permission "wins". @ojeytonwilliams should weigh in on this too.

Though, I think this is also an opportune time to continue [the last conversation about splitting](https://github.com/freeCodeCamp/chapter/pull/735#issuecomment-94376992) the [current _user_event_roles_ table](https://opensource.freecodecamp.org/chapter/tables/user_event_roles.html) into two tables.

_user_event_roles_ and _user_event_details_ would keep permissions related rows is isolated from user preferences related to an event, like email opt-in / out status.  This is likely best for security so we don't have regular users initiating a CRUD on a table that also houses a role.

The diagram from Tom and Oliver was made before that conversation about splitting, so it has different names (_event_users_ vs _user_event_roles_ & _user_event_details_), but the overall feeds and concepts should be nearly the same.

ojeytonwilliams:
What I'd like to avoid, as much as possible, is the conflation of a user's permissions and what they're doing at a given event.  i.e. there's no necessity for an event's keynote speaker to be able to send out invites.  

These two things should be independent.  So, I agree entirely with @allella's point here:

> It's certainly the case that an organizer or administrator could managed an event, but not attend. Or, they can manage the event and also attend.

> So, your question is a good one that I don't think we've directly explored. The rsvp could go in the [rsvp table](https://dbdiagram.io/d/617ae3f8fa17df5ea673e3ea), so if an organizer is attending the event we could have just one event role record for them for the event and then an rsvp record to indicate their attendance.

To @gikf's point about a dual roles:

> Is it intended for organizers to have two user event roles - organizer and attendee?

No, I think that's a mistake.  The permission model we're thinking about is already pretty complicated, and I don't think we'll want to allow a single user to have more than one _x_role_. i.e. you'll be able to be both a chapter admin and an event organizer, but not both a chapter admin and a chapter member.

> Did we decide to keep the user_ prefix?

It's a little redundant, but I think it's worth keeping for now.  It's a touch verbose, but it's unambiguous.

ojeytonwilliams:
As for the overall direction - I think it's looking great. However we want to schedule this, we'll need some way of finding who needs notifying and emailing them.

One thing that leaps out to me is that we'll need a way of ensuring only one reminder gets sent, so we're not spamming people.  Something as simple as a _has_been_reminded_ field, maybe?

gikf:
Yeah, I haven't thought much about how to make sure it can be run multiple times and only one reminder is send. Something like that definitely should be added. Especially since, mentioned in other places, Heroku Scheduler, doesn't actually _guarantee_ task will run _once_ (https://devcenter.heroku.com/articles/scheduler#scheduler-is-a-free-add-on).

Additional field (in event or rsvp/user-role by the subscription status?) seems to be simple, it might be good fit for a basic solution. Or maybe not necessarily much more complicated table queuing events or rvsps that still need reminders send out?

ojeytonwilliams:
How about going with a separate table? That should make it easier to work on this PR without worrying about the shape of the rest of the DB too much.

Also, I think it's a better db design practice. I'm glad you suggested it!

gikf:
> not necessarily much more complicated table

I kind of already regret saying this :laughing:. Below is one version, `events` and `users` have just added reference at the end.

- The idea is to add entry to the table when rsvp is created or accepted (for invite only events) and user is _subscribed_.
- This makes reminder logic care only about presence of entry with `notified === false` and relevant event start datetime.
- Changing _subscribed_ status could remove or add entry again, alternatively flip the `notified` field.
- When querying `event_reminders` we look only for entries with `notified === false`, within specified time range for event's `start_at`.
- Once email is send, we flip `notified`, so next time entry won't be in results of query.

```
model event_reminders {
  created_at   DateTime  @default(now())
  updated_at   DateTime  @updatedAt
  user_id      Int
  event_id     Int
  notified     Boolean   @default(false)
  event        events    @relation(fields: [event_id], references: [id], onDelete: Cascade, onUpdate: NoAction)
  user         users     @relation(fields: [user_id], references: [id], onDelete: Cascade, onUpdate: NoAction)

  @@id([user_id, event_id])
}

model events {
  created_at       DateTime               @default(now())
  updated_at       DateTime               @updatedAt
  id               Int                    @id @default(autoincrement())
  name             String
  description      String
  url              String?
  streaming_url    String?
  venue_type       events_venue_type_enum @default(Physical)
  start_at         DateTime
  ends_at          DateTime
  canceled         Boolean                @default(false)
  capacity         Int
  invite_only      Boolean                @default(false)
  image_url        String
  // TODO: none of these should be optional
  venue_id         Int?
  chapter_id       Int
  chapter          chapters               @relation(fields: [chapter_id], references: [id], onDelete: Cascade, onUpdate: NoAction)
  venue            venues?                @relation(fields: [venue_id], references: [id], onDelete: NoAction, onUpdate: NoAction)
  sponsors         event_sponsors[]
  rsvps            rsvps[]
  tags             tags[]
  user_event_roles user_event_roles[]
  event_reminders    event_reminders[]
}

model users {
  created_at          DateTime              @default(now())
  updated_at          DateTime              @updatedAt
  id                  Int                   @id @default(autoincrement())
  first_name          String
  last_name           String
  email               String                @unique
  rsvps               rsvps[]
  user_bans           user_bans[]
  chapter_roles       user_chapter_roles[]
  user_event_roles    user_event_roles[]
  user_instance_roles user_instance_roles[]
  event_reminders     event_reminders[]
}
```

server/src/controllers/Events/resolver.ts
```diff
@@ -127,16 +140,29 @@ export class EventResolver {
       if (!rsvp.on_waitlist) {
         const waitingList = event.rsvps.filter((r) => r.on_waitlist);
 
-        if (waitingList.length > 0) {
+        if (waitingList) {
```
ojeytonwilliams, line 143: I don't think we can change this, because even if we filter out all the rsvps `waitingList = []` and `[]` is truthy.

gikf: Uh, I think I confused it with just `if (waitingList.length) {`.

ojeytonwilliams:
I really like what you've come up with so far.  Great work, as always.  

What I think needs some consideration is how to make sure that we always send one and only one email.  Given that we know the Heroku scheduler can both fail to schedule jobs and schedule the same job twice, we need to handle the situation where two processes simultaneously try to send emails.

The correct approach wasn't obvious to me, but I think the following should work:

1. update the reminders to prevent any other processes from trying to send emails
2. use the updated reminders to send emails
3. delete the updated reminders
4. clean up (eventually we'll want to delete old reminders)

Where step 1. could update the _notified_ field to true.

This does assume that the process itself is reliable.  If we want to deal with the process itself terminating (vm reboots or the like) we could modify 1. to, say, find events that had been modified more than a minute ago, which would allow another process to retry emailing.

That's quite a deep rabbit hole, though, so I think it's fine to assume the process itself will be allowed to complete.  At least for the time being, we can revisit this if it turns out to be a false assumption.

Finally, if we're wanting to 'lock' the reminders (via update) then I think it makes sense to have the reminders include a _remind_at_ field rather than querying the events directly.  It also leaves room for users to configure the notification time (10 minutes before, 1 hour before etc).  It does mean we'd need to ensure that both notification time and event time changes were reflected in the event reminders. What do you reckon?

ojeytonwilliams:
> Querying via events was convenient for getting all reminders in a single step, with all necessary event info. And without need to query for event multiple times.

So, rather than a query that returns something like this:

```
[ { user_id: 1, event_id: 1, event : {...}}, { user_id: 2, event_id: 1, event : {...}}, ...]
```
you were thinking

```
[ event: { reminders: [...], ... }, ]
```
i.e. only return the event info once, rather than once per reminder?

I think that's reasonable, but it's a performance optimisation.  In my head, at least, it's more intuitive to primarily think in terms of reminders and not (yet) worry about the extra cost of sending redundant copies of the event data.

> This makes me think, we might be better off by querying and processing reminders one-by-one, instead of getting all of them at once.

Yep, that sounds good.  Batching will be faster, but that doesn't matter yet.

> - Query for one reminder not marked, with passed remind_at datetime.
> - Mark it as processing/notified.

I think this should be a single update step - that way there's no possibility that two processes both get the same reminders. 

The rest of the plan sounds solid!

gikf:
Yeah, that was the original intent with querying events, get event info once, then run all reminders for that event, and so on.

I've realized we might not need to query reminders one-by-one. During the update, in the `where` conditions, can be included also the _processing_/_notified_ field with `false`. That way, if other process already marked _this_ specific reminder, the prisma's `update` call will throw, because such not marked reminder isn't there anymore. Alternatively `updateMany` could be used, which would return 0 as a count of changed records. Either of those will work as a guard against processing the same reminder twice.

ojeytonwilliams:
> During the update, in the where conditions, can be included also the processing/notified field with false

Just to check we're on the same page:  if two processes make simultaneous update calls, whichever call is handled first will get n records and mark those n records with 'processing', so the second call will get nothing.

ojeytonwilliams:
Great, thanks for clarifying.  That seems pretty reasonable.

gikf:
Couple things adjusted to keep tests happy (happier).
- I removed `rsvp` - `user_event_role` relation. While it'd be very handy I had problems with making it work when was created just `user_event_role` for event organizer, without `rsvp`, on event creation.
- Changing `subscribed` status to not be `true` for everybody in generated user event roles - separate PR might be better for this.

ojeytonwilliams:
Hey @gikf are you done making changes?  Can I dig in and review it?

gikf:
@ojeytonwilliams at least `reminders.ts` still needs a general clean-up pass (removal of temporal toggles, etc.). But as far as how things are working I'm not planing any more specific changes. Unless requested or I forgot about something.

server/prisma/schema.prisma
```diff
@@ -63,19 +63,35 @@ model events {
...
   user_id      Int
   event_id     Int
-  date         DateTime
-  on_waitlist  Boolean
-  confirmed_at DateTime?
-  canceled     Boolean
-  events       events    @relation(fields: [event_id], references: [id], onDelete: Cascade, onUpdate: NoAction)
-  user         users     @relation(fields: [user_id], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  remind_at    DateTime
+  notified     Boolean   @default(false)
```
ojeytonwilliams, line 91:
```suggestion
  notifying     Boolean   @default(false)
```
Bit of a nitpick, but I think this is clearer.  Reason being, the reminder can be in basically three states: "initial", "notifying" and "done".  Initially `notifying` would be `false`, then when a reminder is being processed `notifying` is set to `true` and finally the reminder is done once its record is deleted.

server/src/controllers/Events/resolver.ts
```diff
@@ -11,8 +11,10 @@ import {
 } from '../../graphql-types';
 import { prisma } from '../../prisma';
 import MailerService from '../../services/MailerService';
+import { getDateTimeNMinutesBefore } from '../../util/dateUtils';
```
ojeytonwilliams, line 14: As above, we might as well use `date-fns`

server/src/controllers/Events/resolver.ts
```diff
@@ -11,8 +11,10 @@ import {
...
 import { CreateEventInputs, UpdateEventInputs } from './inputs';
 
+const DAY_IN_MINUTES = 1440;
```
ojeytonwilliams, line 17:
```suggestion
```

server/src/controllers/Events/resolver.ts
```diff
@@ -200,13 +259,36 @@ ${unsubscribe}
...
 
+    const subscribedRoles = await prisma.user_event_roles.findMany({
+      where: {
+        user_id: userId,
+        event_id: eventId,
+        subscribed: true,
+      },
+    });
+
+    if (subscribedRoles) {
```
ojeytonwilliams, line 278:
```suggestion
    if (subscribedRoles.length) {
```
Otherwise it will always add a reminder.

server/src/controllers/Events/resolver.ts
```diff
@@ -324,20 +416,37 @@ ${unsubscribe}
...
       venue_type: data.venue_type ?? event.venue_type,
-      start_at: new Date(data.start_at) ?? event.start_at,
+      start_at: start_at,
       ends_at: new Date(data.ends_at) ?? event.ends_at,
       capacity: data.capacity ?? event.capacity,
       image_url: data.image_url ?? event.image_url,
       venue: { connect: { id: data.venue_id } },
     };
 
+    if (update.start_at !== event.start_at) {
```
ojeytonwilliams, line 434: This will be true if `data.start_at` exists, right?  Even if it's exactly same date as `event.start_at`.  It would be a bit safer to use [isEqual](https://date-fns.org/v2.28.0/docs/isEqual) to see if anything has changed.

server/src/graphql-types/EventReminder.ts
```diff
@@ -0,0 +1,23 @@
...
+  remind_at: Date;
+
+  @Field(() => User)
+  user: User;
+
+  @Field(() => EventWithEverything)
+  event: Omit<EventWithEverything, 'sponsors' | 'rsvps'>;
+
+  @Field(() => Boolean)
+  notified: boolean;
```
ojeytonwilliams, line 22:
```suggestion
  notifying: boolean;
```
To keep in sync with the db.

ojeytonwilliams (changes requested):
Hey @gikf sorry for the delay getting to this.  Overall this looks great, almost all the issues I found can be summed up by "let's use date-fns more, it's a pretty neat library".

server/prisma/generator/factories/rsvps.factory.ts
```diff
@@ -2,6 +2,7 @@ import { faker } from '@faker-js/faker';
 import { Prisma } from '@prisma/client';
 
 import { prisma } from '../../../src/prisma';
+import { getDateTimeNMinutesBefore } from '../../../src/util/dateUtils';
```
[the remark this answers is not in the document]
gikf: Oh definitely. :+1:

server/src/controllers/Events/resolver.ts
gikf: I believe with unchanged start_at date in form this was being evaluated to false. For clarity and more confidence I've changed it to use `isEqual` though.

gikf:
@ojeytonwilliams no worries. I've added similar _lock_ method to processing old reminders, to not process it twice. In here it's just changed the `updated_at` datetime.

server/src/controllers/Events/resolver.ts
```diff
@@ -124,20 +139,43 @@ export class EventResolver {
...
+                user_id: acceptedRsvp.user_id,
+                event_id: acceptedRsvp.event_id,
               },
             },
             data: { on_waitlist: false },
           });
+
+          const isSubscribed = acceptedRsvp.user.user_event_roles;
+
+          if (isSubscribed.length) {
```
ojeytonwilliams, line 169:
```suggestion
          const isSubscribed = acceptedRsvp.user.user_event_roles.length;

          if (isSubscribed) {
```
Nit: it's a bit surprising when an `isX` isn't just boolean.

server/src/controllers/Events/resolver.ts
```diff
@@ -124,20 +139,43 @@ export class EventResolver {
...
+            await prisma.event_reminders.create({
+              data: {
+                user_id: acceptedRsvp.user_id,
+                event_id: acceptedRsvp.event_id,
+                notifying: false,
```
ojeytonwilliams, line 174:
```suggestion
```
Since it defaults to false.

server/src/controllers/Events/resolver.ts
```diff
@@ -157,6 +202,17 @@ export class EventResolver {
     };
 
     const rsvp = await prisma.rsvps.create({ data: rsvpData });
+    await prisma.user_event_roles.create({ data: roleData });
+    if (!event.invite_only && !waitlist && roleData.subscribed) {
+      await prisma.event_reminders.create({
+        data: {
+          user_id: ctx.user.id,
+          event_id: eventId,
+          notifying: false,
```
ojeytonwilliams, line 211:
```suggestion
```

server/src/controllers/Events/resolver.ts
```diff
@@ -200,13 +256,33 @@ ${unsubscribe}
...
+        subscribed: true,
+      },
+    });
+
+    if (subscribedRoles.length) {
+      await prisma.event_reminders.create({
+        data: {
+          user_id: userId,
+          event_id: eventId,
+          notifying: false,
```
ojeytonwilliams, line 280:
```suggestion
```

ojeytonwilliams (changes requested):
Hey @gikf I found a few little things I'd do differently.  Let me know what you think.

server/reminders/reminders.ts
```diff
@@ -0,0 +1,197 @@
...
+    data: {
+      notifying: true,
+    },
+    where: {
+      user_id: reminder.user_id,
+      event_id: reminder.event_id,
+      notifying: false,
+    },
+  });
+  return lock.count !== 0;
```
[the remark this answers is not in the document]
gikf: I like this very much.

gikf:
Not sure what's going on with the CI build fails, I don't get them locally or on gitpod.

ojeytonwilliams (approved):
Hey @gikf the linting issue showed up on GitHub (but not locally) because it runs the tests against the main branch + PR changes at the point you push a commit.  So, what happened was this: 

- Everything is working both on this PR and locally.
- Another PR, https://github.com/freeCodeCamp/chapter/pull/945, is merged, updating main
- Another commit is added to this PR
- GitHub sees that and (silently) rebases this PR branch to the new main
- Failure ensues

At least that's my understanding of what's going on. The easiest solution is to merge in main whenever there's an inconsistency.

Anyway, I did that and fixed up a few linting issues.

If you're happy that I didn't break anything, I think this can go in.

gikf:
Ah, I see. There weren't mentioned any conflicts, so I didn't suspect something like that. Thanks for explanation and fixing.

You can merge whenever ready. :rocket:

ojeytonwilliams:
Okay, I've undone the damage I caused :cold_sweat: so it's all good.

Thanks again for all your hard work and patience, @gikf!

PR #900, opened 2022-02-01
fix: don't auto add rsvps from waitlist if event is invite only

gikf:

---

- Fixes, for invite only events, bug moving rsvp from waitlist, when confirmed rsvps cancels. As that behavior seems to be fine for non-invite-only events, there's just added check if event is not invite only.

To reproduce issue:
1. Rsvp for invite only event. Rsvp is placed on waitlist.
2. Confirm rsvp in dashboard.
3. Cancel rsvp on event page.
4. See one of the remaining waitlisted rsvp be moved out of the waitlist.

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -124,7 +124,7 @@ export class EventResolver {
         },
       });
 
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (approved):
Nice catch, @gikf

PR #901, opened 2022-02-01
fix: allow returning rsvp with null confirmed_at

gikf:

---

- Changes `confirmed_at` field in `Rsvp` object type to allow being nullable. Rsvping in client expects rsvp object with `confirmed_at`. For invite only events, new rsvp object has `null` for `confirmed_at`, this results in complaints and error.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -3108,13 +3108,9 @@
             "description": null,
             "args": [],
             "type": {
-              "kind": "NON_NULL",
-              "name": null,
-              "ofType": {
-                "kind": "SCALAR",
-                "name": "DateTime",
-                "ofType": null
-              }
+              "kind": "SCALAR",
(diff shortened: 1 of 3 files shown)


ojeytonwilliams (approved):
Am I right in thinking that this doesn't have any direct consequences because we don't use the return value of `confirmRsvpFn`?  Consequences beyond Apollo complaining, I mean.

gikf:
Uhm, yes? Nothing further in client/server is affected by it right now, as far as I know, but it stops problems when trying to _Request_ rsvp for invite only event. Without the fix, I could swear there was appearing "Something went wrong" toast notification, but at the moment, I'm noticing just things getting stuck, after pressing _Request_.

ojeytonwilliams:
Cool, cool. I figured it had to be that simple, but I just wanted to confirm.

PR #916, opened 2022-02-07
fix: use npm ci in ci

ojeytonwilliams:
Rather than relying on `npm i` producing exactly the same dependency tree for each version of npm, this uses `npm ci` in all the test workflows and in the Docker images.


ojeytonwilliams:
Or not. Looks like `npm i -g npm@8` is fine, but it's ignored in CI. Presumably it's using the wrong global node_modules folder.

PR #945, opened 2022-03-01
feat: tags rework

### LINKED ISSUE #790: Tags rework
As @Zeko369 highlighted in https://github.com/freeCodeCamp/chapter/issues/605 we need to re-mplement event tags.  

Tags should be joined to events with a many to many relationship.  To do this, we need two tables: a _tags_ table and an _event_tags_ table that joins them to specific events.  All we need for the tags is to have a 'name' column and then _event_tags_ would be a junction table like _event_sponsors_.  

For reference https://dbdiagram.io/d/617ae3f8fa17df5ea673e3ea is the current database diagram we're working towards and [server/src/models/EventSponsor](https://github.com/freeCodeCamp/chapter/blob/7124ad9760a551ed52cf6f51967c9be394e93c18/server/src/models/EventSponsor.ts) is a good template for the model.  However, as shown in the database diagram, we're moving away from having an _id_ for every table, but that's a work in progress.

Once most of the database rework is done, I think this would be a good first issue,

gikf:

Closes #790

---

- Changes tags implementation.
- Updates schema, resolvers, client.
- Modifies tags generation for generated events.
- Adding/changing tags on event creation/edition takes couple steps, as it's needed to obtain `id`s for each tag and/or add them to `tags` table first. This is slightly different comparing to sponsors, where sponsor ids are already available in the form.
- Adds jest test for `EventCard` with tags displayed, updates snapshots.
- Any comments and nitpicks are appreciated.

### DIFF
--- a/client/src/modules/dashboard/Events/pages/EditEventPage.tsx
+++ b/client/src/modules/dashboard/Events/pages/EditEventPage.tsx
@@ -41,7 +41,7 @@ export const EditEventPage: NextPage = () => {
       const tagsArray = tags
         .split(',')
         .map((tag) => tag.trim())
-        .filter((tag) => tag);
+        .filter(Boolean);
 
       const eventData = {
         ...rest,
--- a/client/src/modules/dashboard/Events/pages/EditEventPage.tsx
+++ b/client/src/modules/dashboard/Events/pages/EditEventPage.tsx
@@ -36,15 +36,20 @@ export const EditEventPage: NextPage = () => {
     setLoadingUpdate(true);
 
     try {
-      const { sponsors, ...rest } = data;
+      const { sponsors, tags, ...rest } = data;
       const sponsorArray = sponsors.map((s) => parseInt(String(s.id)));
+      const tagsArray = tags
+        .split(',')
+        .map((tag) => tag.trim())
+        .filter((tag) => tag);
+
       const eventData = {
         ...rest,
         capacity: parseInt(String(data.capacity)),
         venue_id: parseInt(String(data.venue_id)),
         start_at: new Date(data.start_at).toISOString(),
         ends_at: new Date(data.ends_at).toISOString(),
-        tags: undefined,
+        tags: tagsArray,
         sponsor_ids: sponsorArray,
       };
 
--- a/client/src/modules/dashboard/Events/pages/NewEventPage.tsx
+++ b/client/src/modules/dashboard/Events/pages/NewEventPage.tsx
@@ -26,8 +26,13 @@ export const NewEventPage: NextPage = () => {
 
     try {
       const HARD_CODE = { chapter_id: 1 };
-      const { sponsors, ...rest } = data;
+      const { sponsors, tags, ...rest } = data;
       const sponsorArray = sponsors.map((s) => parseInt(String(s.id)));
+      const tagsArray = tags
+        .split(',')
+        .map((tag) => tag.trim())
+        .filter((tag) => tag);
+
       const eventData = {
         ...rest,
         capacity: parseInt(String(data.capacity)),
@@ -35,7 +40,7 @@ export const NewEventPage: NextPage = () => {
         start_at: new Date(data.start_at).toISOString(),
         ends_at: new Date(data.ends_at).toISOString(),
         ...HARD_CODE,
-        tags: undefined,
+        tags: tagsArray,
         sponsor_ids: sponsorArray,
       };
       const event = await createEvent({
--- a/client/src/modules/dashboard/Events/pages/NewEventPage.tsx
+++ b/client/src/modules/dashboard/Events/pages/NewEventPage.tsx
@@ -31,7 +31,7 @@ export const NewEventPage: NextPage = () => {
       const tagsArray = tags
         .split(',')
         .map((tag) => tag.trim())
-        .filter((tag) => tag);
+        .filter(Boolean);
 
       const eventData = {
         ...rest,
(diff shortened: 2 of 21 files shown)


server/src/controllers/Events/resolver.ts
```diff
@@ -268,6 +268,18 @@ ${unsubscribe}
...
+      .map((tagName) => tagName.trim())
+      .filter((tagName) => tagName);
+    const eventTagsData: Prisma.tagsCreateManyInput[] = tagNames.map(
+      (tagName) => ({ name: tagName }),
+    );
+
+    await prisma.tags.createMany({ data: eventTagsData, skipDuplicates: true });
+    const tags = await prisma.tags.findMany({
+      where: { name: { in: tagNames } },
+    });
```
ojeytonwilliams, line 281: It's a shame we can't get the ids as part of the query that creates them.  Oh well.

server/src/controllers/Events/resolver.ts
```diff
@@ -323,6 +339,26 @@ ${unsubscribe}
...
+    const tagNames = data.tags
+      .map((tagName) => tagName.trim())
+      .filter((tagName) => tagName);
+    const eventTagsData: Prisma.tagsCreateManyInput[] = tagNames.map(
+      (tagName) => ({ name: tagName }),
+    );
+    await prisma.tags.createMany({ data: eventTagsData, skipDuplicates: true });
+    const tags = await prisma.tags.findMany({
+      where: { name: { in: tagNames } },
+    });
```
ojeytonwilliams, line 351: Could you DRY this?  It's the same sequences of operations as for `createEvent`.

client/src/modules/dashboard/Events/pages/EditEventPage.tsx
```diff
@@ -36,15 +36,20 @@ export const EditEventPage: NextPage = () => {
     setLoadingUpdate(true);
 
     try {
-      const { sponsors, ...rest } = data;
+      const { sponsors, tags, ...rest } = data;
       const sponsorArray = sponsors.map((s) => parseInt(String(s.id)));
+      const tagsArray = tags
+        .split(',')
+        .map((tag) => tag.trim())
+        .filter((tag) => tag);
```
ojeytonwilliams, line 44:
```suggestion
        .filter(Boolean);
```

client/src/modules/dashboard/Events/pages/NewEventPage.tsx
```diff
@@ -26,16 +26,21 @@ export const NewEventPage: NextPage = () => {
 
     try {
       const HARD_CODE = { chapter_id: 1 };
-      const { sponsors, ...rest } = data;
+      const { sponsors, tags, ...rest } = data;
       const sponsorArray = sponsors.map((s) => parseInt(String(s.id)));
+      const tagsArray = tags
+        .split(',')
+        .map((tag) => tag.trim())
+        .filter((tag) => tag);
```
ojeytonwilliams, line 34:
```suggestion
        .filter(Boolean);
```

ojeytonwilliams (changes requested):
I found a few very minor things, but this works beautifully. Thanks @gikf!

server/prisma/schema.prisma
```diff
@@ -39,6 +39,17 @@ model event_sponsors {
   @@id([sponsor_id, event_id])
 }
 
+model event_tags {
+  created_at DateTime @default(now())
+  updated_at DateTime @updatedAt
+  event_id   Int
+  tag_id     Int
+  event      events   @relation(fields: [event_id], references: [id], onDelete: Cascade, onUpdate: NoAction)
+  tag        tags     @relation(fields: [tag_id], references: [id], onDelete: NoAction, onUpdate: NoAction)
```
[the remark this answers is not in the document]
gikf: You are right, I haven't even considered situation when tag is deleted on it own.

server/src/controllers/Events/resolver.ts
```diff
@@ -268,6 +268,18 @@ ${unsubscribe}
       subscribed: true, // TODO: even organizers may wish to opt out of emails
     };
 
+    const tagNames = data.tags
+      .map((tagName) => tagName.trim())
+      .filter((tagName) => tagName);
```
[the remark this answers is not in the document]
gikf: I like tricks :+1:, thanks for this one.

server/src/controllers/Events/resolver.ts
gikf: I've digged some deeper into prisma relations, we might not need ids after all.

Alternative is removing duplicates before query and using nested `create`, with `connectOrCreate` for each tag. Per [documentation](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#create-a-single-record-and-multiple-related-records) this might be less performant and internally would use multiple queries, but that doesn't seem to be concern here right now.

ojeytonwilliams (approved):
I'm more than happy to merge as is @gikf I'm just curious about one decision:

server/src/controllers/Events/resolver.ts
```diff
@@ -339,25 +340,27 @@ ${unsubscribe}
...
-        tag_id: tag.id,
-        event_id: id,
-      })),
-      skipDuplicates: true,
-    });
+    await prisma.$transaction([
+      prisma.events.update({
+        where: { id },
+        data: { tags: { deleteMany: {} } },
+      }),
```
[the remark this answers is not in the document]
gikf: I'm not aware of any specific differences, or advantages/disadvantages of one approach over the other. While I do think making both queries _via_ the same table is slight clearer, especially on the first look, I can't say there was any special deliberation put into that change :thinking:. Let me know if you'd rather have it as it was before.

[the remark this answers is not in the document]
ojeytonwilliams: Nope, it's good. I think they're pretty similar, but it is a fraction clearer like this.

ojeytonwilliams:
Thanks again @gikf it's great to be able to get this in.

PR #958, opened 2022-03-09
feat: rework instance role tables

gikf:

---

- Easiest from the three roles (as instance roles are not utilized anywhere), but I still want to take it slow.
- Adds required tables for instance roles, per https://dbdiagram.io/d/617ae3f8fa17df5ea673e3ea
- Removes old table.
- Seeds _administrator_ and _member_ instance roles.
- Users registering from UI are getting _member_ instance role.


allella:
Thanks for moving things along.

@ojeytonwilliams  What do we think about getting the tables and models stubbed out so [our schema](https://opensource.freecodecamp.org/chapter/) is fairly up to date with the proposed ER diagram?

I ask because it's sort of confusing that the [docs link to a schema](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md#schema) that's using different table names and doesn't represent a lot of the decisions that went into the proposed direction.

allella:
By "stubbing out", I mean the most basic things we can do to get the tables into code, as opposed to going deep into implementing all the logic one table or issue at a time.

Basically, build the skeleton and add the meat to the bones in subsequent PRs?

gikf:
@allella with role tables rework my plan is to kinda do exactly that. Only change tables, and whichever parts get broken in the process, and also need updates.

ojeytonwilliams:
I was just going over the db diagram https://dbdiagram.io/d/617ae3f8fa17df5ea673e3ea and it seems we're not consistent about which tables should have foreign keys.  

As a rule of thumb, how about if it makes most sense to say "Table A has a table B" then B should have the foreign key. e.g.  _chapters_ has _chapter_users_ and so _chapter_users_ should be the table with the foreign key.

In this case it makes more sense to say "a user has an instance role" than "an instance role has a user", so how about we move the fk into _instance_roles_

cc: @moT01

moT01:
I don't think that rule of thumb makes sense to me.

> it makes more sense to say "a user has an instance role" than "an instance role has a user", so how about we move the fk into instance_roles

If a user has an instance role, I think we would want to put the instance_role on the user table with the user. Here's an example using javascript that sort of illustrates what I think is the difference...

#### a:
```
const roles = {
  admin: [
    user1,
    user2,
    user3
  ],
  moderator: [
    user3,
    user4,
    user5  
  ]
}

const users = [
  {
    username: 'user1'
  },
  {
    username: 'user2'
  },
  {
    username: 'user3,
  }
  {
    username: 'user4'
  },
  {
    username: 'user5'
  },
  {
    username: 'user6'
  }
]
```

#### b:
```
const roles = {
  admin: 'admin',
  moderator: 'moderator'
};

const users = [
  {
    username: 'user1'
    role: admin
  },
  {
    username: 'user2'
    role: admin
  },
  {
    username: 'user3,
    role: admin
  }
  {
    username: 'user4'
    role: moderator
  },
  {
    username: 'user5'
    role: moderator
  },
  {
    username: 'user6,
    role: moderator
  }
]
```
Maybe it's not a good idea to try and explain database stuff with javascript, but I think that sort of demonstrates the difference. I guess either way technically works, but I would go with b. I dunno, maybe I'm wrong on it, but that's where my opinion is at the moment.

server/src/controllers/Auth/resolver.ts
```diff
@@ -39,7 +39,18 @@ export class AuthResolver {
       throw new Error('EMAIL_IN_USE');
     }
 
-    const userData: Prisma.usersCreateInput = data;
+    const instanceMember = await prisma.instance_roles.findFirst({
+      where: { name: 'member' },
+    });
```
ojeytonwilliams, line 44:
```suggestion
    const instanceMember = await prisma.instance_roles.findUnique({
      where: { name: 'member' },
    });
```
I think we can tighten this, since the schema declares these as unique.

ojeytonwilliams (changes requested):
Nice work, @gikf. I only spotted the one thing:

allella:
@ojeytonwilliams I think having an _instance_users_ table would make sense if a _user_  is going to have more than 1 _instance_ role.  However, as it's designed a _user_ can have just one _role_ for the entire _instance_ as set in _users.instance_role_id_.

Based on our terminology, the instance roles may be: _owner_, _administrator_, and _user_
I think your point was to make the instance tables work the same as the _event_users_ and and _chapter_users_ pattern, but those are obviously setup to allow
- _event_users_ - 0 or more event roles per user
-  _chapter_users_ - 0 or more chapter roles per user

allella:
I don't think we need to revisit the finer points, but I'm linking this issue to earlier conversations since it's easy to lose track of these conversations. Plus, these links tend to help those who are new to the conversation.

Earlier Conversations About Roles and Separation of Authorization vs User Details
https://github.com/freeCodeCamp/chapter/pull/898#issuecomment-1027311868
https://github.com/freeCodeCamp/chapter/pull/735#issuecomment-944167783
https://github.com/freeCodeCamp/chapter/issues/270#issuecomment-946363351

Database table dump - https://github.com/freeCodeCamp/chapter/issues/967
Terminology - https://github.com/freeCodeCamp/chapter#terminology

ojeytonwilliams:
Thanks for the context, @allella, it is hard to keep track of everything. 

The only thing I can add is that each _user_, _event_user_ and _chapter_user_ should have exactly one role each.  Since the higher roles should grant all the permissions of lower roles, there should not be a need for a single user to have more than one.

This is enforced in https://dbdiagram.io/d/617ae3f8fa17df5ea673e3ea by having the _user_ be unique by email and both _chapter_user_ and _event_user_ have unique pairs of ids.  That and the fact that all the foreign keys (_instance_role_id_, _chapter_role_id_ and _event_role_id_) cannot be null.

server/src/controllers/Auth/resolver.ts
gikf: Nice catch, this was before I added (missing) unique constraint on the `instance_roles` name.

ojeytonwilliams (approved):
LGTM 👍 

Are you happy with this, @gikf? If so, I'll merge it.

gikf:
:D That was close.

ojeytonwilliams:
I may have accidentally clicked the revert button while aiming for the laugh emoji...

Ahem.

PR #987, opened 2022-04-03
feat: add iCal file to email invitation

gikf:

---

- Adds `ical-generator` dependency, for generating iCal files/events.
- Generates iCal file/event with couple of basic fields.
- Adds generated iCal file/event to email invitation.
- I'm not entirely sure on what's the usual convention for referring to iCal calendar files with events. Some name adjusting might be in order.
- Adjusts some basic tests.
- I'll be refactoring `MailerService` constructor, to use single object as an argument, in a separate PR.

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -658,7 +658,8 @@ ${unsubscribe}`;
 
     const chapterURL = `${process.env.CLIENT_LOCATION}/chapters/${event.chapter?.id}`;
     const eventURL = `${process.env.CLIENT_LOCATION}/events/${event.id}?emaillink=true`;
-    const calendar = ical().createEvent({
+    const calendar = ical();
+    calendar.createEvent({
       start: event.start_at,
       end: event.ends_at,
       summary: event.name,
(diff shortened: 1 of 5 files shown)


server/src/controllers/Events/resolver.ts
```diff
@@ -657,7 +658,12 @@ ${unsubscribe}`;
 
     const chapterURL = `${process.env.CLIENT_LOCATION}/chapters/${event.chapter?.id}`;
     const eventURL = `${process.env.CLIENT_LOCATION}/events/${event.id}?emaillink=true`;
-    // TODO: this needs to include an ical file
+    const calendar = ical().createEvent({
+      start: event.start_at,
```
ojeytonwilliams, line 662:
```suggestion
    const calendar = ical();
    calendar.createEvent({
      start: event.start_at,
```
Turns out `createEvent` returns *just* that event, rather than the original calendar.  Unfortunately Google calendar won't import single events, only entire calendars :man_shrugging:

ojeytonwilliams (changes requested):

server/src/controllers/Events/resolver.ts
gikf: Sorry about that. I had it on separate lines initially, and when I tested it :man_facepalming:.

ojeytonwilliams: No worries, it *looks* like it should be a fluent interface, but apparently not :(

ojeytonwilliams (approved):

ojeytonwilliams:
Thanks @gikf this works great.  Any refinements (if we even need them) can definitely wait for a later date.

PR #990, opened 2022-04-06
test: venueType util

gikf:

---

- Adds tests for `client/src/util/venueType.ts`


ojeytonwilliams (approved):
Another nice easy one to review, thanks @gikf. Also, I never knew that `it.each` was a thing, so thanks for demoing it to good effect!

gikf:
> Another nice easy one to review, thanks @gikf. Also, I never knew that `it.each` was a thing, so thanks for demoing it to good effect!

I also just learned about it :grinning:

PR #996, opened 2022-04-08
test: getId util

gikf:

---

- Adds tests for `client/src/util/getId.ts`

### DIFF
--- a/client/tests/getId.test.ts
+++ b/client/tests/getId.test.ts
@@ -1,0 +1,27 @@
+import { getId } from '../src/util/getId';
+
+describe('getId', () => {
+  it('returns correct id for query including id', () => {
+    const query = { id: '10', notId: '15' };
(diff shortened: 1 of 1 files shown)


ojeytonwilliams:
Thanks for adding the tests.  They made me wonder if we should be returning 1 by default. For one thing it's mostly used like this:

```js
const id = getId(router.query) || -1;
```

 I think it makes more sense to throw an error if there's no id.  At least for now - we can come up some more sensible approaches later.  What do you reckon?

gikf:
Sure, I'll make PR with throwing and update tests once that's in.

ojeytonwilliams:
It is, yeah.  We can always fold these into that PR if it's easier.

No rush, either way.

gikf:
Closing as well, since #1396 removed `getId` util.

PR #997, opened 2022-04-08
test: date util

gikf:

---

- Adds tests for `client/src/util/date.ts`.
- Might be a bit (or more) overkill, but I wanted tests to hit at least once each month, and show how various date, time configurations will look (mainly if that's 24-hour notation and where zero-padding is used).


ojeytonwilliams (approved):
Tbh, it's hard to overdo it when it comes to tests.  As long as they're easy to understand and run quickly, it's all good.

So, with that in mind, thanks, LGTM :+1:

PR #1000, opened 2022-04-11
fix: refetch home page query on event changes

gikf:

---

- Adds refetching of the home page query when event is canceled, created, edited, deleted.
- Adds tests checking refetching for two of these cases: editing and deleting of event. To ensure cache is being used, navigation is done with on-page links, instead of visiting pages directly.
- With data generating, to test two other cases, some additional tinkering is needed to make tests consistent.


ojeytonwilliams (approved):
Nice work, @gikf 

LGTM :+1:

PR #1001, opened 2022-04-11
fix: throw when no id in query

gikf:
Co-authored-by: Oliver Eyton-Williams <ojeytonwilliams@gmail.com>

---

- `getId` throws when there's no id in query.
- Removes obsolete ` || -1` from where it's used.

### DIFF
--- a/client/src/hooks/useFetchIfId.ts
+++ b/client/src/hooks/useFetchIfId.ts
@@ -1,0 +1,20 @@
+import { useEffect } from 'react';
+
+interface Options {
+  variables: {
+    id: number;
+  };
+}
+
+export function useFetchIfId(
+  id: number | null,
+  fetchData: (options: Options) => void,
+): void {
+  useEffect(() => {
+    if (id !== null) {
+      fetchData({
+        variables: { id },
+      });
+    }
+  }, [id]);
+}
--- a/client/src/modules/dashboard/Chapters/Users/pages/ChapterUsersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/Users/pages/ChapterUsersPage.tsx
@@ -11,7 +11,7 @@ import { getId } from '../../../../../util/getId';
 export const ChapterUsersPage: NextPage = () => {
   const router = useRouter();
 
-  const id = getId(router.query) || -1;
+  const id = getId(router.query);
 
   const { loading, error, data } = useChapterUsersQuery({
     variables: { id },
--- a/client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
@@ -12,7 +12,7 @@ import { Layout } from '../../shared/components/Layout';
 
 export const ChapterPage: NextPage = () => {
   const router = useRouter();
-  const id = getId(router.query) || -1;
+  const id = getId(router.query);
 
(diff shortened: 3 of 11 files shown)


ojeytonwilliams:
Oops, I wasn't thinking about building the static site.  I've looked around a bit and it it seems like `router.isReady` can be used to work around this: https://github.com/vercel/next.js/discussions/11484

Something as simple as passing `router` into `getId` and returning `null` if it's not ready or it can't find an id should be fine.  Then we can render null until the page is ready.

Does that sound reasonable?

gikf:
I'm having troubles with using `router.isReady` without errors or warnings.

When using it in `getId` function:
```ts
export const getId = (router: NextRouter): number | null => {
  if (!router.isReady) {
    return null;
  }
/* ... */
}
```

```ts
  const id = getId(router);

  if (!id) {
    return null;
  }
```
This gives out warnings like:
```
Warning: React has detected a change in the order of Hooks called by ChapterUsersPage. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://reactjs.org/link/rules-of-hooks

   Previous render            Next render
   ------------------------------------------------------
1. useContext                 useContext
2. undefined                  useContext
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

---

On the other hand https://nextjs.org/docs/api-reference/next/router suggests to use it within `useEffect`, with that and removing check for `id` (which was causing for this case different errors):
```ts
  const [id, setId] = useState(Number);
  useEffect(() => {
    if (!router.isReady) {
      return;
    }
    setId(getId(router));
  }, [router.isReady]);
```
This approach often (although not every time) is throwing error as if the query is executed with not existing `id`.

Any ideas?

ojeytonwilliams:
Hey @gikf, so I played around with this a little and this is what I came up with.  The idea is that `getId` returns `null` if it can't find the id. I don't believe we need to use `isReady`, because we only really care if the id is defined (as far as I can see).

Since the id is always `null` initially, I did end up using lazy queries. Without them, there's always going to be a useless request.

Anyways, before I go too deep down this rabbit hole I could use some feedback.  Does this whole thing seem reasonable?

gikf:
Keep in mind I'm still weirded out by most of the react/next things.

For what this does, it adds some, but not that much complexity. And what are known alternatives? Keeping things like they are right now and/or just changing `getId` to return `-1` as default.

ojeytonwilliams:
> Keep in mind I'm still weirded out by most of the react/next things.

No worries, I'm still not *that* used to next and I can definitely get better at react!

> And what are known alternatives? Keeping things like they are right now and/or just changing getId to return -1 as default.

Three possibilities I'm aware of:

1. The approach you mentioned: fetch an invalid id (i.e. -1) on first render 
2. Fetch nothing on first render and only fetch once a valid id has been found
3. Use another approach to get the id on the first render (e.g. parse `window.location` to extract the id)

To be honest, I'm 100% not sure which is best, but 2 seems to be the way to go.  Reason being, 1. means pointless requests and 3. is a pain when you try to generate static pages.  You can't access `window.location` at that point, so you have to conditionally get the id... at which point you're back to 1. or 2. but more convoluted.

ojeytonwilliams:
Since https://github.com/freeCodeCamp/chapter/pull/1396 has gone in, this seems redundant. We can try to optimize later, if we need to!

Issue #274, opened 2019-11-30
UI: As a chapter member, I can unsubscribe from emails.


QuincyLarson:
Depends on #106

allella:
@ceciliaconsta3 so we agree on a label.  Is the checkbox suggestion vs switch for accessibility?  What's the experience when a screen reader encounters a label and a checkbox vs a label and a swtich?

> @allella A label is definitely the surest bet. If, and it sounds like, we will use the Subscribe component in a form alongside other elements to complete an event sign-up, (if MVP won't have a User Account Settings page) we should treat the Subscribe switch as a checkbox using checkbox role instead. It'll get sent at form submission with the other fields.
> 
> I think "Active" color as Subscribe/unchecked status and grey/disabled for Unsubscribe/checked status would do.
Copied this from #107 since this is a UI conversation.
https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-562417958

ceciliaconsta3:
That's a good question; just used Narrator to review the [site you referenced](https://material-ui.com/components/switches/#switches-with-formcontrollabel) and because the switches are essentially set up as checkboxes, I had no problem receiving the status information (ya know, aside from normal screenreader difficulties). I'll look for a less optimized site using switches to compare later.
 
 **Functionality:**  Technically, based on the usage defined, if we are waiting until the RSVP form is sent (subscription depending on form submission) we require a checkbox (form input) or button (performs some action) here as a field in the RSVP form. If the subscription is being set independently of the RSVP, say on toggle, switch it is. [This article has a cool comparison](https://uxplanet.org/checkbox-vs-toggle-switch-7fc6e83f10b8) 
 Checkboxes/buttons both are natively supported in modern screenreaders and make semantic sense like labels.  _For reference/competitive analysis, I see Meetup uses buttons/checkboxes on desktop and switches on their app for the same task (probably due to screen real estate)._
 
 **Visually:** We can totally create switch appearing checkboxes. Although, there can be some ambiguity when it comes to switch visual cues (which side is on or off? Is it the side I want?).
 
 **Current UI:** Reviewing our Figma mockups, we don't have a view for a RSVP confirmation screen yet so I'm going by the current minimal/high contrast design.

allella:
#107 mentions the likelihood of needing to subscribe to `chapter` notifications separate from `rsvp` notifications, so we can let the context guide the UI elements.

[...]

Issue #275, opened 2019-11-30
UI: As a future participant after the event, I will automatically get emails notifying me of subsequent events.


QuincyLarson:
Depends on #107

allella:
The text on this [comment on #276](https://github.com/freeCodeCamp/chapter/issues/276#issuecomment-596913322) can likely be tweaked to cover upcoming _event_ notifications.

As an rough example,

```
Upcoming Events for {chapters.name}

When: {events.start_at} to {events.end_at}
Where: {venues.name}

All Upcoming Event Details: [URL to the event page, like https://{instance domain name}/chapters/:chapterId]

- Stop receiving upcoming event notifications for {chapters.name}: [URL to unsubscribe, like https://{instance domain name}/upcoming/unsubscribe/:user_id/:chapter_id/{a unique hash value to prevent 3rd party tampering}]
- More about {chapters.name} or to unfollow this chapter: [URL to chapter page, like https://{instance domain name}/chapters/:chapterId]

----------------------------
You are receiving this email because you previously followed {chapters.name}.

See the options above to change your notifications.
```

allella:
The assumption is this would be for a single upcoming event, but it could be modified to show multiple upcoming events, or a subset of upcoming events and then link to the main events page for the specific _chapter_.

allella:
I wanted to add the notes below since his email will likely include some options to allow _members_ to unfollow or unsubscribe a type of notification or the entire _chapter_.

The previous conversations about following / unfollowing a Chapter and unsubscribing from emails should provide good context on the type of links and privacy settings we imagined to make for easy opt-in and opt-out.

That conversation also goes into interfaces on the website and name of buttons, type of buttons and such.
For the emails, we also touched slightly on the security aspects of allowing _members_ to unsubscribe or opt-out without needing to login. For that, something like a secure / unique hash was discussed. Whether that anonymous / hash ideas is the correct approach, or easy to do is up for debate, but I thought to mention it.

The core of those email and UI opt-in / out conversations were on #107 and I tried to summarize the conversation at a few points, like:

https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-596905797
https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-562374895
https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-562432174

bappyasif:
@allella hello again, how can i help with this issue as well? looking forward to hear from you, thanks.

allella:
[...]

PR #1004, opened 2022-04-13
feat: unsubscribe link

gikf:

To do
- [ ] Tune wording, and unify generating unsubscribe links.
- [x] Figure out sending batch emails with unsubscribe links included.
- [ ] Be more explicit about what is being unsubscribed?
- [x] Tests.
- [ ] More tests?
- [ ] Clean up.

---

- Related to #274
- Sets pieces for implementing unsubscribing via link, sent in email.
- ~Very drafty, some parts are just put in the appropriate places, some are waiting for new roles (actual mutation).~
- Still messy, but basic functionality is in.
- Uses general `/unsubscribe?token=${token}` route. With signed token including user id, whether it's chapter or event subscription and id of the element. As that information is in token, I'm not sure if it's worth to use more detailed links.
- Unsubscribing doesn't require being logged in. That means anybody with link can unsubscribe the owner of link. Without explicitly forcing user to log in, some more secure option could be, when visiting link with token, sending them another email, as a two-step process.
- Adds `batchSender` allowing sending multiple emails, with changed subject/text per recipient, it expects generator function yielding at least email address, subject and text.
- As long as token is valid, unsubscribing is idempotent.


allella:
@gikf I wanted to be sure you've seen the earlier conversations about unsubscribe links, the difference between unfollowing / unsubscribing from a Chapter vs an event.  

Also, there was some discussion on using unsigned JWOT tokens. I think @ojeytonwilliams  made that suggestion.

https://github.com/freeCodeCamp/chapter/issues/276#issuecomment-1000442379
https://github.com/freeCodeCamp/chapter/issues/275#issuecomment-868748463

allella:
There are also some draft examples of how the unsubscribe / unfollow link might appear in the email body or footer.
https://github.com/freeCodeCamp/chapter/issues/276#issuecomment-1000442379

gikf:
This is still messy and bit raw, but things are working. Some more general comments would be appreciated. The easiest way to get links is rsvping to the event and checking email with confirmation.

allella:
@gikf it seems this is code, but let me know if there's any less technical questions where I can help.

Thanks

gikf:
@allella Does trying out functionality matches that? :slightly_smiling_face: Some insights on how currently the whole process works also would be useful.

ojeytonwilliams:
@gikf Cypress worked fine locally, so I restarted CI. :crossed_fingers:

gikf:
I was able to get it locally as well. Once every few runs.

ojeytonwilliams:
Okay, I'll see if I can reproduce it.

ojeytonwilliams:
@gikf  I think your diagnosis was correct - sometimes `waitUntilMail` would find some mail before the new set of emails had been sent and use that as `allMail`.

By clearing before checking that should never happen.

gikf:
Huh, that solution is a bit weird :smiley:, thanks for it nevertheless.

server/src/services/MailerService.ts
```diff
@@ -92,3 +92,23 @@ export default class MailerService {
...
+export async function batchSender(
+  mailData: () => Generator<BatchEmailData, void>,
+) {
+  for (const { email, subject, text, options } of mailData()) {
+    await new MailerService({
+      emailList: [email],
+      subject,
+      htmlEmail: text,
+      ...options,
+    }).sendEmail();
```
ojeytonwilliams, line 112: This waits until each email has been sent before starting to send the next one.  Is that the plan?  It's possible to wait for the whole batch, if that's desired.

gikf: _Plan_ is a strong word, not really a plan. I didn't think too much about it, I might have assumed it's possible it will be used with potentially high number. Which approach do you think would be better?

ojeytonwilliams: So, generally what I do if there's a bunch of async tasks I want to perform and then wait for them all to be done is the following.  Start by creating an array (of email data, in this case), map that to an array of Promises and then `await Promise.all(theArrayOfPromises)`.

That way all the async calls get fired as soon as possible and you still control the size of the batch.

Does that make sense? If not, I can show you some code.

gikf: I feel like we already had similar conversation somewhere :stuck_out_tongue:. I've added change, see if that works for you.

ojeytonwilliams: It does seem eerily familiar!

Yeah, this seems good, thanks :)

server/src/services/UnsubscribeToken.ts
```diff
@@ -0,0 +1,32 @@
...
+  Chapter,
+  Event,
+}
+
+export type UnsubscribeToken = {
+  type: unsubscribeType;
+  id: number;
+  userId: number;
+  iat: number;
+  exp: number;
```
ojeytonwilliams, line 15:
```suggestion
```
See below

server/src/services/UnsubscribeToken.ts
```diff
@@ -0,0 +1,32 @@
...
+  return sign(
+    {
+      type: type,
+      id: id,
+      userId: userId,
+    },
+    getConfig('JWT_SECRET'),
+    { expiresIn: '14d' },
+  );
+};
```
ojeytonwilliams, line 32:
```suggestion
) => sign({ type, id, userId }, getConfig('EMAIL_SECRET'));
```
For this, it's very important that we use a different secret than for the bearer tokens.  Reason being, a user could easily share an unsubscribe token by accident (they email an event invite to a friend, say). If the token is signed with this secret, then that user just shared a bearer token and their friend can use that to login as them.  Kinda dangerous.

Given that I almost didn't notice, I'm going to rename `JWT_SECRET` to something less generic to save my future self.

Aside from that, once we're confident that the token can *only* be used to unsub, then I think it should never expire.  Reason being: we have to (legally, I believe) make it easy to unsubscribe and we don't want our emails reporting as spam.  Most people will click on a recent email, sure, but you never know.

Finally, this is a little neater!

server/src/controllers/Unsubscribe/resolver.ts
```diff
@@ -0,0 +1,73 @@
...
+};
+
+const typeToUnsubscribe = {
+  [unsubscribeType.Chapter]: unsubscribeFromChapter,
+  [unsubscribeType.Event]: unsubscribeFromEvent,
+};
+
+@Resolver()
+export class UnsubscribeResolver {
+  @Mutation(() => Boolean)
```
ojeytonwilliams, line 60: I vaguely recall there being some issue with Boolean return types... Don't we have to return `{ id }` or something?

ojeytonwilliams (changes requested):

cypress/e2e/unsubscribe.cy.js
```diff
@@ -0,0 +1,75 @@
...
+  beforeEach(() => {
+    cy.exec('npm run db:seed');
+    cy.mhDeleteAll();
+  });
+
+  it('should allow for unsubscribing without being logged in', () => {
+    const chapterId = 1;
+    const emailAddress = 'foo@bar.com';
+    cy.login();
+    cy.mhDeleteAll();
```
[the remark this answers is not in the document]
gikf: Oh right, `cy.login()` doesn't use email.

server/src/controllers/Unsubscribe/resolver.ts
gikf: I'm not remembering anything troublesome with it, there's few other resolvers returning `Boolean`.

server/src/services/UnsubscribeToken.ts
```diff
@@ -0,0 +1,32 @@
+import { sign } from 'jsonwebtoken';
+
+import { getConfig } from '../config';
+
+export enum unsubscribeType {
+  Chapter,
+  Event,
+}
```
gikf, line 8: Now it crossed my mind whether shouldn't there be added values to these enums, to ensure they stay consistent in future (ie. in case of adding instance unsubscribing, or just in general).

server/src/services/UnsubscribeToken.ts
gikf: This sounds good as well!

allella:
@ojeytonwilliams I'm assuming the new unsubscribe link doesn't require logging in, which still would allow someone else to unsubscribe if the person happened to forward the event email.

Even if for post-MVP, do we want to send a confirmation email "You've been unsubscribed. We this a mistake, if so you can undo this action?"

and/or would we want to put a warning note on the top or bottom of emails that use any sort of token to warn people of the risks of sharing?

server/src/controllers/Unsubscribe/resolver.ts
ojeytonwilliams: Okay, cool. Since it's working now, that's fine.

ojeytonwilliams:
> Even if for post-MVP, do we want to send a confirmation email "You've been unsubscribed. We this a mistake, if so you can undo this action?"

I'm not sure we do, honestly.  If someone doesn't want email, then sending them a confirmation email is probably not the way to go.  We could show them a resubscribe page with that same info, though.

> and/or would we want to put a warning note on the top or bottom of emails that use any sort of token to warn people of the risks of sharing?

The only token we should be sending via email is this one due to the risk that someone accidentally shares it.  

As for whether or not we should warn people, I'm not sure.  If we want to mitigate the downsides of sharing, I think it makes sense to have each token only work once.  If we do that, then all you have to do is resubscribe once and stop sharing emails to the wrong people.

For instance, we could have an unsubscribeId in the db that gets regenerated each time the user subscribes.  That doesn't feel MVP worthy, though.

ojeytonwilliams:
@gikf since I picked a slightly too generic name for the secret, I went ahead and changed it.

ojeytonwilliams (approved):
LGTM :+1: 

Another MVP issue down :rocket:

Awesome work @gikf!

PR #1007, opened 2022-04-14
fix: don't compile server tests

gikf:

---

- Removes folder with server tests from compiling.
- Prevents duplicated tests runs - once from `.ts` and once from `.js` files.
- Client doesn't have this issue.


ojeytonwilliams (approved):
Good catch !

PR #1013, opened 2022-04-16
test: calendar file in invitation email

gikf:

---

- Adds test if invitation email contains calendar file.
- With briefly checked contents.

### DIFF
--- a/cypress/integration/dashboard/events.js
+++ b/cypress/integration/dashboard/events.js
@@ -161,6 +161,38 @@ describe('events dashboard', () => {
     cy.mhDeleteAll();
   }
 
+  it('invitation email should include calendar file', () => {
+    cy.visit('/dashboard/events/1');
+
+    cy.findByRole('button', { name: 'Email Attendees' }).click();
+
+    // try to make sure there will be recipient
+    cy.findByLabelText('Waitlist').click({ force: true });
+    cy.findByLabelText('Cancelled').click({ force: true });
+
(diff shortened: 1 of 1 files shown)


cypress/integration/dashboard/events.js
```diff
@@ -161,6 +161,38 @@ describe('events dashboard', () => {
...
   }
 
+  it('invitation email should include calendar file', () => {
+    cy.visit('/dashboard/events/1');
+
+    cy.findByRole('button', { name: 'Email Attendees' }).click();
+
+    // try to make sure there will be recipient
+    cy.findByLabelText('Waitlist').click({ force: true });
+    cy.findByLabelText('Cancelled').click({ force: true });
```
ojeytonwilliams, line 171: Do we need to force, here? It's a bit of an indication that something's not quite right.

ojeytonwilliams (commented):
This looks nice.  It's not a big deal, but I was curious about the forcing:

cypress/integration/dashboard/events.js
gikf: This is the matter in few other tests in `events.js` as well, for some reason cypress detects those elements as hidden by another element, and without forcing, it doesn't work. I didn't dig too deep to find what's up with that.

ojeytonwilliams: Understood. I'll add that to my backlog, then.

ojeytonwilliams (approved):

PR #1025, opened 2022-04-21
fix: click parent of input

gikf:

---

- Stops requiring to force click in cypress. 
- This is achieved by clicking the label, instead of the input. This is one of the possible ways, `cy.findByLabelText` selects the input, `.parent()` goes back to the label itself.


ojeytonwilliams (approved):
Thanks for figuring this out. It slightly bugs me that we have to navigate the DOM for something as trivial as this, but this is clearly the way to go.

Issue #1024, opened 2022-04-21
Client might not receive data after resetting db


gikf:
### Describe the bug:
I mentioned this issue long time ago in chat, but just today I've found what's relevant for the reproducing it (I think).

1. Register as new user
1. Login as new user
1. Run `npm run db:reset`
1. `F5` in browser
1. Client doesn't get any data from server, spinners will spin infinitely
1. To make things work again - remove site data in browser

- Issue doesn't occur when logged in as seeded user (`foo@bar.com`).

### Tell us about your browser and operating system:
- Browser(s) (checked on): Firefox 99, Brave 1.37.116
- Operating System: ubuntu 20.04

allella:
By chance, is this related to the [_JsonWebTokenError: invalid signature_ error](https://github.com/freeCodeCamp/chapter/issues/471)?

gikf:
I haven't seen any explicit error, but it's likely that token is involved here.

PR #1027, opened 2022-04-22
fix: handle missing users

ojeytonwilliams:

Closes https://github.com/freeCodeCamp/chapter/issues/1024

The issue was that the client was sitting patiently waiting for a response that never came.  We might want to expand this and do more with the returned messages, but I think this is okay for now.


gikf (approved):
Thanks for this, it works well here :+1:

PR #1035, opened 2022-05-10
fix: venue type affecting venue change test

gikf:

---

- Fixes inconsistent test, failing when venue type of the event didn't show venue dropdown.
- Creates event at the start of test, which also simplifies ensuring venue will be changed later in test.
- Adds two custom cypress commands. For creating and deleting event via graphql query. As doing that with UI is way slower, this can be a start towards replacing seeding before each test with quicker and smaller commands.
- `cy.createEvent`
  - Request with query uses manually saved token, otherwise authorization in resolver wasn't successful.
  - In function declaration I wasn't able to import `CreateEventInputs` interface for the accepted argument. For some reason it was causing troubles with `Chainable` type.


gikf:
Note: this should start working once merged with event role changes. Sorry about that.

cypress/support/commands.js
```diff
@@ -177,3 +177,47 @@ Cypress.Commands.add('waitUntilMail', (alias) => {
...
+    variables: {
+      data: { ...data },
+    },
+    query: `mutation createEvent($data: CreateEventInputs!) {
+      createEvent(data: $data) {
+        id
+      }
+    }`,
+  };
+  cy.login();
```
ojeytonwilliams, line 193:
```suggestion
```
Since, in part, we're trying to improve performance, we shouldn't login here.  It's a little less DRY, but means that the command is clearer (`createEvent()` should just create an event).

ojeytonwilliams (changes requested):
This is way less silly than my UI manipulation approach, thanks!

Just one minor thing:

cypress/support/commands.js
gikf: I'm not a fan of this either. Problem is request needs to have token added, without calling login here, it needs to be ensured token is there. Maybe then pass the authorization token when calling function? `createEvent(data, token)`
```

Comment (ojeytonwilliams) [reply cypress/support/commands.js:193]: Right, we definitely need to be logged in when calling this. That said, I think we can separate the calls and do this:

```js
cy.login();
...
cy.createEvent(eventData);
```

In this spec file we're already logging in the `beforeEach` hook, so we're good.

Also, if we _do_ forget, we get a nice error in the console: `Error: User must be logged in to create events`

ojeytonwilliams (approved):

PR #1046, opened 2022-05-13
fix: ensure seeded events are upcoming

gikf:

---

- `cypress/integration/events/index.js` intermittent fails (ie. https://github.com/freeCodeCamp/chapter/runs/6427610818?check_suite_focus=true), test seems to be caused by `/events` displaying only upcoming events. In the rare occurrence of event being seeded as already passed, the number of displayed events will be less.
- Adds one day to the random date of event, ensuring all seeded events are upcoming.
- Some alternative might be changing test seeding from `before` to `beforeEach`, but it might be not ideal to still have seeds for which test can't success.

### DIFF
--- a/server/prisma/generator/factories/events.factory.ts
+++ b/server/prisma/generator/factories/events.factory.ts
@@ -38,7 +38,7 @@ const createEvents = async (
     date.setMinutes(0);
 
     const start_at = add(date, {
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (approved):
Nice catch 👍

PR #1047, opened 2022-05-14
fix: wait for ui before login try

gikf:

---

- Slight tweak to test going a bit spotty recently - https://github.com/freeCodeCamp/chapter/runs/6429048621?check_suite_focus=true.
- As right now modal changes automatically on successful registration, explicit clicking to swap back to login form is not needed.


ojeytonwilliams (approved):
I don't remember changing that, but no matter, this is the behaviour we want!

PR #1049, opened 2022-05-15
feat: use bcc field to send email

### LINKED ISSUE #956: Bulk emails sent by Chapter should be BCC
We can consider the cases where we'd want to use the 'to' and 'cc' fields.  However, for the MVP, we should stick to BCC for all emails to respect people's privacy.  

Brought up by @allella here: https://github.com/freeCodeCamp/chapter/pull/909#issuecomment-1046392703

Nirajn2311:

Closes #956

This PR sends all emails with the recipients in the `BCC` field instead of the `To` field. The tests have also been updated to check for it and I've created a custom command for it. 

I was unsure if **every** or only bulk emails should use `BCC` so I went with every mail for now. If it's only for bulk emails then I think the `sendEmail` function will need an extra argument for the case where bulk emails are sent to the event organizers(assuming these recipients are not to be BCCed).

### DIFF
--- a/cypress/integration/dashboard/events.js
+++ b/cypress/integration/dashboard/events.js
@@ -63,6 +63,9 @@ describe('events dashboard', () => {
         .mhGetRecipients()
         .should('have.members', subscriberEmails);
     });
+    cy.get('@invitation').then((mail) => {
+      cy.checkBcc(mail).should('eq', true);
+    });
   });
 
   function createEvent(chapterId) {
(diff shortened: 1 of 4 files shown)


ojeytonwilliams (approved):
Nice work, @Nirajn2311 this is a smart way to check that the emails are bcc.

allella:
@all-contributors add @Nirajn2311 for code

ojeytonwilliams:
Good catch @c-ehrlich, I'll take care of it.

PR #1051, opened 2022-05-18
feat: events adjust to local times and time updates are saved correctly

### LINKED ISSUE #959: Use user's timezone when displaying/sending dates
Currently event times are handled a little inconsistently.  

- When you enter a time for an event via the ui (on http://localhost:3000/dashboard/events/[id]/edit) that time is implicitly local time.  
- This time is stored without a timezone (i.e. in UTC)
- The time is formatted in the user's local time on http://localhost:3000/dashboard/events
- The time is formatted *incorrectly* if you subsequently visit http://localhost:3000/dashboard/events/[id]/edit (because it's set to UTC first)
- Notification emails are sent formatted in the server's local time
- Reminder emails are (well, will be https://github.com/freeCodeCamp/chapter/pull/898) explicitly set to GMT

## What to do?

Let users specify a timezone and use that.  Since browsers can detect timezones, we can default to that when creating the user account.

Nice to have: warn users if their browser detected timezone doesn't match their account timezone

pdotsani:

Closes #959 

Times were displaying inconsistently throughout the app due to a bug in how the datepicker was saving dates to the database. While we thought we saved times locally via the UI, it was actually sending the time back to the server in another format, messing up the offset and throwing all our times off. This pr uses react-datepicker, which handles the time object in the datepicker instead of a selected parsed string. This ensures we get the correct UTC time and offset, which when tested, should reflect accurate to other local times.

To test, run the app and try updating / creating events and check to see if the dates and times match up. Additionally, switch timezones by going to the developers console > More Tools > sensors. Test to see if the offsets work.

todo:
- [x] Times adjust to local time depending on timezone


Sboonny:
Hey, if this sounds offensive, you can ignore it, I am tired.

I am wondering, 

why do we need to use `"timezones-list"` library? I thought the plan was to use something like [Intl.DateTimeFormat.prototype.resolvedOptions()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions). I am asking this because reading [its github](https://github.com/omsrivastava/timezones-list) wasn't helpful, and helpfully you can make it clear for me.

I wonder how [daylight saving time permanent](https://www.reuters.com/world/us/us-senate-approves-bill-that-would-make-daylight-savings-time-permanent-2023-2022-03-15/0) will effect this in general? should it kept in mind going forward with this issue or am I overthinking it?

pdotsani:
> Hey, if this sounds offensive, you can ignore it, I am tired.
> 
> I am wondering,
> 
> why do we need to use `"timezones-list"` library? I thought the plan was to use something like [Intl.DateTimeFormat.prototype.resolvedOptions()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions). I am asking this because reading [its github](https://github.com/omsrivastava/timezones-list) wasn't helpful, and helpfully you can make it clear for me.
> 
> I wonder how [daylight saving time permanent](https://www.reuters.com/world/us/us-senate-approves-bill-that-would-make-daylight-savings-time-permanent-2023-2022-03-15/0) will effect this in general? should it kept in mind going forward with this issue or am I overthinking it?

@Sboonny this is not offensive at all! Thanks for the question.

We technically don't need to use the timezones library if we want to only compare the current timezone with the assigned timezone for the event. I was just experimenting with allowing the user to select from a dropdown of timezones as part of the feature. We would need to add a library like `dates-fn-tz` to handle timezone conversion as `dates-fn` doesn't seem to have that functionality.

cypress/integration/dashboard/events.js
```diff
@@ -5,8 +5,6 @@ const testEvent = {
   streamingUrl: 'https://test.event.org/video',
   capacity: '10',
   tags: 'Test, Event, Tag',
-  startAt: '2022-01-01T00:01',
-  endAt: '2022-01-02T00:02',
```
ojeytonwilliams, line 9: Github won't let me suggest, here, but I'd like to keep these so that we confirm that it's possible to enter in a date just by typing.

I tested locally and 
```
  startAt: 'January 1, 2022 12:01 AM',
  endAt: 'January 2, 2022 12:02 AM',
  ```
  worked nicely.  I'll add a comment below about how the tests would need to change.

File: cypress/integration/dashboard/events.js [hunk 2]
```diff
@@ -83,10 +81,6 @@ describe('events dashboard', () => {
     cy.findByRole('textbox', { name: 'Tags (separated by a comma)' }).type(
       'Test, Event, Tag',
     );
-
-    cy.findByLabelText(/^Start at/).type(testEvent.startAt);
-
-    cy.findByLabelText(/^End at/).type(testEvent.endAt);
```
ojeytonwilliams, line 89:
```
    cy.findByLabelText(/^Start at/)
      .clear()
      .type(testEvent.startAt)
      .type('{esc}');
    cy.findByLabelText(/^End at/)
      .clear()
      .type(testEvent.endAt)
      .type('{esc}');
```

This should work.  The escaping being necessary to hide the picker and let Cypress get to the next field.

pdotsani:
> This is looking way more robust now. Nice one!
> 
> I noticed a few things (nothing major) so I left some comments below.
> 
> One other thing I was wondering: can we just get rid of
> 
> ```js
>   if (key.endsWith('_at')) {
>     return new Date(store[field.key]).toISOString().slice(0, 16);
>   }
> ```
> 
> in EventFormUtils? I couldn't see any obvious downside to removing it, but I figured I'd check with you.

@ojeytonwilliams I think we can remove this. I also think we can remove the default values from this file as well since we declare the default date objects in the form.

ojeytonwilliams:
@pdotsani yeah, good point, that's no longer necessary.  Since this PR is ready, I'll merge it and create a quick PR to clean up the formatValue stuff.

ojeytonwilliams (approved):

ojeytonwilliams:
Thanks again, @pdotsani, it's great to be able to get this in.

PR #1052, opened 2022-05-18
feat: default user event subscription with chapter subscription

gikf:

---

- Uses subscription from _chapter user_ (if exists) for a default subscription of events.
- This means, when user unsubscribes from chapter (not yet implemented), user will be by default painlessly not subscribed for new events.
- I'm leaving TODOs, as I'm assuming some user setting to override if user is subscribed by default still might be useful.


ojeytonwilliams (approved):
LGTM :+1:

PR #1053, opened 2022-05-18
fix: handling of cancel

gikf:

---
- Fixes behavior when cancelling in modal when confirming waitlist or kicking user.
- Adds tests.

### DIFF
--- a/client/src/modules/dashboard/Events/pages/EventPage.tsx
+++ b/client/src/modules/dashboard/Events/pages/EventPage.tsx
@@ -39,18 +39,20 @@ export const EventPage: NextPage = () => {
 
   const confirmRSVP =
     ({ eventId, userId }: MutationConfirmRsvpArgs) =>
-    () => {
-      return confirm().then(() =>
-        confirmRsvpFn({ variables: { eventId, userId } }),
-      );
+    async () => {
+      const ok = await confirm();
+      if (ok) {
+        return confirmRsvpFn({ variables: { eventId, userId } });
+      }
     };
 
   const kick =
     ({ eventId, userId }: MutationDeleteRsvpArgs) =>
-    () => {
-      return confirmDelete().then(() =>
-        kickRsvpFn({ variables: { eventId, userId } }),
-      );
+    async () => {
+      const ok = await confirmDelete();
+      if (ok) {
+        return kickRsvpFn({ variables: { eventId, userId } });
+      }
     };
 
   if (loading || error || !data || !data.event) {
--- a/client/src/modules/dashboard/Events/pages/EventPage.tsx
+++ b/client/src/modules/dashboard/Events/pages/EventPage.tsx
@@ -41,7 +41,7 @@ export const EventPage: NextPage = () => {
     ({ eventId, userId }: MutationConfirmRsvpArgs) =>
     async () => {
       const ok = await confirm();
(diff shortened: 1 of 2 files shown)


client/src/modules/dashboard/Events/pages/EventPage.tsx
```diff
@@ -39,18 +39,20 @@ export const EventPage: NextPage = () => {
...
     ({ eventId, userId }: MutationConfirmRsvpArgs) =>
-    () => {
-      return confirm().then(() =>
-        confirmRsvpFn({ variables: { eventId, userId } }),
-      );
+    async () => {
+      const ok = await confirm();
+      if (ok) {
+        return confirmRsvpFn({ variables: { eventId, userId } });
+      }
```
ojeytonwilliams, line 46:
```suggestion
      if (ok)  confirmRsvpFn({ variables: { eventId, userId } });
```
Since we don't use the return value, we can compress this slightly.

client/src/modules/dashboard/Events/pages/EventPage.tsx
```diff
@@ -39,18 +39,20 @@ export const EventPage: NextPage = () => {
...
     ({ eventId, userId }: MutationDeleteRsvpArgs) =>
-    () => {
-      return confirmDelete().then(() =>
-        kickRsvpFn({ variables: { eventId, userId } }),
-      );
+    async () => {
+      const ok = await confirmDelete();
+      if (ok) {
+        return kickRsvpFn({ variables: { eventId, userId } });
+      }
```
ojeytonwilliams, line 55:
```suggestion
      if (ok) kickRsvpFn({ variables: { eventId, userId } });
```
Same nitpick as above.

cypress/integration/dashboard/events/[eventId].js
```diff
@@ -0,0 +1,89 @@
...
+        .click();
+
+      cy.get('@request').should('not.have.been.called');
+
+      cy.get('@userName').then((userName) => {
+        cy.get('@waitlistRow').contains(userName);
+      });
+    });
+
+    it.only('canceling kicking user should not remove user from event', () => {
```
Comment (ojeytonwilliams) [inline cypress/integration/dashboard/events/[eventId].js:70]:
```suggestion
    it('canceling kicking user should not remove user from event', () => {
```
This was disabling the other `it`s

ojeytonwilliams (changes requested):
For the tests,if possible, it would be good to avoid explicit DOM traversal.  Things like finding rows and searching for the ones with the appropriate name.  Ideally we want to be using either findByX methods or just attaching data-cys to the target elements.  

Those approaches help catch elements that aren't very accessible and makes our tests less likely to fail when we inevitably reorganise the UI.

cypress/integration/dashboard/events/[eventId].js
Comment (gikf) [reply cypress/integration/dashboard/events/[eventId].js:70]: Sorry, I'm always forgetting about removing it :eyes:

gikf:
I'll still be taking one more look at the tests, trying to make them more flexible.

cypress/integration/dashboard/events/[eventId].js
Comment (ojeytonwilliams) [reply cypress/integration/dashboard/events/[eventId].js]: No worries. I've never used it before, so thanks for introducing me to it 😁

ojeytonwilliams (approved):
Cheers, this LGTM :+1:

PR #1063, opened 2022-05-21
fix: Dates not matching up on all pages

pdotsani:

Closes #XXXXX

When viewing timestamp data vs the timestamp reflected in the events pages, there is a slight difference in the timestamps. By adding some logic to the input in `formatDate` we can correct this issue without having to modify the way we edit events.

### DIFF
--- a/client/src/components/EventCard.tsx
+++ b/client/src/components/EventCard.tsx
@@ -56,7 +56,7 @@ export const EventCard: React.FC<EventCardProps> = ({ event }) => {
           lineHeight="tight"
           isTruncated
         >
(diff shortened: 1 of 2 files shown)


ojeytonwilliams (approved):
Thanks @pdotsani this looks good to me. 

Timezones always confuse me, so, for future reference: the slice removes the trailing Z. Without that, format assumes the date is in the browser's timezone and formats it correctly.

pdotsani:
@ojeytonwilliams not sure about what we'll have to do about the slicing. Will need to play around with it a little more.

PR #1064, opened 2022-05-22
feat: chapter user role changing

gikf:

---
- Adds changing chapter user roles from chapter users dashboard (`/dashboard/chapters/{id}/users`).
- Modal component is easily reusable for the event user role changing.
- Basic tests included.
- Some polishing could still be used, although functionality should be working.
- Any comments and nitpickings are appreciated. I've experimented a bit, so some things may look more weird than usual.

---

image
image

### DIFF
--- a/client/src/modules/dashboard/Chapters/graphql/mutations.ts
+++ b/client/src/modules/dashboard/Chapters/graphql/mutations.ts
@@ -27,3 +27,22 @@ export const updateChapter = gql`
     }
   }
 `;
+
+export const changeChapterUserRole = gql`
+  mutation changeChapterUserRole(
+    $chapterId: Int!
+    $roleId: Int!
+    $userId: Int!
+  ) {
+    changeChapterUserRole(
+      chapterId: $chapterId
+      roleId: $roleId
+      userId: $userId
+    ) {
+      chapter_role {
+        id
+        name
+      }
+    }
+  }
+`;
--- a/client/src/modules/dashboard/shared/components/RoleChangeModal.tsx
+++ b/client/src/modules/dashboard/shared/components/RoleChangeModal.tsx
@@ -1,0 +1,52 @@
+import { Button, Select, Text, UseDisclosureReturn } from '@chakra-ui/react';
+import React from 'react';
+import { useForm } from 'react-hook-form';
+
+import { Modal } from '../../../../components/Modal';
+
+export interface RoleChangeModalData {
+  roleId: number;
+  userId: number;
+  userName: string;
+}
+
+export const RoleChangeModal: React.FC<{
+  modalProps: UseDisclosureReturn;
+  data: RoleChangeModalData;
+  roles: { id: number; name: string }[];
+  title: string;
+  onSubmit: (submitData: {
+    newRoleId: number;
+    userId: number;
+  }) => Promise<void>;
+}> = ({ modalProps, data, roles, title, onSubmit }) => {
+  const { handleSubmit, register, setValue } = useForm<{
+    newRoleId: number;
+    userId: number;
+  }>();
+
+  setValue('userId', data.userId);
+
+  return (
+    <Modal
+      modalProps={modalProps}
+      title={title}
+      buttons={<Button type="submit">Change</Button>}
+      wrapBody={(children) => (
+        <form onSubmit={handleSubmit(onSubmit)}>{children}</form>
+      )}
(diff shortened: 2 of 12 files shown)


client/src/modules/dashboard/Chapters/graphql/mutations.ts
```diff
@@ -27,3 +27,22 @@ export const updateChapter = gql`
...
+  ) {
+    changeChapterUserRole(
+      chapterId: $chapterId
+      roleId: $roleId
+      userId: $userId
+    ) {
+      chapter_role {
+        id
+        name
+      }
```
ojeytonwilliams, line 45:
```suggestion
      id
```
I know we have to return something, but since the returned values aren't used by the client, is `id` alone sufficient?

client/src/modules/dashboard/shared/components/RoleChangeModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+    newRoleId: number;
+    userId: number;
+  }) => Promise<void>;
+}> = ({ modalProps, data, roles, title, onSubmit }) => {
+  const { handleSubmit, register, setValue } = useForm<{
+    newRoleId: number;
+    userId: number;
+  }>();
+
+  setValue('userId', data.userId);
```
ojeytonwilliams, line 28: Silly question, probably, but what does `setValue` achieve over passing in `defaultValues`? If it's essentially the same, I'd recommend using `defaultValues` for consistency.

cypress/integration/dashboard/chapters/users.js
```diff
@@ -17,4 +17,18 @@ describe('Chapter Users dashboard', () => {
...
+
+    cy.get('[data-cy=changeRole]').first().click();
+    cy.findByRole('combobox').select('member');
+    cy.findByRole('button', { name: 'Change' }).click();
+    cy.get('[data-cy=role]').first().contains('member');
+
+    cy.get('[data-cy=changeRole]').first().click();
+    cy.findByRole('combobox').select('organizer');
+    cy.findByRole('button', { name: 'Change' }).click();
+    cy.get('[data-cy=role]').first().contains('organizer');
```
ojeytonwilliams, line 32: How about filtering the list to select the first member and the first organiser? Then it's possible to check that the click on that row's button changed that specific user in the expected way.

ojeytonwilliams (changes requested):
This looks nice, I only have a few nitpicks:

client/src/modules/dashboard/shared/components/RoleChangeModal.tsx
gikf: Some questions can be explained with me not being too versed with react related things.

I think ideally would be to not need passing `userId` to the component at all, as it doesn't do anything there. I did that because I was having some issues with _Object is possibly 'undefined'._ when trying to use `chapterUser.userId` in the `onModalSubmit`.
```ts
    changeRoleMutation({
      variables: {
        chapterId: id,
        roleId: data.newRoleId,
        userId: chapterUser.userId,
      },
    });
```

When checking using `defaultValues`:
```ts
  const { handleSubmit, register } = useForm<{
    newRoleId: number;
    userId: number;
  }>({ defaultValues: { userId: data.userId } });
```
Role is changed only for the single user - for whom modal was open first. Opening modal for different user will still change role for the first one.

---
Looking at this exposed bug that I didn't notice before - the selected role in the modal doesn't change when opening it for different users. So I'm going to look at it some more.

ojeytonwilliams: Ah, thanks for the explanation, that's some strange behaviour from react-hook-form.

ojeytonwilliams (commented):
This all looks great from my perspective.  

Did you squash the bug you were looking into?  If not, let me know if I can be of any help.

gikf:
> This all looks great from my perspective.
> 
> Did you squash the bug you were looking into? If not, let me know if I can be of any help.

Yes, last part of the test is trying to target that, so it doesn't sneak back.

ojeytonwilliams:
> Yes, last part of the test is trying to target that, so it doesn't sneak back.

Ah, nice, that bit makes sense now.

ojeytonwilliams (approved):

ojeytonwilliams:
Thanks again, @gikf always a pleasure!

Issue #782, opened 2021-11-09
As a user, I will receive an email when my RSVP is confirmed


ojeytonwilliams:
Organisers can confirm the RSVPs of members who are on the waitlist.  When they do this, the user should be notified by email that they can attend the event.

This should not require any UI changes, just a new notification triggered by confirmation.

allella:
I don't recall if waitlist was an MVP feature. It was added into the database, but if it's complicated we can leave the fields there for future use and add close the issue and add it to the Roadmap.

It's up to you if you think it's going to prolong the MVP.
https://github.com/freeCodeCamp/chapter/pull/27

ojeytonwilliams:
It should not be overly tricky to implement, since we already have similar functionality for venue changes.  So, it shouldn't create much delay.

Jagrutiti:
Hi, @ojeytonwilliams 

I would like to work on this issue. 

I didn't exactly understand what is required to be done. Could you please go into more detail.

Thanks

allella:
@Jagrutiti 

#909 has the most recent work on sending emails.  That's not been merged in, but you might take a look at what @gikf was doing on that PR related to sending emails using the pattern they've been working out on that PR.

Ismailtlem:
Hi @Jagrutiti 

Did you start working on that ?

Jagrutiti:
Hey @Ismailtlem 

I am not working on this one. Feel free to take it up.

Ismailtlem:
Hi,

- How can we link an event to a chapter on the UI ?
- Where do we specify organizers on an event ?
- Can we register to the app by an email other than foo@bar.com

Nirajn2311:
Hi @Ismailtlem , have you started working on this?

Ismailtlem:
Hi, I started working on it, I am just stuck a bit on sending the email part. I made the UI part. I will make a draft PR asap and we can take it from there

Ismailtlem:
Should we close this ?

allella:
@ojeytonwilliams is there anything more to do on this one since #1065 was merged?

ojeytonwilliams:
I don't believe so, no, @Ismailtlem's PR looks to have covered it. We can tick off another MVP issue :rocket:

PR #1065, opened 2022-05-24
feat(client): send email on rsvp confirmation

Ismailtlem:

Related to #782

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -322,8 +322,14 @@ export class EventResolver {
     if (!ctx.user) throw Error('User must be logged in to confirm RSVPs');
     const eventUser = await prisma.event_users.findUnique({
       where: { user_id_event_id: { user_id: userId, event_id: eventId } },
-      include: { event: true },
+      include: { event: true, user: true },
     });
+
+    await new MailerService({
+      emailList: [eventUser.user.email],
+      subject: 'Your RSVP is confirmed',
+      htmlEmail: `Your reservation is confirmed. You can attend the event ${eventUser.event.name}`,
+    }).sendEmail();
 
     return await prisma.event_users.update({
       data: { rsvp: { connect: { name: 'yes' } } },


server/src/controllers/Events/resolver.ts
```diff
@@ -324,6 +324,19 @@ export class EventResolver {
       where: { user_id_event_id: { user_id: userId, event_id: eventId } },
       include: { event: true },
```
gikf, line 325:
```suggestion
      include: { event: true, user: true },
```
Gets also user from the relation, event was already included.

server/src/controllers/Events/resolver.ts
```diff
@@ -324,6 +324,19 @@ export class EventResolver {
...
     });
+    const userInfos = await prisma.users.findUnique({
+      where: { id: userId },
+    });
+
+    const eventInfos = await prisma.events.findUnique({
+      where: { id: eventUser.event_id },
+    });
```
gikf, line 333: Then this isn't needed.
```suggestion
```

server/src/controllers/Events/resolver.ts
```diff
@@ -324,6 +324,19 @@ export class EventResolver {
...
+
+    await new MailerService({
+      emailList: [userInfos?.email],
```
gikf, line 336: And wanted data can be taken from `eventUser`.
```suggestion
      emailList: [eventUser.user.email],
```

server/src/controllers/Events/resolver.ts
```diff
@@ -324,6 +324,19 @@ export class EventResolver {
...
+      subject: 'Your RSVP is confirmed',
+      htmlEmail: `Your reservation is confirmed. You can attend the event ${eventInfos?.name}`,
```
gikf, line 338:
```suggestion
      htmlEmail: `Your reservation is confirmed. You can attend the event ${eventUser.event.name}`,
```

gikf (commented):
This works nicely. There's way to do this without additional queries, but relying on the `event_users` relations with `events` and `users`. See specific suggestions for details.

server/src/controllers/Events/resolver.ts
```diff
@@ -322,9 +322,15 @@ export class EventResolver {
...
     const eventUser = await prisma.event_users.findUnique({
       where: { user_id_event_id: { user_id: userId, event_id: eventId } },
-      include: { event: true },
+      include: { event: true, user: true },
     });
 
+    await new MailerService({
+      emailList: [eventUser.user?.email],
+      subject: 'Your RSVP is confirmed',
+      htmlEmail: `Your reservation is confirmed. You can attend the event ${eventUser.event?.name}`,
```
gikf, line 331: Might be a bit nitpicky, but I don't think the optional chaining is needed here. Not sure if it's even possible for `user` or `event` to not be there at this point.

gikf (approved):
Looks good! :+1:

PR #1066, opened 2022-05-26
refactor: Reminders service

gikf:

---
- Extracts creating, deleting, getting, etc. of reminders to separate _service_ from resolver and reminders sending. This basically means any direct interaction with prisma.
- Other than replacing old code with function call there aren't any changes.
- Name _service_ might not be too accurate, but moving this to _util_ seems a bit more out of place.
- Thing that currently sticks out the most is whole locking while processing reminders. Sending reminder might not need to care about that. Some further step, for better isolating, could be extracting that to service as well, and using custom iterator. Which would yield specific reminder only if lock is obtained.

### DIFF
--- a/server/reminders/reminders.ts
+++ b/server/reminders/reminders.ts
@@ -1,5 +1,12 @@
-import { prisma } from '../src/prisma';
 import MailerService from '../src/services/MailerService';
+import {
+  deleteReminder,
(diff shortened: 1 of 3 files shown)


moT01 (approved):
LGTM 🎉

Issue #98, opened 2019-10-20
As an owner or Chapter admin I can ban a participant whom I believe is toxic or who has previously broken my organization's code of conduct.


QuincyLarson:
This needs UI and resolvers to handle the requests. The DB schema has _user_bans_ already, so a given user can be banned from a specific Chapter.

Until https://github.com/freeCodeCamp/chapter/pull/814 is in, we can make some progress by implementing the UI for banning users and a resolver to update the DB. After the auth PR is in, we'll need to consider if 'banned' becomes a role that denies you access or if we need to handle banned user access differently.

- [ ] As an owner I can ban anyone except other owners
- [ ] As an admin of a Chapter I can ban anyone in a Chapter except other admins
- [ ] As any other role, I cannot ban anyone

The main thing to check is that requests from unauthorised users fail, but it's also helpful to confirm that the UI reflects their role.

We will also need tests to confirm that banned users can no longer interact with Chapters they are banned from.

ojeytonwilliams:
Once https://github.com/freeCodeCamp/chapter/pull/1069 goes in, authorisation will need to be modified to take account of ban status.

ojeytonwilliams:
While not all resolvers are restricted by Authorized decorators, whenever they are restricted, ban status is automatically taken into account.

As a result, it doesn't seem necessary to track this any more.

PR #1069, opened 2022-05-30
feat: adding chapter user bans

gikf:

To do
- [x] I guess deleting bans too?
- [x] ~~Don't query `user_bans` in client~~
- [x] Tests
  - [x] Test
  - [x] Fully working

---
- Related to #98
- Adds UI portion for banning chapter user in the `dashboard/chapters/{id}/users`.
- Adds resolver adding ban to database.
- Nothing of the actual blocking access is tackled here.
- Uses ~~two~~ one custom `FieldResolver`s on the `ChapterUser` to handle most of the verification _who can what and when_ on server side.
  - ~~`isBanned` checks if user is among banned for the chapter. This might need changing/tuning, as it makes additional query for each user.~~
  - `canBeBanned` checks if user permissions (user chapter role) allows for banning specified chapter user.
- Currently `owner` (although such chapter role isn't created) and `organizer` can ban, this is mainly due to `organizer` being seeded, and being relied upon in other places.
- Any comments and nitpicks are appreciated.


gikf:
I'm having weird issue with making part of test work as intended.
https://github.com/freeCodeCamp/chapter/blob/7f13764f49537592d731486ff7373e6573e73ff7/cypress/integration/dashboard/chapters/users.js#L65
Idea is to find user for test banning, by filtering all members and leaving only those that don't have element with `data-cy=isBanned`. Then pick first from them. That doesn't work however, and once in a while user selected (first `member` on page) for banning will be already banned.

ojeytonwilliams:
@gikf did your last commit fix the test reliability issues you were running into?

gikf:
@ojeytonwilliams yes, although using something else to filter-out users that are not relevant for test.

ojeytonwilliams (approved):
This looks great @gikf. I'm going to merge this as is and then do a bit of refactoring.  Because we have quite a lot of ids floating around I'd like to use 'named' ids (e.g. `chapterId` not just `id`).  We need that for the authorization logic and I think it'll make some of the logic a bit easier to parse.

PR #1079, opened 2022-06-11
chore: add no-only-tests eslint rule

gikf:

---

- Adds no-only-tests eslint rule to catch leftover `.only` in the tests during linting.
- Editor will complain with it, and it will be caught during CI.
- This however won't stop committing it, as lint isn't run on commit (maybe it should be added?), but only prettier is run.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,7 +6,7 @@
     "plugin:cypress/recommended",
     "plugin:import/recommended"
   ],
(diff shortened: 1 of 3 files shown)


ojeytonwilliams (approved):
As soon as I learned about `only` from your PR, I started committing it :man_facepalming:

So, yeah, this LGTM!

gikf:
With great power comes great responsibility :smile:

PR #1081, opened 2022-06-14
feat: require authorization to rsvp

ojeytonwilliams:
This is a trivial requirement, because all instance members have it and
it might be smarter to modify the authorization checker to allow empty
`Authorized()` decorators, but let's try this approach for now.

### DIFF
--- a/server/prisma/generator/factories/chapterRoles.factory.ts
+++ b/server/prisma/generator/factories/chapterRoles.factory.ts
@@ -4,6 +4,7 @@ const chapterPermissions = [
   'chapter-edit',
   'event-create',
   'event-edit',
+  'rsvp',
 ] as const;
 
 type Permissions = typeof chapterPermissions[number];
@@ -11,11 +12,11 @@ type Permissions = typeof chapterPermissions[number];
 const roles: Array<{ name: string; permissions: Permissions[] }> = [
   {
     name: 'administrator',
-    permissions: ['chapter-edit', 'event-create', 'event-edit'],
+    permissions: ['chapter-edit', 'event-create', 'event-edit', 'rsvp'],
   },
   { name: 'organizer', permissions: ['event-create', 'event-edit'] },
   { name: 'moderator', permissions: ['event-edit'] },
-  { name: 'member', permissions: [] },
+  { name: 'member', permissions: ['rsvp'] },
 ];
 
 const createChapterRoles = async () => {
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -3225,6 +3225,22 @@
             "description": null,
             "args": [
               {
+                "name": "chapterId",
+                "description": null,
+                "type": {
+                  "kind": "NON_NULL",
+                  "name": null,
+                  "ofType": {
+                    "kind": "SCALAR",
+                    "name": "Int",
+                    "ofType": null
+                  }
+                },
+                "defaultValue": null,
+                "isDeprecated": false,
+                "deprecationReason": null
+              },
+              {
                 "name": "eventId",
                 "description": null,
                 "type": {
--- a/client/src/generated/graphql.tsx
+++ b/client/src/generated/graphql.tsx
@@ -359,6 +359,7 @@ export type MutationRegisterArgs = {
 };
 
 export type MutationRsvpEventArgs = {
+  chapterId: Scalars['Int'];
   eventId: Scalars['Int'];
 };
 
@@ -1198,6 +1199,7 @@ export type VenueQuery = {
 
 export type RsvpToEventMutationVariables = Exact<{
   eventId: Scalars['Int'];
+  chapterId: Scalars['Int'];
 }>;
 
 export type RsvpToEventMutation = {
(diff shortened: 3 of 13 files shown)


ojeytonwilliams:
Thanks for raising these concerns, @gikf. Bans really complicate things.

So, here's my current take on this (TL;DR: this PR won't work in the current form)

Once a user has been banned from a particular level, they shouldn't be able to do anything at that level. I.e. it should be equivalent to revoking their permissions. However, bans below that level should matter. Why? Because it doesn't make sense to me that, say, an event organiser could ban the administrator of their chapter. If that needs to happen, it's something to be handled in a different way. Not in code, basically.

This means this PR can't work (since chapter bans would not stop a banned user RSVPing to that chapter's events) and that we'd need to rework RSVPing to first attempt to join the chapter and to check ban status.

Another approach I considered was to check ban status first and reject a request if they're banned at any level relevant to the request. I.e. make bans take precedence over permissions. However, that means we have to prevent lower level users from banning higher level users. We will probably want to do that, regardless, but I suspect encoding that in the authorization logic is a reasonably safe approach.

What do you think? Do either of the approaches make sense to you?

Apologies if that wasn't clear, I'm still working out what I think. Just let me know if I need to clarify anything

gikf:
To clarify, the way I currently understand user bans, and how adding bans fits in that. Bans are only per chapter, no instance or event bans as such. When user is banned from chapter I'm assuming no further interactions with chapter and events from that chapter will be allowed.

I haven't really explored possibility of situation when chapter owner (A), bans chapter member (B), who (B) is also instance administrator. Some easy sounding way to deal with it (or actually to try to not have to deal with it too much) that comes to my mind, is making that certain roles cascades lower in the `instance > chapter > event` relation. So for example instance administrator, when joining chapter will automatically get the chapter role that's equivalent of administrator.

ojeytonwilliams:
> To clarify, the way I currently understand user bans, and how adding bans fits in that. Bans are only per chapter, no instance or event bans as such. When user is banned from chapter I'm assuming no further interactions with chapter and events from that chapter will be allowed.

Sorry, my bad: for the MVP that's exactly right. Post MVP it will likely get more complicated and I was trying to make sure the broad plan is okay.

> So for example instance administrator, when joining chapter will automatically get the chapter role that's equivalent of administrator.

That might be a level of complexity we don't need, tbh. Ideally a user should have just one role. After all, if I'm a chapter administrator I don't need additional permissions for any events in my chapter. Giving me roles for those events feels unnecessary.

ojeytonwilliams:
> But yeah, this is all getting complicated.

Yes, unfortunately it is. 

Now that I come back to this, I reckon you were right about needing roles at each level.  While a single user having multiple roles complicates things (obviously) it dramatically simplifies the auth+ban logic, so it's probably worth it.

So, even if I'm an instance owner, joining chapters and events should grant me roles at those levels.  Now consider RSVPing. We don't have to give new members an instance permission of `rsvp`.  Instead, we grant them a chapter role to give them the permission.

Once we allow each user to have _instance_, _chapter_ and _event_ roles then the auth+ban logic can be as simple as this:

1. If allowed by instance role, allowed
2. If has chapter ban, forbidden
3. If allowed by chapter role, allowed
4. If has event ban, forbidden 
5. If allowed by event role, allowed

Not that we're doing event bans in the MVP, but you get the idea. Also, we could add instance bans, but, again, not in the MVP.

This means that instance users cannot be banned at all, so it won't be possible for a chapter admin to ban an instance owner.

Final thing: this means that we can grant the lowest possible role to *all* users that join chapters and events, regardless of other any other roles they have.  This has the benefit of simplicity and is secure by default.

ojeytonwilliams:
@gikf if that logic makes at least some sense to you, I'll add the ban-auth logic, tests and whatnot in a followup PR.

gikf:
@ojeytonwilliams It makes sense as far as I can tell. I was wondering about `rsvp` permission in the administrator instance role, that's added so administrator is still not affected by potential chapter or event bans, right?

server/prisma/generator/factories/chapterRoles.factory.ts
```diff
@@ -4,18 +4,19 @@ const chapterPermissions = [
...
 type Permissions = typeof chapterPermissions[number];
 
 const roles: Array<{ name: string; permissions: Permissions[] }> = [
   {
     name: 'administrator',
-    permissions: ['chapter-edit', 'event-create', 'event-edit'],
+    permissions: ['chapter-edit', 'event-create', 'event-edit', 'rsvp'],
   },
   { name: 'organizer', permissions: ['event-create', 'event-edit'] },
   { name: 'moderator', permissions: ['event-edit'] },
```
gikf, line 18: I guess these two also should get `rsvp` permission.

ojeytonwilliams:
> I was wondering about rsvp permission in the administrator instance role, that's added so administrator is still not affected by potential chapter or event bans, right?

This is all still being nailed down, but my current understanding is that the top-level roles will be able to do everything.  You're right that they also need the `rsvp` permission to make the ban logic work, but it simplifies our mental model too.

So, the current plan for the roles in MVP is this:

- Instance owners
- Chapter admins
- Chapter members

In order of how broad their permissions are.  Instance owners have all permissions on all Chapters.  Chapter admins have permission to do anything within their Chapter. Chapter members have the least permissions of all - RSVPing/Subscribing and not much else.

Post MVP we can think about adding more roles (event organisers and so on), but I think these three will be enough for now.

server/prisma/generator/factories/chapterRoles.factory.ts
ojeytonwilliams: Yeah, post MVP, but for now I think we can get rid of these roles and (try to) keep things reasonably simple.

ojeytonwilliams:
As for the actual plan, once this PR and https://github.com/freeCodeCamp/chapter/pull/1092 are in, I'll get to work fleshing out those three roles, removing the others and adding the Authorization decorators.

ojeytonwilliams:
Lovely. Thanks for going through this, it's always appreciated.

PR #1084, opened 2022-06-15
refactor: generic $id to named ids

ojeytonwilliams:
This is necessary for authorization, since the checker cares about which chapter/event a request wants to affect.  It lacks context about how a variable will be used, so the variable must be named appropriately (`chapterId` if it targets a chapter, `eventId` for events).

It only cares about the variable name, not what the resolver is actually called with, so I normalised them to use `id` as their argument.  For consistency, it's probably an idea to change those too, but I didn't want to do too much in one PR.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -3093,7 +3093,7 @@
             "description": null,
             "args": [
               {
-                "name": "event_id",
+                "name": "id",
                 "description": null,
                 "type": {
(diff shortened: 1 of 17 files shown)


gikf (commented):
Just a couple of nitpicks below

ojeytonwilliams:
Thanks for catching those @gikf

gikf (approved):
Looks good! :+1:

PR #1091, opened 2022-06-20
Wording changes and minor tweaks to CONTRIBUTING.md

allella:
- Removed "Fork" screenshot and linked to the GitHub Fork docs instead as it seems unnecessary to show where the "Fork" button is when the docs can help people who are confused.
- Reworded some of the development terminology to be more concise or to clarify

Closes nothing.

These are some changes I made to on older branch but never got into a PR.


allella:
We might consider deleting docs/assets/how-to-fork.png if it's not going to be used.

ojeytonwilliams (approved):
I don't see a reason to keep the png around, so please feel free to delete it.  Regardless, the updates LGTM

allella:
@ojeytonwilliams the only two screenshots I think we're still using are below, and one of them still shows "master".

- https://github.com/freeCodeCamp/chapter/blob/main/docs/assets/comparing-forks-for-pull-request.png
- https://github.com/freeCodeCamp/chapter/blob/main/docs/assets/pull-request-prompt.png

There are a [few other images someone dropped into the docs/assets which were from an issue and we figured could be deleted](RegularSER.png) because someone uploaded files to the repo instead of embedding them into the issue.

Do you gather we can kill off the docs/assets and then we have one less confusing top-level docs/ directory in the repo?

ojeytonwilliams:
I think all the images can go.  The two that are still in use do not seem vital, particularly since they're dealing with generic contributing process rather than Chapter specifics.  So, yes, if you agree, then I'm happy for the docs directory to be removed.

ojeytonwilliams (approved):
LGTM :+1:

PR #1092, opened 2022-06-21
feat: handle bans in authorization

ojeytonwilliams:
With these changes a user that is banned from a Chapter will fail every authorization check for requests related to that Chapter. Users with instance level permissions will not be affected.

While implementing this I also cleaned up the types and made `authorizationChecker` explicitly synchronous.  Since we get the user's data in the _authentication_ middleware, there's no need to query the db during _authorization_.

Cypress tests to follow - I'd like to get https://github.com/freeCodeCamp/chapter/pull/1081 in first.


gikf (approved):
Looks good to me :+1:

Did you want to add something else here?

PR #1096, opened 2022-06-23
feat: reduce available roles

ojeytonwilliams:
This is almost an anti-feature since it removes some functionality.  Specifically this reduces the number of roles to 5

- Instance owner and member
- Chapter administrator and member
- Event member

The original discussion can be found here: https://github.com/freeCodeCamp/chapter/pull/1081#issuecomment-1155492726

Since it no longer made sense to change event roles, I removed that feature.  Hopefully to return post MVP - it was working well, after all.

Next steps after this PR: move all the authorization logic into the `authorizationChecker` function.


server/prisma/generator/setupRoles.ts
```diff
@@ -4,23 +4,44 @@ import { prisma } from '../../src/prisma';
...
+    };
+
+    usersData.push(ownerData);
+
+    const adminData: Prisma.chapter_usersCreateManyInput = {
       joined_date: new Date(),
       chapter_id: chapterId,
       user_id: adminId,
-      chapter_role_id: chapterRoles.organizer.id,
+      chapter_role_id: chapterRoles.administrator.id, // This user is an instance owner
```
gikf, line 36:
```suggestion
      chapter_role_id: chapterRoles.member.id, // This user is an instance owner
```
Seems like member was intended here.

gikf:
Also with removing event organizer role, what happens with notifying event organizers about new rsvps?

server/prisma/generator/setupRoles.ts
ojeytonwilliams: The code is how I intended it, but the comment is completely wrong.  The idea with this user is that we can write tests that differentiate between an instance owner and a chapter admin, so they do need the `administrator` role.

ojeytonwilliams:
> Also with removing event organizer role, what happens with notifying event organizers about new rsvps?

Very good question.  For the MVP the chapter admins will have to handle everything that will, eventually, be the event organizer's responsibility. So, that includes getting notifications when someone RSVPs.

I'm surprised none of the tests failed when I changed this, so I'll address that next.

ojeytonwilliams:
@gikf Okay, I think that's everything.  If you've got a moment to check, awesome, but I'm happy with the current state of it.

server/src/controllers/Events/resolver.ts
```diff
@@ -207,6 +212,11 @@ export class EventResolver {
       },
     });
 
+    const chapterUsers = await prisma.chapter_users.findMany({
+      where: { chapter_id: chapterId },
```
gikf, line 216:
```suggestion
      where: {
        chapter_id: chapterId,
        subscribed: true,
        chapter_role: { is: { name: 'administrator' } },
      },
```
Not too sure if this is worth changing, but after this, only chapter administrators would be passed to the `rsvpNotifyAdministrators`. This could slightly cut function, and `chapterUsers` variable could be renamed.

My mind was having hard time for a while, to connect why exactly `chapterUsers` are being passed to `rsvpNotifyAdministrators`.

gikf (commented):
Generally this looks good to me :+1: 

There's one place that stood out for me, but I wouldn't say that's a necessarily required change.

server/src/controllers/Events/resolver.ts
ojeytonwilliams: Thanks, @gikf, this approach makes much more sense.

gikf (approved):

PR #1100, opened 2022-06-24
feat: users dashboard

gikf:

---

- Adds users dashboard - `/dashboard/users`.
- Adds changing instance role.
- Includes new permissions for viewing instance users, and changing role. It's quite fun to use `Authorized` decorator.
- Adds some basic tests for the happy path (user having required permissions). Side note - tests _describe_ that with permission (instead of administrator) these actions are allowed, as that seemed slightly more correct for me.
- Not entirely sure what scope pick for the sad paths. Funny story, I had to change test, as changing seeded administrator (logged in user) will block from viewing users.
- Some naming could be better.


ojeytonwilliams (changes requested):
The decorators are pretty nice, I agree. It puts the information exactly where you need it.

Could you add a confirm modal before the role is changed? It's fairly high impact and not something that users will be doing often, so the extra inconvenience seems appropriate.

Otherwise this looks great - just the one tiny nitpick:

gikf:
I used the suggestion and added confirmation modal. :+1:

ojeytonwilliams:
Perfect, thanks @gikf.  As went through again I spotted one async function with no awaits which seemed safe to drop, so I did. I hope that's okay!

Anyways, this LGTM :+1:

ojeytonwilliams (approved):

PR #1121, opened 2022-06-29
fix: remove redundant cancelled/lock elements

ojeytonwilliams:
The extra UI elements were not necessary. Also, occasionally the tests
would fail since showing the canceled tag would hide the lock icon and
the tests assumed at least one lock icon would exist.


gikf (approved):
LGTM :+1: 

It also seem to look nicer now.

ojeytonwilliams:
Thanks @gikf I'm not sure how the extra elements crept in, but they didn't look great.

To be honest, I was only trying to fix the test, but it turned out the test was doing its job!

PR #1122, opened 2022-06-29
feat: prevent non-owners from creating chapters

ojeytonwilliams:

With this change, it's no-longer sufficient to be logged in if you want to create a chapter, you must have the correct permission.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -38,7 +38,7 @@
               "name": null,
               "ofType": {
                 "kind": "OBJECT",
-                "name": "User",
+                "name": "UserWithInstanceRole",
                 "ofType": null
               }
             },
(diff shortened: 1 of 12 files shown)


ojeytonwilliams:
Ah, nice catch. I'm guessing that the me query needs refetching when a user logs in.  I'll take a look.

ojeytonwilliams:
> This appears to be related to fact that Auth resolver - authenticate, returns just User.

Yeah, you're right.  One of my mistakes was using the me query in the first place. It should have all been done via the `useAuth` hook.  Ah well, I'll fix that tomorrow.

Thanks again for the thorough QA.

gikf (approved):
LGTM :+1: 

Thanks for the update.

PR #1124, opened 2022-06-29
feat: separate role creation from seeding

ojeytonwilliams:
This isn't necessary *yet*, but it will be once we want to deploy. Rather than waiting for the last minute, I want to start separating development from production now.

(while I was in the code, I switched to `reduce` to make sure TypeScript will not resort to assigning types to `any`)


gikf (approved):
Looks good :+1:

PR #1130, opened 2022-06-30
fix(gitpod): remove extra db config

ojeytonwilliams:
Gitpod declares a environment variable DATABASE_URL of
postgresql://gitpod@localhost. We could delete that and create our
own but if that is sufficient, we can reduce the complexity of the
configuration.


gikf (approved):
Looks good :+1: 

I haven't found any issues.

ojeytonwilliams:
It's all a bit strange. About half the time it works flawlessly, but about half the time it's not possible to reach the DB.  I'll keep tinkering.

PR #1131, opened 2022-06-30
doc(markdown): changed links of chat to discord

Sboonny:

This came to my notice by a message in discord, we are moved to discord as a way to communicate with contributors.

changed the links to our discord link.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ Hello :wave: and welcome to **_Chapter_**, a project of [freeCodeCamp](https://w
 
 We strictly enforce our ["Code of Conduct"](https://www.freecodecamp.org/code-of-conduct), so please take a moment to read the 196 word policy.
 
-[Join our chat](https://chat.freecodecamp.org/channel/chapter) to get connected with the project's development team.
+[Join our chat](https://discord.gg/PRyKn3Vbay) to get connected with the project's development team.
 
 # Contributing Code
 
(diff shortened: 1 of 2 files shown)


ojeytonwilliams:
Is it possible to create a link directly to the chapter channel?  I think that would be great, since we have a *lot* of channels in the discord.

ojeytonwilliams:
It's intentional, if unfortunate. Chat was working quite well for staff conversations, but not very well for contributors.  We weren't seeing a great deal of contributor activity. I think this was in part because we never retired the discord (so had split the community) and in part because we don't have notifications in chat.

Hopefully this will be the last move for a while.

allella:
Okay. So, I agree we should like the the #chapter channel

https://discord.com/channels/692816967895220344/988495800852152450

ojeytonwilliams:
I think so too. If someone lands in the chapter channel and later wants to explore the others, that's fine, but it would be annoying to have to search for it initially.

What I'm not sure about (and hoping @nhcarrigan knows more) is do we need anything other than that link you shared?  Should I create one for the chapter channel or not?

allella:
@Sboonny Here's my experience with using the #chapter specific link.

I already had an account with Discord from Chapter's previous use of the fCC Discord.  I was able to log in, but it treated me like a new user and forced a dialog box to accept the rules (photo below) before I was allowed to start posting. 

If the concern about the #chapter specific link is that someone new might miss the rules, then the force pop-up my eliminate that as a concern.

If we end up linking to the more general Discord link, then we should at least update the docs with text directing people to the #chapter.

image

naomi-lgbt (approved):
This approach is correct. We can rely on Discord's membership screening feature (that rules pop up) to onboard folks.

allella:
@Sboonny @nhcarrigan the links in the latest commit go to the correct place, but they also prompted me with an invite screen saying _Sboon_ invited me, even though I was already logged in.  Is that a desired feature?

image

allella:
Alright, I think the [unshorted URL](https://discord.com/channels/692816967895220344/988495800852152450) is also an option.

Sboonny:
If make new account and press url unshorten link, I can't access anything in the page

image

but if search preview, I can see the server.

image


I want to create a direct preview link, but Google isn't helping much.

nhcarrigan:
> but they also prompted me with an invite screen saying Sboon invited me, even though I was already logged in. Is that a desired feature?

Yes, that works.

> Alright, I think the [unshorted URL](https://discord.com/channels/692816967895220344/988495800852152450) is also an option.

This isn't available for users who haven't already joined the server.

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix

ojeytonwilliams (approved):
As always, thanks for the insight @nhcarrigan and the testing, @allella. Based off that, I think this does what we want.

LGTM. Thanks @Sboonny

PR #1172, opened 2022-07-03
chore: cypress custom commands clean up

gikf:

---
- Adds missing custom commands.
- Slightly cleans two existing commands.


ojeytonwilliams (approved):
The fact we have to maintain this makes me wonder if we'd be better off just migrating to TS :thinking: 

Regardless, LGTM :+1:

PR #1186, opened 2022-07-06
feat: restrict sub/unsub from event

ojeytonwilliams:
- refactor: DRY logic testing if user is rejected
- test: confirm users cannot sub/unsub without perms
- chore: pass chapterId to resolver
- chore: regenerate code
- fix: pass chapterId to mutation
- fix(test): make assertions more specific
- fix: add sub/unsub permissions to member

While I don't entirely love sending `chapterId` as a variable to a resolver that doesn't use it, I don't see a better approach.  The auth checker needs to know if a user is allowed to sub/unsub depending on their chapter role.  It's particularly weird here because the resolver will throw if the user has no _event_users_ entry.  Regardless, that isn't the concern of the auth checker and it should allow the resolver to be reached.

Finally, we can't remove `chapterId` and rely on `getRelatedChapterId`, because that will only take account of the _event_users_ table and a given chapter admin may not have an entry in that table.


ojeytonwilliams:
In draft until I

- [ ] Reduce the no. of permissions to 1
- [ ] Provide types for the commands

which is to say, make it more like https://github.com/freeCodeCamp/chapter/pull/1187 😰

ojeytonwilliams:
> Without event_users entry, the chapter administrator's subscription to an event is in a limbo.

Good point, you're absolutely right.

Just to reiterate (mostly for my benefit): if a Chapter admin attempts to change the subscription to an event for which they don't have a `user_events` entry, it should fail.  The fact it fails in the authorization layer is slightly weird since they should have permission to do anything in that chapter. However, it's fine, because it doesn't make sense to manage a non-existent subscription.

Thanks a ton, @gikf. I'll sort this mess out next week!

gikf (approved):
Looks good to me :+1:

PR #1187, opened 2022-07-06
feat: restrict sponsors management

gikf:

---
- Restricts management of sponsors to owner ~~and chapter admin~~. (It doesn't make sense for this to be chapter role permission).
- Uses just single single permission, scope here seemed little enough to not split it in two.
- Doesn't display _Sponsors_ button in dashboard if user doesn't have permission. A bit of experimentation.
- Adding deleting sponsors for MVP might be worth consideration (and also venues).

### DIFF
--- a/client/src/modules/dashboard/shared/components/Layout.tsx
+++ b/client/src/modules/dashboard/shared/components/Layout.tsx
@@ -11,7 +11,7 @@ const links = [
   {
     text: 'Sponsors',
     link: '/dashboard/sponsors',
-    permission: 'sponsors-manage',
+    requiredPermission: 'sponsors-manage',
   },
   { text: 'Users', link: '/dashboard/users' },
 ];
@@ -24,16 +24,15 @@ export const Layout = ({ children }: { children: React.ReactNode }) => {
       <HStack as="nav" my="2">
         {links
           .filter(
-            ({ permission }) =>
-              typeof permission === 'undefined' ||
-              useCheckPermission(permission),
+            ({ requiredPermission }) =>
+              !requiredPermission || useCheckPermission(requiredPermission),
           )
           .map((item) => (
             <LinkButton
               key={item.link}
               href={item.link}
               colorScheme={
-                router.pathname.startsWith(item.link) ? 'blue' : undefined
+                router.pathname.startsWith(item.link) ? 'blue' : 'gray'
               }
             >
               {item.text}
--- a/client/src/modules/dashboard/shared/components/Layout.tsx
+++ b/client/src/modules/dashboard/shared/components/Layout.tsx
@@ -2,12 +2,17 @@ import { HStack } from '@chakra-ui/layout';
 import { LinkButton } from 'chakra-next-link';
 import { useRouter } from 'next/router';
 import React from 'react';
+import { useCheckPermission } from 'hooks/useCheckPermission';
 
 const links = [
   { text: 'Chapters', link: '/dashboard/chapters' },
   { text: 'Events', link: '/dashboard/events' },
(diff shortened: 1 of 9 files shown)


cypress/e2e/dashboard/sponsors.cy.js
```diff
@@ -0,0 +1,69 @@
...
+
+    cy.findByRole('button', { name: 'Save Sponsor Changes' }).click();
+    cy.findByRole('heading', { name: 'Sponsors' });
+    cy.findByRole('link', { name: testSponsor.name }).click();
+    cy.get('[data-cy=name]').contains(testSponsor.name);
+    cy.get('[data-cy=website]').contains(testSponsor.website);
+    cy.get('[data-cy=type]').contains(testSponsor.type);
+  });
+
+  it('prevents members from managing sponsors', () => {
```
ojeytonwilliams, line 59:
```suggestion
  it('prevents chapter admins from managing sponsors', () => {
```
We might as well test against admins. If they can't do it, members should also be prevented.

client/src/modules/dashboard/shared/components/Layout.tsx
```diff
@@ -2,12 +2,17 @@ import { HStack } from '@chakra-ui/layout';
...
 
 const links = [
   { text: 'Chapters', link: '/dashboard/chapters' },
   { text: 'Events', link: '/dashboard/events' },
   { text: 'Venues', link: '/dashboard/venues' },
-  { text: 'Sponsors', link: '/dashboard/sponsors' },
+  {
+    text: 'Sponsors',
+    link: '/dashboard/sponsors',
+    permission: 'sponsors-manage',
```
ojeytonwilliams, line 14:
```suggestion
    requiredPermission: 'sponsors-manage',
```
It's a bit verbose, but I think it's clearer.  

Also, I think it would be good to share `Permission` between the client and the server. That's a little involved and out of scope, though, so I'll take care of that separately.

client/src/modules/dashboard/shared/components/Layout.tsx
```diff
@@ -17,17 +22,23 @@ export const Layout = ({ children }: { children: React.ReactNode }) => {
...
-            }
-          >
-            {item.text}
-          </LinkButton>
-        ))}
+        {links
+          .filter(
+            ({ permission }) =>
+              typeof permission === 'undefined' ||
+              useCheckPermission(permission),
```
ojeytonwilliams, line 29:
```suggestion
            ({ requiredPermission }) =>
              !requiredPermission || useCheckPermission(requiredPermission),
```
I think this reads a little better and (particularly if we use `Permission`) we should not need to worry about the distinction between `''` and `undefined`.

client/src/modules/dashboard/shared/components/Layout.tsx
```diff
@@ -17,17 +22,23 @@ export const Layout = ({ children }: { children: React.ReactNode }) => {
...
+          )
+          .map((item) => (
+            <LinkButton
+              key={item.link}
+              href={item.link}
+              colorScheme={
+                router.pathname.startsWith(item.link) ? 'blue' : undefined
```
ojeytonwilliams, line 36:
```suggestion
                router.pathname.startsWith(item.link) ? 'blue' : 'gray'
```
Since we're here, `gray` is more declarative than `undefined`.

ojeytonwilliams (changes requested):
Looking good, thanks @gikf. Just some minor nitpicks:

gikf:
Thanks @ojeytonwilliams, good ideas all around :rocket:.

ojeytonwilliams (approved):
Cheers, @gikf. We're getting there :rocket:

PR #1190, opened 2022-07-07
refactor: move permissions to common module

ojeytonwilliams:

@gikf no great rush on this one.  I'd be happy to merge https://github.com/freeCodeCamp/chapter/pull/1187 first and then fix the conflicts in this PR.


ojeytonwilliams:
Now this is working, is it a good idea?

Other options:

1. Similarly to generated.tsx, we could copy the permissions module across from the server
2. Put permissions in the client and import them into the server

Neither seem great, to be honest, so I'm leaning towards using the experimental feature.

gikf (approved):
Looks good to me :+1: 

I don't have any comments whether this is best option or not. It seems to be currently the most convenient for usage, with not mixing client/server and clearly keeping things in a single place.

ojeytonwilliams:
Cool.  I'm pretty confident that any problems this causes won't be subtle, so I think we're good.

PR #1203, opened 2022-07-11
feat: add pagination to event-feed page

### LINKED ISSUE #788: Event feed should contain a paginated list of events
/events should have the upcoming events feed.  This should be the same as the landing page, but more events should be visible by default.

In order to show the pagination, the db generator will need to generate more events.

shootermv:

Closes #788


Sboonny (commented):
This is good for clearing the issue, I need to polish the font-size and the animation being weird. But that isn't what the original issue about, so thank you for the hard work.

@allella I wonder what are your thoughts about this change?

allella:
@Sboonny , I defer to @ojeytonwilliams , but for an MVP it looks good to me.

The database seeding may only produce events in the future. I changed the date on an existing event to be in the past and it correctly  disappears from /events.  Though, that same past event appears on the homepage, so we may need to bring the homepage into alignment with the /events feed.

I wonder if we'd want the same pagination on the homepage. It currently has a button to load more.  If we're going to keep the button style on the homepage, then I'd try to avoid the word "Click" and change the button to something like "Show More Events".

ojeytonwilliams (changes requested):
I don't have a strong preference for either UI, to be honest.  The 'click', however, is definitely redundant since it's a button and that's implicit.

As for the pagination itself, I think we should be using the `paginatedEvents` resolver, rather than fetching all events and splitting them on the client. It's not a big deal when we only have a small number of events, but since we already have the resolver it would be better to use it.

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -51,41 +51,49 @@ export default function Pagination({
...
-  const { loading, error, data } = useMinEventsQuery();
   const [currentPage, setCurrentPage] = useState(1);
-  const pageSize = 5;
-  const paginated = useMemo(() => {
-    const offset = (currentPage - 1) * pageSize;
-    return data?.events.slice(offset, offset + pageSize); // 0 , 4 // 5, 10
-  }, [currentPage, pageSize, data?.events]);
+  const { loading, error, data, fetchMore } = useHomeQuery({
+    variables: { offset: (currentPage - 1) * pageSize, limit: pageSize },
+  });
```
ojeytonwilliams, line 59: `useHomeQuery` works, but it's doing too much: it's getting the chapters, too.  Instead, it would be good to use a dedicated hook so that we only request what we need. Namely, the events.

Since we already have the `useMinEventsQuery` hook, we can modify that to return paginated events and we're good.

The process is a little involved, but what you can do is start with

https://github.com/freeCodeCamp/chapter/blob/84c2f6152d36d4a6dcc7ada94a838b4161073a1a/client/src/modules/events/graphql/queries.ts#L3-L4

(which should be called `eventFeed`, rather than `minEvents`) and modify it to use the `paginatedEvents` resolver.  Once that's changed, `graphql.tsx` should regenerate (if not, `npm run gen` will do that).  That will create a new hook you can use in `eventsPage` to get the paginated data.

ojeytonwilliams (changes requested):

client/src/modules/events/pages/eventsPage.tsx
shootermv: created  usePaginatedEventsQuery,
thing that currently bother me - i still need to use `minEventsQuery` for getting total events number

ojeytonwilliams: Yeah, I see the problem.  

What you can do is use `prisma.events.count()` in the `paginatedEvents` resolver to get the total and then return an ObjectType including the events and their count.  There's a few ways to do this, depending on the shape of the data you want on the client.  A single object containing an array of events and a single count seems reasonable to me, but whatever works.

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -1,17 +1,76 @@
...
+        limit: pageSize,
+      },
+    });
+  }, [currentPage]);
   if (loading) {
     return <h1>Loading...</h1>;
   }
 
-  if (error || !data?.events) {
+  if (error || !data?.paginatedEventsWithTotal?.events?.length) {
```
ojeytonwilliams, line 73:
```suggestion
  if (error || !data?.paginatedEventsWithTotal.events.length) {
```
Once we know `data` exists, the rest have to be defined.

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -24,9 +83,15 @@ export const EventsPage: NextPage = () => {
     <VStack>
       <Stack w={['90%', '90%', '60%']} maxW="600px" spacing={6} mt={10} mb={5}>
         <Heading>Events: </Heading>
-        {data.events.map((event) => (
+        {data?.paginatedEventsWithTotal?.events.map((event) => (
```
ojeytonwilliams, line 86:
```suggestion
        {data.paginatedEventsWithTotal.events.map((event) => (
```
Same logic.

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -24,9 +83,15 @@ export const EventsPage: NextPage = () => {
...
           <EventCard key={event.id} event={event} />
         ))}
+        <Pagination
+          currentPage={currentPage}
+          setCurrentPage={setCurrentPage}
+          pageSize={pageSize}
+          records={data?.paginatedEventsWithTotal?.total || 0}
```
ojeytonwilliams, line 93:
```suggestion
          records={data.paginatedEventsWithTotal.total}
```
Last one.

client/src/modules/events/graphql/queries.ts
```diff
@@ -1,4 +1,55 @@
...
+        }
+      }
+      chapter {
+        id
+        name
+        category
+      }
+    }
+  }
+`;
```
ojeytonwilliams, line 52: This does not appear to be used. Can we delete it?

ojeytonwilliams (changes requested):
Nice work, @shootermv this is looking pretty cool.

I found a few minor issues.  Nitpicks, to be honest:

client/src/modules/events/graphql/queries.ts
shootermv: done

ojeytonwilliams:
Hey @shootermv, since I needed to fix conflicts, I also went ahead and regenerated the code so we can get this merged.  Thanks a ton for working on this!

ojeytonwilliams (approved):

ojeytonwilliams:
Closed and re-opened to restart tests :disappointed:

ojeytonwilliams (approved):

PR #1212, opened 2022-07-12
feat: restrict venue creation

### LINKED ISSUE #1195: Associate Venues with Chapters
Do we want each venue to belong to one chapter? There are couple of reasons I can see to do this, one practical and one technical.

The practical reason is that a venue is typically going to be close to the chapter members and so forcing event creators to choose from all possible venues is unhelpful.  There is also the potential for confusion if admins of multiple chapters update the same venue.

The technical reason is that the authorisation logic uses the target `chapterId` to determine if a user's request should be rejected.  If `chapterId` is not sent as part of the request, then chapter admins will not be able to manage venues.  There are a few potential workarounds: 

1) Give chapter admins the VenuesManage instance role
2) Handle the authorization logic in the resolver
3) Rework authorization logic

Issues:

1) is pretty awkward. In part because it's messy (users shouldn't need more than one role) and in part because instance permissions cannot be stopped by bans (yet).
2) mixes up concerns
3) is the nuclear option, since I'm not sure what we'd do instead

## Summing up

Given that there are positive reasons to want this as well as a lack of obvious workarounds, I think this is the way to go.

Any concerns before I get on with implementing this?

ojeytonwilliams:

Closes https://github.com/freeCodeCamp/chapter/issues/1195

This is a little more extensive than most auth PRs, because of the reasons mentioned in the linked issue.

To ensure that the associated chapterId was available for the request, I put venue creation inside /dashboard/chapters/[chapterId]/ much like event creation.  Similar story for editing events.

TODO:

- [x] ~~Move event edit page inside /dashboard/chapters/[chapterId]/~~ (necessary, but nothing to do with this) 
- [x] ~~Make sure VenueForm gets passed `chapter_id`~~ (it's fine, the `onSubmit` method is doing it)

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -3181,7 +3181,23 @@
             "description": null,
             "args": [
               {
-                "name": "id",
+                "name": "chapterId",
+                "description": null,
+                "type": {
+                  "kind": "NON_NULL",
+                  "name": null,
+                  "ofType": {
+                    "kind": "SCALAR",
+                    "name": "Int",
+                    "ofType": null
+                  }
+                },
+                "defaultValue": null,
+                "isDeprecated": false,
+                "deprecationReason": null
+              },
+              {
+                "name": "venueId",
                 "description": null,
                 "type": {
                   "kind": "NON_NULL",
@@ -3201,8 +3217,8 @@
               "kind": "NON_NULL",
               "name": null,
               "ofType": {
-                "kind": "SCALAR",
-                "name": "Boolean",
+                "kind": "OBJECT",
+                "name": "Venue",
                 "ofType": null
               }
             },
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -5405,6 +5405,22 @@
         "description": null,
         "fields": [
           {
+            "name": "chapter_id",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Int",
+                "ofType": null
+              }
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
             "name": "city",
             "description": null,
             "args": [],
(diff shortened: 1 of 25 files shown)


gikf (commented):
How much of the chapter - venue relation (venues are per chapter, one venue can be used only in one chapter), is expected in this PR? Venues are referencing single chapter, but the constrain isn't really enforced in any way.

Suggestions are for two unrelated, minor, things spotted.

ojeytonwilliams:
> How much of the chapter - venue relation (venues are per chapter, one venue can be used only in one chapter), is expected in this PR? Venues are referencing single chapter, but the constrain isn't really enforced in any way.

I'm not entirely happy with how I've done this, so I'd welcome your thoughts.

The idea is this: giving _venues_ a foreign key referencing _chapters_ enforces the requirement that each venue is associated with a single chapter.  

Making the [id, chapter_id] pair unique was a hack to make use of the `chapterId` variable.  It's too hacky, though, since id alone is already unique.  It seems fine to send `chapterId` with the query, so that the authChecker accepts the request, and then just not use the variable inside the resolver.  That's still kind of a hack, but less so.

One thing I was wondering is how else do think this should be enforced?  Do we need further UI changes or did you have something else in mind?

gikf:
What I meant that it isn't enforced was that there were events in other chapters seeded with venues from chapter `1`, or that in event creation form all venues can be been picked (and used).

I thought about adding to the `venue` field in `events` table another relation - to the `chapter_id` https://github.com/ojeytonwilliams/chapter/blob/2051af4ee3d0b89253afadad29b209b593d4a14b/server/prisma/schema.prisma#L113-L120
In my mind this takes care of setting constraint in the db. The `@@unique([id, chapter_id])` still seems to be needed though, otherwise prisma complains.

ojeytonwilliams:
Got you.  I was trying to do as little as possible in this PR, but thinking a little ahead would have been good... The approach you're suggesting sounds great; my hack stops being a hack if it's required by Prisma!

How about this approach: revert the last commit, merge this (assuming you're otherwise happy) and then constrain the db to ensure that the event's venue is from the right chapter?

gikf (approved):
Sounds good to me :+1:

PR #1237, opened 2022-07-16
feat(client): delete chapter

### LINKED ISSUE #1216: Chapter Deletion functionality
We have the resolver, but we don't have a UI to make requests to it.

Since this is the most destructive operation we have, it should only be accessible to instance owners, should come with dire warnings and be difficult to do accidentally.

Something like the Github repo deletion UI would be nice: i.e. require owners to type in the chapter name before the delete button is enabled.

Sboonny:

Ref #1216


client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
```diff
@@ -1,9 +1,14 @@
...
+import { Tabs, TabList, TabPanels, Tab, TabPanel } from '@chakra-ui/tabs';
+
 import { LinkButton } from 'chakra-next-link';
 import { NextPage } from 'next';
 import { useRouter } from 'next/router';
-import React from 'react';
+import React, { useRef } from 'react';
+import { Button } from '@chakra-ui/react';
 import { Card } from '../../../../components/Card';
+// uncomment the dialog tomorrow, I am going auto pilot without thinking
```
Sboonny, line 10:
```suggestion
```

client/src/components/SettingAlert.tsx
```diff
@@ -0,0 +1,69 @@
...
+      >
+        <AlertDialogOverlay>
+          <AlertDialogContent>
+            <AlertDialogHeader fontSize="lg" fontWeight="bold">
+              {title}
+            </AlertDialogHeader>
+
+            <AlertDialogBody>
+              {DialogBody}
+              {/* Make sure that the Chapter name people see, is the same of what they are expected to input */}
```
Sboonny, line 48:
```suggestion
```

client/src/modules/auth/pages/Confirm.tsx
```diff
@@ -0,0 +1,21 @@
+import { NextPage } from 'next';
+import React from 'react';
+import { Box, VStack, Heading } from '@chakra-ui/react';
+
+// ToDo create confirm mutation
+// import { useLoginMutation } from '../../../generated/graphql';
```
Sboonny, line 6:
```suggestion
```

client/src/components/SettingAlert.tsx
```diff
@@ -0,0 +1,69 @@
...
+  AlertDialogHeader,
+  AlertDialogContent,
+  AlertDialogOverlay,
+  Button,
+  useDisclosure,
+  Input,
+} from '@chakra-ui/react';
+import React from 'react';
+
+// Make sure this has no animation because people who use it know what it's so there is no need for attention driven animation.
```
Sboonny, line 14:
```suggestion
```

Sboonny (commented):
Remove personal comments

Sboonny:
I just thought of something, with Google auth will there be passwords? or we have to send verification email with code in it

client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
```diff
@@ -1,21 +1,37 @@
...
+
+  const [deleteChapter] = useDeleteChapterMutation();
 
   const { loading, error, data } = useChapterQuery({
     variables: { chapterId },
   });
 
+  const clickDelete = () => {
+    deleteChapter({ variables: { chapterId } });
+    window.location.href = '/dashboard/chapters';
```
Sboonny, line 32: Is there a better way to do this with react?

Sboonny:
I don't know what I can do more. Maybe remove the background and hover effect because they are unnecessary, but this is ready for review. 

Because of the nature of this change, I won't merge until I get an approval.

client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
gikf: See https://nextjs.org/docs/api-reference/next/router#routerpush

Sboonny: Thank you for the link, I couldn't follow the steps in https://nextjs.org/docs/api-reference/next/router#resetting-state-after-navigation well. 

And I don't want to use `useEffect` which I am not comfortable with yet, I am delaying this to post MVP. 

Here is my attempt, if I did something stupid, and it can work with small change. Then please guide me https://github.com/freeCodeCamp/chapter/pull/1237/commits/50f523de8ace70689e87a0e7175417ac5eec71e7

gikf: The update can be done by passing to the `useDeleteChapterMutation()`, the `refetchQueries` argument with query that should be updated  
```typescript
import { CHAPTERS } from '../../../chapters/graphql/queries';
/* ... */
    const [deleteChapter] = useDeleteChapterMutation({
    refetchQueries: [{ query: CHAPTERS }],
  });
```

gikf (changes requested):
Chapter gets deleted even when you don't input its name in the deletion modal.

`chakra-confirm` module that's used in few places in the client has `useConfirmDelete` hook with optional `doubleConfirm` option. Was that considered to be used?

I'd also try to limit the changes in the PR to deleting chapters, and adding tabs or events list do in another PR, as it's getting a bit noisy in here. This requires splitting focus to three areas really.

Sboonny:
I didn't think, it was needed. But looking at how it's function, I am going to implement it

--- 

I don't know what to replace the `tabs` with, and it's doing its functionality, I will look how the dashboard is functioning, and I will create something similar, if it takes awhile, let's move it to post MVP and keep the taps.

About the events, I will remove them, I don't see any reason to have them in this pull request, and it isn't like everywhere has them and the page don't 👍

Sboonny:
I have added Chakra-confirm hook, but it's still not working as intended. I have looked over for the documentation and found nothing, but I found this https://github.com/Zeko369/chakra-confirm/blob/master/src/context.tsx. Is this what we are using?

gikf:
I mean, tabs might be a good way to separate some chapter operations. But lets do them with separate PR, (I'm also not entirely sure some options needs to be hidden, after all this is the administrative dashboard). For PR adding delete chapter functionality, just another button along the existing ones should be enough.

That's the `chakra-confirm` package, yes. Enabling the double confirm in it would be:
```
    const ok = await confirmDelete({ doubleConfirm: true });
```

ojeytonwilliams:
@Sboonny could you take a look at the conflicts?

client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
```diff
@@ -18,13 +18,15 @@ import { Layout } from '../../shared/components/Layout';
 import { CHAPTERS } from '../../../chapters/graphql/queries';
 
 export const ChapterPage: NextPage = () => {
-  const { param: chapterId } = useParam('id');
+
+  const { param: chapterId, isReady } = useParam('id');
```
Sboonny, line 22:
```suggestion
  const { param: chapterId, isReady } = useParam('id');

```

client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
```diff
@@ -18,13 +18,15 @@ import { Layout } from '../../shared/components/Layout';
...
   const confirmDelete = useConfirmDelete();
 
   const [deleteChapter] = useDeleteChapterMutation({
     refetchQueries: [{ query: CHAPTERS }],
   });
 
+
```
Sboonny, line 29:
```suggestion
```

Sboonny (commented):
revert to the old format

ojeytonwilliams (approved):
LGTM :+1: 

The double-confirm UI should be fine for now.  

Also, I've edited the post so this doesn't close the issue. I'll add some tests and make sure only instance owners can do this and then we can close the issue.

@gikf do you still have any concerns?

gikf:
@ojeytonwilliams I'm mostly good. However it occurred to me now, that more refetching might be needed. Events are being removed with chapter, so their queries (`DATA_PAGINATED_EVENTS_TOTAL_QUERY` and `HOME_PAGE_QUERY`) also can be affected.

Sboonny:
I have updated the query to 

```ts
  const [deleteChapter] = useDeleteChapterMutation({
    refetchQueries: [
      { query: CHAPTERS },
      { query: DATA_PAGINATED_EVENTS_TOTAL_QUERY, variables: { chapterId } },
      { query: HOME_PAGE_QUERY, variables: { chapterId } },
    ],
  });
```

ojeytonwilliams:
@Sboonny why, though? Those queries fetch lists of chapters.

Sboonny:
@ojeytonwilliams oh I thought that would fix Event dashboard not updating, after chapter is deleting but refreshing the dashboard show Events are deleted, I just need to refetch `EVENT` and `VENUE` data, so people won't need to refresh the page after deleting function

ojeytonwilliams:
What's happening here is the Apollo client's cache needs updating, otherwise the pages will still render the deleted chapter.  In order to do that, we have two options 1) update the cache directly and 2) refetch any queries that will return different results.  For simplicity, we're going with option 2) see https://github.com/freeCodeCamp/chapter/pull/866 for the last discussion about this.

So, to get option 2) to work we need to refetch all the queries that might have changed.  In practice, for this, that means *most* of the queries that include chapters, events and venues (since all three have been modified).

As far as I can see that means we want

```
    refetchQueries: [
      { query: CHAPTERS },
      { query: EVENTS },
      { query: HOME_PAGE_QUERY, variables: { offset: 0, limit: 2 } },
      {
        query: DATA_PAGINATED_EVENTS_TOTAL_QUERY,
        variables: { offset: 0, limit: 5 },
      },
      { query: VENUES },
    ],
```

client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
```diff
@@ -1,21 +1,57 @@
...
-import { useChapterQuery } from '../../../../generated/graphql';
+import {
+  useChapterQuery,
+  useDeleteChapterMutation,
+} from '../../../../generated/graphql';
 import { useParam } from '../../../../hooks/useParam';
 import styles from '../../../../styles/Page.module.css';
 import { Layout } from '../../shared/components/Layout';
+import { CHAPTERS } from '../../../chapters/graphql/queries';
+import { VENUE } from '../../Venues/graphql/queries';
```
ojeytonwilliams, line 19:
```suggestion
import { VENUES } from '../../Venues/graphql/queries';
```
The `VENUE` query just fetches a single venue, but we need to update the collection of venues.

client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
```diff
@@ -1,21 +1,57 @@
...
+  const [deleteChapter] = useDeleteChapterMutation({
+    refetchQueries: [
+      { query: CHAPTERS },
+      { query: EVENTS },
+      { query: HOME_PAGE_QUERY, variables: { offset: 0, limit: 2 } },
+      {
+        query: DATA_PAGINATED_EVENTS_TOTAL_QUERY,
+        variables: { offset: 0, limit: 5 },
+      },
+      { query: VENUE },
```
ojeytonwilliams, line 38:
```suggestion
      { query: VENUES },
```

ojeytonwilliams (changes requested):

ojeytonwilliams (approved):
LGTM :+1: 

Nice work, @Sboonny :rocket:

ojeytonwilliams:
@gikf couldn't spot any pages that weren't updating on chapter delete.  Could you merge this if it seems okay?

gikf (approved):

PR #1251, opened 2022-07-18
refactor: rename specs for readability

ojeytonwilliams:
Cypress only shows the filename in the logs, so this disambiguates them
a little.

This is a little silly, but I spend a lot of time looking at failing specs trying to figure out which index.cy is failing :man_shrugging:


Sboonny (approved):

PR #1257, opened 2022-07-19
refactor: rename secrets to disabiguate them

ojeytonwilliams:
It's crucially important that the AUTH_JWT_SECRET is not used to create
tokens that can be accidentally shared, so this name should make it
obvious when the secret is being used in the wrong context.

- [] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md).
- [] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [] My pull request targets the `main` branch of Chapter.

Closes #XXXXX

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -18,9 +18,9 @@ EMAIL_HOST=localhost
 
 # change this variable to at least 32 random characters
 # to enforce 256-bit cryto security on Auth0's JsonWebToken
-JWT_SECRET=SetThisTo32orMoreRandomCharacters
+AUTH_JWT_SECRET=SetThisTo32orMoreRandomCharacters
 
-UNSUBSCRIBE_SECRET=SetThisToDifferent32orMoreCharacters
+UNSUBSCRIBE_JWT_SECRET=SetThisToDifferent32orMoreCharacters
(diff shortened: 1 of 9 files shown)


allella:
@ojeytonwilliams are there any easy tools in GitHub to flag PRs with certain words with a warning / reminder message so we're less likely to forget this in the future?

Also, I don't think it would hurt to add another line of comments saying to not use that for anything other than Auth0.

ojeytonwilliams:
> are there any easy tools in GitHub to flag PRs with certain words with a warning / reminder message so we're less likely to forget this in the future?

I don't know of any. I suspect it's just something we have to catch as humans.

> Also, I don't think it would hurt to add another line of comments saying to not use that for anything other than Auth0.

Actually, since I'm in the process of setting Auth0 up, and that will definitely change the secrets we need, let's hold off on merging until that's done. An extra warning in .env.example is a solid plan, though.

ojeytonwilliams:
Closing in favour of https://github.com/freeCodeCamp/chapter/pull/1516

PR #1264, opened 2022-07-19
feat: augment chapter - venue relation

### LINKED ISSUE #1195: Associate Venues with Chapters

gikf:

Closes #1195 (unless I forgot about something)

---
- Adds db constraint to ensure event venue is from appropriate chapter.
- Venue seeding seeds separate venues for each chapter.
- When adding new event, available venues are only from chapter to which event is being added.
- Displays chapter on venues dashboard, venue dashboard and add/edit venue page.
- Chapter is optionally required in the `Venue` graphql-type to not require it returned in every mutation/query involving venues.


ojeytonwilliams (approved):
LGTM. Cheers, @gikf, thanks a ton for finishing the feature

PR #1291, opened 2022-07-26
fix: infer chapterId from events, not user_events

ojeytonwilliams:

The authChecker would incorrectly reject requests from chapter administrators when the requests only have `eventId` as the variable. This changes the logic from looking for the `chapter_id` in the user's _user_events_ record (which may not exist) to looking in the _events_ table (which will).

Credit to @shootermv for letting me know about this issue.

### DIFF
--- a/server/src/authorization/index.ts
+++ b/server/src/authorization/index.ts
@@ -2,7 +2,7 @@ import { GraphQLResolveInfo } from 'graphql';
 import { AuthChecker } from 'type-graphql';
 
 import { GQLCtx } from '../common-types/gql';
-import type { User } from '../controllers/Auth/middleware';
+import type { Events, User } from '../controllers/Auth/middleware';
 
 // This is a *very* broad type, but unfortunately variableValues is only
 // constrained to be a Record<string, any>, basically.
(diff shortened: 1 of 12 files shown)


server/src/common-types/gql.ts
```diff
@@ -1,13 +1,15 @@
 import { Request as ExpressRequest, Response } from 'express';
 
-import type { User } from '../controllers/Auth/middleware';
+import type { User, Events } from '../controllers/Auth/middleware';
 
 export interface GQLCtx {
```
Sboonny, line 5: Can we have a different name than `GQLCtx` 🥺

ojeytonwilliams: ResolverCtx?

Sboonny: Sure, 👍 it took me a minute to understand it's related to graphql. 

I like the new name more💚

Sboonny (approved):

server/src/authorization/index.ts
```diff
@@ -80,6 +84,8 @@ function isAllowedByEventRole(
 }
 
 function getUserPermissionsForInstance(user: User): string[] {
+  console.log('checking instance permissions');
+  console.log('user', user);
```
shootermv, line 87: may be better to remove this console.log, if it not necessary

ojeytonwilliams: Thanks, they were just debugging tools, so I should have removed them.

PR #1295, opened 2022-07-27
feat: name users after Harry Potter characters

### LINKED ISSUE #1294: better to have example users named after Harry Potter characters
better to have example users (we currently have) named after Harry Potter characters -
Because that way it will be more make sense - 
If there is Dumbledore is who created event  - everybody will want to register (unlike now that is some name nobody cares of)

shootermv:

Closes #1294


Sboonny (approved):
🤣🤣

ojeytonwilliams (changes requested):
While having fixed names for the special users seems good, we should not make use of any particular IP.  In part because there may be legal implications, but mostly because it is distracting.

PR #1302, opened 2022-07-28
refactor: split rsvpEvent resolver

gikf:

---
- Splits rsvping into rsvpEvent and cancelRsvp mutations.
- Removes `initUserInterestForChapter` mutation, in favor of using slightly modified `joinChapter` mutation.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -2735,6 +2735,35 @@
             "deprecationReason": null
           },
           {
+            "name": "cancelRsvp",
+            "description": null,
+            "args": [
+              {
+                "name": "eventId",
+                "description": null,
+                "type": {
+                  "kind": "NON_NULL",
+                  "name": null,
+                  "ofType": {
+                    "kind": "SCALAR",
+                    "name": "Int",
+                    "ofType": null
+                  }
+                },
+                "defaultValue": null,
+                "isDeprecated": false,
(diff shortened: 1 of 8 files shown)


gikf:
@ojeytonwilliams There's one detail that I've forgotten. Without user having `chapter_user` entry before trying to rsvp, they are not going to be authorized to rsvp (unless they are instance owner). I still think it's worth to split canceling rsvp to simplify logic, but removing separate `chapter_user` creation is not going to be that easy.

Any ideas?

ojeytonwilliams:
@gikf that's true - I'd not thought about the lack of permissions.  To be honest, it's probably smarter to keep joining a chapter separate from rsvping.  The only reasons to combine them are a) new users will need to join b) one less round trip to the server.  a) isn't that compelling since we have to over-complicate a resolver to do it and b) may not be necessary in practice.

So, on reflection, I think combining `initUserInterestForChapter` with rsvping is a mistake.

However, while we're at it: do we need both `initUserInterestForChapter` and `joinChapter`? As far as I can see, `initUserInterestForChapter` is there because it can be supplied `eventId`, but we know what the `chapterId` is.

ojeytonwilliams (approved):
Nice work, @gikf it's always nice to see the code get clearer and the number of lines go down for a change!

PR #1305, opened 2022-07-29
feat: seed chapter 2 admin

### LINKED ISSUE #1279: Create an admin of chapter 2 for testing
We have multiple different possible roles whose permissions we want to test.  However, the tests currently only have one admin, so have to approach testing indirectly (changing request variables, rather than users).

To complement admin@of.a.chapter, another user admin@of.chapter.two would make for concise, readable tests. c.f. https://github.com/freeCodeCamp/chapter/blob/2d8cafcf74cf892166986ef4ae350bd45e052284/cypress/e2e/dashboard/venues.cy.ts#L131-L144

which uses admin@of.a.chapter, but changes the request to act on chapter 2.

It's not _terrible_, but it's a minor pain point.

gikf:

Closes #1279

---
- Seeds admin for chapter 2.
- Uses chapter 2 admin in test.


server/prisma/generator/factories/user.factory.ts
```diff
@@ -21,13 +22,21 @@ const createUsers = async (
   };
   const owner = await prisma.users.create({ data: ownerData });
 
-  const adminData: Prisma.usersCreateInput = {
+  const chapter1AdminData: Prisma.usersCreateInput = {
     email: 'admin@of.a.chapter',
```
ojeytonwilliams, line 26: Since we have `admin@of.chapter.two` how about renaming this to `admin@of.chapter.one`? I think that's a little nicer.

gikf: That's fine with me!

ojeytonwilliams (approved):
Looks good. Thanks @gikf!

PR #1315, opened 2022-07-31
docs: add server side authorization documentation

gikf:

---

- Adds role/permission based authorization documentation.
- As supposedly tables and flowcharts can be easier for comprehension there are both here! :P
- Any comments regarding wording, what might be missing/not needed, or positioning of the section are appreciated.


CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
 
 [This](https://github.com/freeCodeCamp/chapter/contribute) is a good place to go if you are looking to get started.
 
+## Authorizing GraphGQL requests with `@Authorized` decorator
```
Sboonny, line 434: I think it's better to be less specific, so it's more comfortable and easier for our future self to broaden it when necessary.

```suggestion
## Authorized GraphQL Requests
```

CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
...
+
+### Overview or roles
```
Sboonny, line 436:
```suggestion
### Current Roles
```

CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
...
+
+Chapter uses role based permission system, with specific permission assigned to one or multiple roles, on one or more different _levels/scopes_).
```
Sboonny, line 438:
```suggestion
Chapter has a role based permission system, with specific permissions and different **scopes** assigned to each role.
```

CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
...
+
+* Instance role - per instance - every registered user has instance role.
+* Chapter role - per chapter - chapter role is created for user when explicitly joining chapter, or performing on event actions requiring chapter role. Ie. Rsvping to event. Elevated chapter role for one chapter, doesn't impact chapter roles for different chapters.
+* Event role - per event - chapter role is created for user when Rsvping to an event. Elevated event role (permissions) for a one event doesn't impact event roles for different events.
+
+_Note: in MVP Event roles are not used._
+
+### `@Authorized` decorator
```
Sboonny, line 446:
```suggestion
### `@Authorized` Decorator Usage in Chapter
```

CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
...
+
+Decorator is available as an import from `type-graphql`.
+
+```ts
+import { Authorized } from 'type-graphql';
+```
+
+Adding it in resolver before specific request logic, wraps the request in the authorization logic. Only for authorized users logic of the request will be executed.
```
Sboonny, line 454:
```suggestion
Adding it in the resolver before specific request logic, will wrap the request in the authorization logic, which will execute only authorized users' logic of the request.
```

CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
...
+
+By default this will mean only logged in user can be authorized.
```
Sboonny, line 456:
```suggestion
By default, this means only logged in users can be authorized.
```

CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
...
+  async createEvent(
+    @Arg('chapterId', () => Int) chapterId: number,
+    @Arg('data') data: CreateEventInputs,
+    @Ctx() ctx: Required<ResolverCtx>,
+  ): Promise<Event | null> {
+    /* ... */
+  }
+```
+
+To allow authorization logic connecting user roles on specific chapter or event, with their permissions for performing request, certain arguments are required in the request.
```
Sboonny, line 470:
```suggestion
To allow authorization logic connecting users' roles on a specific chapter or event, with their permissions for performing the request, certain arguments are required in the request.
```

CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
...
+    CP-->|Banned| N;
+    C-->|No| E;
+    C-->|Yes| Y;
+    E-->|Yes| Y;
+    E-->|No| N;
+```
+
+### Example
+
+Adding `@Authorize(Permissions.EventEdit)` before event editing logic will authorize user if one of the following applies:
```
Sboonny, line 503:
```suggestion
Adding `@Authorize(Permissions.EventEdit)` before an event -which edits the logic- will authorize the user if one of the following applies:
```

CONTRIBUTING.md
```diff
@@ -431,6 +431,83 @@ After you have added new feature, to make sure it stays working, we recommend us
...
+
+* Their Instance role has `Permissions.EventEdit` permission.
+* User is not banned on the chapter, on which edited event is, and for that chapter their Chapter role has `Permission.EventEdit` permission.
+* Their Event role, for the edited event, has `Permission.EventEdit` permission.
+
+Otherwise request will be rejected.
```
Sboonny, line 509:
```suggestion
Otherwise, the request will be rejected.
```

Sboonny (commented):
Here is how it looks https://github.com/gikf/chapter/blob/docs/role-permission-based-authorization/CONTRIBUTING.md#authorizing-graphgql-requests-with-authorized-decorator 

We can move it to here https://github.com/gikf/chapter/blob/docs/role-permission-based-authorization/CONTRIBUTING.md#api-specification

About wording, here are suggestions

allella:
Thanks. It looks like a lot of work went into it, but I can't comment on the technical parts.

I do think it would be good to put the bulk of it in an expandable section, like we have with the "Steps" throughout the docs.

Also, adding a heading and in-page link to the section in the Table of Contents at the top of the document would be good once you determine where it fits in the hierarchy.

CONTRIBUTING.md
gikf: The last part of the suggestion:
> which will execute only authorized users' logic of the request.

Changes the sense of the original sentence. The point is that when user isn't authorized, their request will not reach the actual code intended for handling that request (when user is authorized), but will be stopped earlier, at the authorization step.

CONTRIBUTING.md
gikf: Could you take another look here? Suggestion is also altering the sense of original sentence.

Sboonny: Sure, that what I originally thought was going one 👍. Hopefully I will think of better suggestion

CONTRIBUTING.md
Sboonny: What do you think of this?

```suggestion
Structuring your authorization logic, by adding a resolver before the specific request, will wrap that request. With this, the request will execute that logic only for authorized users
```

CONTRIBUTING.md
Sboonny: I had an issue with `event editing logic` because it feels like it's messing something. But that maybe me overthinking, so feel free to discard these suggestions if it's harmful to the overall document

gikf:
I've moved it to server section. Not being expandable doesn't make it standout here as much, but expandable sections still could be added if needed. I've put it in the TOC and updated as well some wording.

Sboonny (approved):

ojeytonwilliams (approved):
I'm sorry for letting this sit here for so long, it looks great.

By the way, that flowchart is amazing!  Did you generate it with a tool or just learn the syntax?

gikf:
@ojeytonwilliams :rocket: done by hand, with some documentation/examples and live preview.

ojeytonwilliams:
Good to know, thanks. I'll have to try that myself,

Issue #1217, opened 2022-07-13
Resolvers requiring Authorization decorators


ojeytonwilliams:
To complete the role-based authorization the Authorized decorator needs to be applied to the following resolvers:

- [x] #1298 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/Events/resolver.ts#L480
- [x] #1332 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/Events/resolver.ts#L616 (chapter admin)
- [x] #1323 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/Events/resolver.ts#L651 (chapter admin)
- [x] #1346 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/Events/resolver.ts#L661
- [x] #1280 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/ChapterUser/resolver.ts#L45 (instance member)
- [x] #1280 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/ChapterUser/resolver.ts#L73 (instance member)
- [x] #1284 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/ChapterUser/resolver.ts#L183 (instance owner)
- [x] #1413 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/ChapterUser/resolver.ts#L212 (chapter admin)
- [x] #1416 https://github.com/freeCodeCamp/chapter/blob/65beab1cbb57e2d5d5eaca80fd3b8d997c6743c2/server/src/controllers/ChapterUser/resolver.ts#L209 (chapter admin)
- [x] #1416 https://github.com/freeCodeCamp/chapter/blob/bee4c91ce972bb50e3fb3185b33b5e6157a8903f/server/src/controllers/ChapterUser/resolver.ts#L257 (chapter admin)
[...]

PR #1332, opened 2022-08-04
feat: add authorized decorator to cancelEvent endpoint

shootermv:

Closes part of #1217


ojeytonwilliams (approved):
I confirmed that the cypress/e2e/dashboard/events/events-index.cy.ts spec fails if the user doesn't have permission, so I think this is tested enough.

So, thanks @shootermv, this LGTM :+1:

Issue #1339, opened 2022-08-05
Add jest snapshot for chapterCard component


Sboonny:
ChapterCard doesn't have snapshot, it would be lovely to add one

- [ ] add jest snapshot for it 

----

<s>
As discussed in #323, the current display of chapter info in the homepage isn't helpful, for that we can do the following

-  add tags to chapter #1344 
- [x] tags should be meaningful #1341
- [x] change how chapter's name is displayed to make it easier to understand #1361 
- [x] display chapter's type #1361 
- [x] display chapter's description #1361 
- [x] remake the presentation #1361 

----

not MVP

- add icons to make them clearer

</s>

PR #1341, opened 2022-08-05
chore: seed less random tags

Sboonny:

clear task in #1339

I am planning to use event tags for chapter tags, but chapter should have three tags or maybe different tags? idk after the mvp

### DIFF
--- a/server/prisma/generator/factories/events.factory.ts
+++ b/server/prisma/generator/factories/events.factory.ts
@@ -84,19 +84,27 @@ const createEvents = async (
       }),
     );
 
-    const tagNames = Array.from(new Array(1 + random(3)), () => lorem.word());
+    const tagNames = [
+      'Front End Development',
+      'Back End Development',
+      'FullStack Development',
+      'Cat Adoring',
+      'Dog Adoring',
+      'SNAKES',
+    ];
+    const randomTag = Math.floor(Math.random() * tagNames.length);
     await prisma.events.update({
       where: { id: event.id },
       data: {
         tags: {
-          create: [...new Set(tagNames)].map((tagName) => ({
+          create: {
             tag: {
               connectOrCreate: {
-                create: { name: tagName },
-                where: { name: tagName },
+                create: { name: tagNames[randomTag] },
+                where: { name: tagNames[randomTag] },
               },
             },
-          })),
+          },
         },
       },
     });
--- a/server/prisma/generator/factories/events.factory.ts
+++ b/server/prisma/generator/factories/events.factory.ts
@@ -107,9 +107,9 @@ const createEvents = async (
       'Ruby',
       'Rust',
     ];
-    const randomNum = Math.round((Math.random() * tagNames.length) / 5);
+    const tagsAmount = Math.round((Math.random() * tagNames.length) / 5);
 
-    const tags = randomItems(tagNames, randomNum, true);
+    const tags = randomItems(tagNames, tagsAmount, true);
     const connectOrCreateTags = tags.map((tag) => ({
       tag: {
         connectOrCreate: {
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (commented):
This looks good, but is it necessary to reduce the number of tags?

Sboonny:
at first I though I could swap `create` with `createMany` and add multiple tag.

```js
          createMany: {
            tag: {
              connectOrCreate: {
                create: { name: tagNames[randomTag] },
                where: { name: tagNames[randomTag] },
              },
            },
            tag: {
              connectOrCreate: {
                create: { name: tagNames[randomTag] },
                where: { name: tagNames[randomTag] },
              },
            },
          },
```

but that didn't work, so I pushed to create style and planning to change it later. 

If you have suggestion of what I can do, I am open for it

ojeytonwilliams:
The issue is that `createMany` doesn't [accept nested inputs ](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#createmany), so you can't use `connectOrCreate` inside of it.

You could create all the tags and then connect them to each event, but it's easier to use `connectOrCreate` inside the `prisma.events.update` call.

Something like this, if you always want 3 tags.

```ts
    const tags = randomItems(tagNames, 3, true);
    const connectOrCreateTags = tags.map((tag) => ({
      tag: {
        connectOrCreate: {
          where: { name: tag },
          create: { name: tag },
        },
      },
    }));

    await prisma.events.update({
      where: { id: event.id },
      data: {
        tags: {
          create: connectOrCreateTags,
        },
      },
    });
```

server/prisma/generator/factories/events.factory.ts
```diff
@@ -107,19 +107,23 @@ const createEvents = async (
       'Ruby',
       'Rust',
     ];
-    const randomTag = Math.floor(Math.random() * tagNames.length);
+    const randomNum = Math.round((Math.random() * tagNames.length) / 5);
+
+    const tags = randomItems(tagNames, randomNum, true);
```
Sboonny, line 112:
```suggestion
    const tagsAmount = Math.round((Math.random() * tagNames.length) / 5);

    const tags = randomItems(tagNames, tagsAmount, true);
```

server/prisma/generator/factories/events.factory.ts
```diff
@@ -84,19 +84,46 @@ const createEvents = async (
...
+
+    const tags = randomItems(tagNames, tagsAmount, true);
+    const connectOrCreateTags = tags.map((tag) => ({
+      tag: {
+        connectOrCreate: {
+          where: { name: tag },
+          create: { name: tag },
+        },
+      },
+    }));
```
ojeytonwilliams, line 120:
```suggestion
    const tagNames = randomItems(tagNames, tagsAmount, true);
    const connectOrCreateTags = tagNames.map((name) => ({
      tag: {
        connectOrCreate: {
          where: { name },
          create: { name },
        },
      },
    }));
```
I know I suggested the first way of naming it, but this seems nicer!

server/prisma/generator/factories/events.factory.ts
```diff
@@ -84,19 +84,46 @@ const createEvents = async (
...
+      'Svelete',
+      'SveleteKit',
+      'Vite',
+      'Prisma',
+      'Ruby',
+      'Rust',
+    ];
+    const tagsAmount = Math.round((Math.random() * tagNames.length) / 5);
+
+    const selectedTags = randomItems(tagNames, tagsAmount, true);
```
gikf, line 112:
```suggestion
    const tagsCount = Math.round((Math.random() * tagNames.length) / 5);

    const selectedTags = randomItems(tagNames, tagsCount, true);
```

Nitpicking

Issue #1215, opened 2022-07-13
Event and Venue creation pages


ojeytonwilliams:
After https://github.com/freeCodeCamp/chapter/pull/1212 goes in it will not be possible to create events or venues from their respective dashboards (dashboard/venues and dashboard/events).

It would be nice if we brought back that functionality, updated to reflect the new permissions system.  Specifically:

- As an instance owner, I can go to the events/venues dashboard, click on 'add new' and then be able to create an event/venue for any chapter
- As a chapter admin, I can go to to the events/venues dashboard, click on 'add new' and then be able to create an event/venue for the chapters I am an admin of

gikf:
I was exploring ways to do this, I'm having hard time with finding good way to return appropriate list of admined chapters from server. In other words - differentiating between instance owner that can have all chapters, and filtering only chapters admined as chapter admin.

Ideal would be single query, that would return appropriate chapters. There's however problem with authorizing for it.

- For instance owner some permission with `@Authorize` can be used, but this will not work for chapter admin.
- Not using `@Authorize` decorator would require some additional logic in the resolver, to figure out if user has instance owner permissions or not.

Some alternatives:
- Running two different queries? One for instance owner returning all chapters, and if that fails - user is not authorized, run another to return only admined chapters?
- Client side filtering?

All this feels convoluted and more complicated than it should be.

ojeytonwilliams:
Thanks for the detailed breakdown, it really helped.

I've got a couple of ideas for how to attack this. (Two is _probably_ the most sensible, but it's up for debate)

### One

1. give both instance owners and chapter admins the `ListChapters` instance permission.
2. have the server send, respectively, all the chapters or all the chapters that that user administrates

My logic was just this: both instance users and chapter admins need to know which chapters they control.  Any permission has to be a instance permission, because, without making a query how do they know which chapterIds to send? If they don't know those ids (and have chapter level permissions) they'll be rejected by Auth.

### Two

1. return the chapterIds in response to the `useMeQuery`

It doesn't seem like a user should need special permission to see information about themselves (since that's all this really is - we're not talking about letting you see info about _other_ users). So how about sending it as part of the user data that every user needs?

gikf:
I like both of these. I agree it might not be really needed to additionally authorize listing of the admined chapters, after all, user not admining any chapter would simply receive empty list.

This still leaves the need to include some permission logic in query listing admined chapters. With pseudocode:
```text
query list admined chapters {
[...]

PR #1345, opened 2022-08-06
feat: venue create with admined chapters dropdown

gikf:

---

- Ref #1215
- When trying to use `useMeQuery` I was getting often on the fresh load `null` returned, instead of the appropriate data returned. Hence using `useAuth`.
- There might be better place for added resolver than `server/src/controllers/User/resolver.ts`.
- Tests checking if the correct number of chapters are in the chapter dropdown are a bit far fetched. More appropriate place could be, after replacing displaying all chapters in the `/dashboard/chapters`, with only those admined, but until then...


client/src/modules/dashboard/Venues/components/VenueForm.tsx
```diff
@@ -112,24 +123,69 @@ const VenueForm: React.FC<VenueFormProps> = (props) => {
...
+    id: number;
+    name: string;
+  }
+  let adminedChapters: Chapter[] = [];
+  if (isChaptersDropdownNeeded) {
+    const { user } = useAuth();
+    adminedChapters = user?.admined_chapters ?? [];
+    if (adminedChapters.length) {
+      setValue('chapter_id', adminedChapters[0].id);
+    }
```
ojeytonwilliams, line 148: This way of using `setValue` creates a kinda weird bug.  The first selection from the dropdown is ignored.  Subsequent clicks work as expected.

As I was digging into that I wondered, do we need to use `getValue` or `setValue` at all?  If we set the default value to be `chapterId` if it exists or `adminedChapters[0].id` otherwise, it should Just Work.

server/src/controllers/User/resolver.ts
```diff
@@ -0,0 +1,44 @@
+import { FieldResolver, Resolver, Root } from 'type-graphql';
+import { Chapter, UserWithInstanceRole } from '../../graphql-types';
+import { prisma } from '../../prisma';
+
+import { Permission } from '../../../../common/permissions';
+
+const CHAPTER_ADMIN_PERMISSION = Permission.ChapterEdit;
```
ojeytonwilliams, line 7:
```suggestion
```
It doesn't seem necessary to create a variable for this.

server/src/controllers/User/resolver.ts
```diff
@@ -0,0 +1,44 @@
...
+@Resolver(() => UserWithInstanceRole)
+export class UserWithInstanceRoleResolver {
+  @FieldResolver(() => [Chapter])
+  async admined_chapters(
+    @Root() user: UserWithInstanceRole,
+  ): Promise<Chapter[]> {
+    if (
+      user.instance_role.instance_role_permissions.some(
+        ({ instance_permission }) =>
+          instance_permission.name === CHAPTER_ADMIN_PERMISSION,
```
ojeytonwilliams, line 18:
```suggestion
          instance_permission.name === Permission.ChapterEdit,
```

server/src/controllers/User/resolver.ts
```diff
@@ -0,0 +1,44 @@
...
+      where: {
+        chapter_users: {
+          some: {
+            AND: [
+              { user_id: user.id },
+              {
+                chapter_role: {
+                  chapter_role_permissions: {
+                    some: {
+                      chapter_permission: { name: CHAPTER_ADMIN_PERMISSION },
```
ojeytonwilliams, line 33:
```suggestion
                      chapter_permission: { name: Permission.ChapterEdit },
```

ojeytonwilliams (changes requested):
> When trying to use useMeQuery I was getting often on the fresh load null returned, instead of the appropriate data returned. Hence using useAuth.

That makes sense.  Looking back at our discussion, I was too vague. You're absolutely right - the only place meQuery should be used is in the auth/store.

> There might be better place for added resolver than server/src/controllers/User/resolver.ts.

Seems fine to me!

client/src/modules/dashboard/Venues/components/VenueForm.tsx
gikf: I'm having problem with the default value not getting picked up when navigating to the `/dashboard/venues/new` via UI (Dashboard -> Venues -> Add new). When accessing page directly, or after refresh, the default value is selected.

client/src/modules/dashboard/Venues/pages/EditVenuePage.tsx
```diff
@@ -29,6 +29,8 @@ export const EditVenuePage: NextPage = () => {
 
   const onSubmit = async (data: VenueFormData) => {
     setLoadingUpdate(true);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { chapter_id: _, ...updateData } = data;
```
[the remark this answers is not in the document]
gikf: I was just trying to be very explicit that it will not be used :stuck_out_tongue:

client/src/modules/dashboard/Venues/components/VenueForm.tsx
```diff
@@ -112,24 +123,65 @@ const VenueForm: React.FC<VenueFormProps> = (props) => {
...
+              <FormLabel>Chapter</FormLabel>
+              <Select
+                {...register('chapter_id' as const, {
+                  required: true,
+                  valueAsNumber: true,
+                })}
+                defaultValue={
+                  chapterId ??
+                  (adminedChapters?.length ? adminedChapters[0].id : undefined)
+                }
```
ojeytonwilliams, line 173: There are a few bits of JS boolean fun here, so I'd make the default value this:

```
defaultValue =chapterId !== -1
      ? chapterId
      : adminedChapters.length
      ? adminedChapters[0].id
      : -1;
```
Otherwise you can end up with the length being the default value or -1 (because both are truthy).

I also ran into the issue with reloads, too.  The problem seems to be that the default value is getting 'stuck' at one of the earlier versions.  I've been going through the docs and it seems like the `defaultValues` get fixed on first render, but I can't see an obvious fix, since we're always going to render multiple times

```

  useEffect(() => {
    resetField('chapter_id', { defaultValue });
  }, [adminedChapters]);
```
seems to work, if you provide it the new default value. Not pretty, but also not buggy (I think!)

gikf: Nice, thanks. `useEffect` approach seems to deal with both of these.

ojeytonwilliams (approved):
Cheers, LGTM 👍

PR #1346, opened 2022-08-07
feat: add authorized decorator to sendEventInvite

shootermv:

Closes part of #1217

### DIFF
--- a/cypress/e2e/dashboard/events/events-index.cy.ts
+++ b/cypress/e2e/dashboard/events/events-index.cy.ts
@@ -338,7 +338,14 @@ describe('events dashboard', () => {
 
   it('chapter admin should be allowed to send email to attendees', () => {
     const eventId = 1;
+    cy.login('test@user.org');
 
+    cy.sendEventInvite(eventId, ['confirmed']).then(expectToBeRejected);
+
+    // banned admin should be rejected
+    cy.login('banned@chapter.admin');
+
+    cy.sendEventInvite(eventId, ['confirmed']).then(expectToBeRejected);
     // admin of chapter 1
     cy.login('admin@of.chapter.one');
 
--- a/cypress/e2e/dashboard/events/events-index.cy.ts
+++ b/cypress/e2e/dashboard/events/events-index.cy.ts
@@ -295,23 +295,28 @@ describe('events dashboard', () => {
 
     cy.mhDeleteAll();
   });
+});
+
+describe('events endpoints', () => {
+  beforeEach(() => {
+    cy.exec('npm run db:seed');
+    cy.login('admin@of.chapter.one');
+    cy.mhDeleteAll();
+    cy.interceptGQL('events');
+  });
 
   it('chapter admin should be allowed to edit event, but nobody else', () => {
     const eventId = 1;
-    // admin of chapter 1
-    cy.login('admin@of.chapter.one');
 
     cy.updateEvent(eventId, eventData).then((response) => {
       expect(response.body.errors).not.to.exist;
     });
     // newly registered user (without a chapter_users record)
     cy.login('test@user.org');
-
     cy.updateEvent(eventId, eventData).then(expectToBeRejected);
 
     // banned admin should be rejected
     cy.login('banned@chapter.admin');
-
     cy.updateEvent(eventId, eventData).then(expectToBeRejected);
   });
 
@@ -320,17 +325,13 @@ describe('events dashboard', () => {
 
     // newly registered user (without a chapter_users record)
     cy.login('test@user.org');
-
     cy.deleteEvent(eventId).then(expectToBeRejected);
-
     // banned admin should be rejected
     cy.login('banned@chapter.admin');
-
     cy.deleteEvent(eventId).then(expectToBeRejected);
 
     // admin of chapter 1
     cy.login('admin@of.chapter.one');
-
     cy.deleteEvent(eventId).then((response) => {
       expect(response.body.errors).not.to.exist;
     });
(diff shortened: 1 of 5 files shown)


cypress/support/e2e.ts
```diff
@@ -54,6 +54,13 @@ declare global {
        */
       updateEvent(eventId: number, data): Chainable<any>;
 
+      /**
+       * Delete rsvp of user with userId for the event with eventId
+       * @param eventId Id of the event
+       * @param emailGroups groups to send email
+       */
+      sendEventInvite(eventId: number, emailGroups: [string]): Chainable<any>;
+
```
ojeytonwilliams, line 63:
```suggestion
```
Since you added the command in commands.ts, this is not needed.

cypress/support/commands.ts
```diff
@@ -673,6 +690,7 @@ declare global {
       createSponsor: typeof createSponsor;
       deleteEvent: typeof deleteEvent;
       deleteVenue: typeof deleteVenue;
+      sendEventInvite: typeof endEventInvite;
```
ojeytonwilliams, line 693:
```suggestion
      sendEventInvite: typeof sendEventInvite;
```

cypress/support/commands.ts
```diff
@@ -661,6 +661,23 @@ const getChapterRoles = () => {
...
+/**
+ * sends event invites for attendees
+ * @param eventId event id
+ * @param emailGroups Chapter id
+ */
+const sendEventInvite = (eventId: number, emailGroups: [string]) => {
+  const sendEventInviteMutation = {
+    operationName: 'sendEventInvite',
+    variables: { eventId, emailGroups },
+    query: `  mutation sendEventInvite($eventId: Int!, $emailGroups: [String!]) {
```
ojeytonwilliams, line 673:
```suggestion
    query: `mutation sendEventInvite($eventId: Int!, $emailGroups: [String!]) {
```
Nit

server/src/controllers/Events/resolver.ts
```diff
@@ -740,6 +740,7 @@ ${venueDetails}`;
 
   // TODO: This will need a real GraphQL return type (AFAIK you have to return
   // an object type)
+  @Authorized(Permission.EventEdit)
```
ojeytonwilliams, line 743:
```suggestion
  @Authorized(Permission.EventSendInvite)
```
This deserves its own permission, to be honest.

cypress/e2e/dashboard/events/events-index.cy.ts
```diff
@@ -334,4 +335,15 @@ describe('events dashboard', () => {
...
+  it('chapter admin should be allowed to send email to attendees', () => {
+    const eventId = 1;
+
+    // admin of chapter 1
+    cy.login('admin@of.chapter.one');
+
+    cy.sendEventInvite(eventId, ['confirmed']).then((response) => {
+      expect(response.body.errors).not.to.exist;
+    });
+  });
```
ojeytonwilliams, line 348: Ideally these simple api calls should be to check that users who don't have permission can't do the thing.  

Could you try changing the `cy.login` in `beforeEach` to use `admin@of.chapter.one`? Then this 'it' can be used to check that `admin@of.chapter.two` and `banned@chapter.admin` can't.

ojeytonwilliams (changes requested):

cypress/e2e/dashboard/events/events-index.cy.ts
```diff
@@ -16,6 +16,7 @@ const eventData = {
   sponsor_ids: [],
   chapter_id: 1,
 };
+
 describe('events dashboard', () => {
   beforeEach(() => {
     cy.exec('npm run db:seed');
```
ojeytonwilliams, line 22: I can't comment on the next line, but could change it to log in as `admin@of.chapter.one`? We shouldn't need to be logged in as an owner to do this.

shootermv: the issue here is that there are 2 tests which broke:
`edit` and `delete` event
both of them trying to do operations on **first** event at homepage (which is not always event of chapter 1)

shootermv: may it can be good - to create different tests file, for `admin` user, which will have  `login('admin@of.chapter.one')` at its `beforeach` block

ojeytonwilliams: I'd like to keep the tests together, if possible.  It's not perfect, but it is helpful to know which page is supposed to be under test.  We could split them further, but I'd rather keep the system simple (no more than one spec per type of page).

So, what to do here?  I think it's okay for now to just log in as the instance user where necessary (when operating on events that may not be in chapter one).  Ideally the tests should target specific chapters/events, but it's not priority number one!  If you could leave a comment about *why* it's necessary to change user, that should help people understand and know what to fix.

shootermv: will it be better if  create different `describe` section on the same file?

ojeytonwilliams: Sure, that sounds good!

cypress/e2e/dashboard/events/events-index.cy.ts
```diff
@@ -294,23 +295,28 @@ describe('events dashboard', () => {
 
     cy.mhDeleteAll();
   });
+});
+
+describe('events endpoints', () => {
```
ojeytonwilliams, line 300:
```suggestion
describe('events dashboard', () => {
```

cypress/e2e/dashboard/events/events-index.cy.ts
```diff
@@ -16,6 +16,7 @@ const eventData = {
   sponsor_ids: [],
   chapter_id: 1,
 };
+
 describe('events dashboard', () => {
```
ojeytonwilliams, line 20:
```suggestion
// TODO: Move these specs into the other describe block, once we can make sure
// that Cypress is operating on an event from chapter 1.
describe('spec needing owner', () => {
```
Just so we know why the specs have been split.

ojeytonwilliams (changes requested):
Nitpicks: just a few things we can explain a bit better.

ojeytonwilliams (approved):
LGTM :+1: 

Thanks for your patience, @shootermv. It's nice to be able to get this in, though!

PR #1353, opened 2022-08-08
feat: use single field for user name

### LINKED ISSUE #1337: Replace first_name and last_name with a single entry
The _users_ schema includes two fields, _first_name_ and _last_name_ https://github.com/freeCodeCamp/chapter/blob/c3f40fa342d5e65397aa494ccc149f8815daf020/server/prisma/schema.prisma#L247-L252

This creates complexity for the MVP.  Not all cultures order given names and family names the same way.  To handle this correctly we would need to (at least) record the user's nationality and order appropriately. Either that or stick with first name + last name and require users to figure out if their family name should be their first name. Both approaches risk causing needless frustration.

Instead, we should replace the two entries with _name_.

gikf:

Closes #1337

---
- Only place with two fields left is `/schema.graphql`. Not sure if that file is used for anything anymore, it definitely is outdated.


ojeytonwilliams (changes requested):
Looks good! Only the tiniest of nitpicks:

ojeytonwilliams (approved):

ojeytonwilliams:
Perfect, thanks!

Issue #1263, opened 2022-07-19
Send both html and plain text in emails


ojeytonwilliams:
Tasks:

- [ ] Require both plain text and html when sending email
- [ ] Update the sendgrid logic to use both
- [ ] Provide plain text versions of all emails

### Original issue

@naomi-lgbt raised the concern that html emails are more likely to be marked as spam.  To improve deliverability we should consider sending text only emails.

## Pros

- Email should be more likely to arrive

## Cons

- We embed various links into our emails, so it'll be a little messy.

allella:
I'm not sure if text vs HTML is as relevant as in the old days, but since fCC likely sends a lot of emails, I'll defers to if that's still a thing.

The most important elements I'm aware of are having an SPF and DKIM that's designated to send email from the receiving server.  We already have some notes in the FAQs about this, but for anybody using a 3rd part email service, those services should be requiring the domain's DNS records to be properly validated.

The most likely way to end up in spam is to send emails from a random web server / VPS with no SPF or DKIM and generic configuration, like (from localhost, or from root) and anybody hosting their own servers surely has run into this problem and likely either knows how to properly configure, or else relays through a 3rd party.

naomi-lgbt:
> The most important elements I'm aware of are having an SPF and DKIM that's designated to send email from the sending server. We already have some notes in the FAQs about this, but for anybody using a 3rd part email service, those services should be requiring the domain's DNS records to be properly validated.

Oliver and I paired up this morning and are working on a Sendgrid integration for production environments - which will ensure that our emails aren't blocked (we already have the SPF, DKIM, and DMARC configured for our Sendgrid account).

> I'm not sure if text vs HTML is as relevant as in the old days, but since fCC likely sends a lot of emails, I'll defers to if that's still a thing.

I'm honestly not sure either, but I know when I first migrated our email newsletter Quincy was very adamant about ensuring the content-type was text and not html, so it might still be a concern.

allella:
GitHub sends a multipart email that includes both the plaintext and HTML version, so I have to assume that's something which will work for people in most parts of the world.  Is there someone more involved with the fCC email that would know if fCC is sending both text and HTML using this multipart header pattern?

```
Mime-Version: 1.0
Content-Type: multipart/alternative; boundary="--==_mimepart_62d6eda1bea04_1adfd77890670"; charset=UTF-8
Content-Transfer-Encoding: 7bit
Precedence: list
X-GitHub-Sender: sender
[...]

PR #1357, opened 2022-08-08
fix: resolve mail error

naomi-lgbt:

Closes #XXXXX

The API gets mad if you send both `text` and `html`, so just send `html`.

### DIFF
--- a/server/src/services/MailerService.ts
+++ b/server/src/services/MailerService.ts
@@ -110,7 +110,6 @@ export default class MailerService {
             from: this.sendgridEmail,
             subject: this.subject,
             html: this.htmlEmail,
-            text: this.backupText,
             trackingSettings: {
               clickTracking: {
                 enable: false,
--- a/server/src/services/MailerService.ts
+++ b/server/src/services/MailerService.ts
@@ -109,7 +109,7 @@ export default class MailerService {
             to: email,
             from: this.sendgridEmail,
             subject: this.subject,
-            html: this.htmlEmail,
+            html: this.htmlEmail || this.backupText,
             trackingSettings: {
               clickTracking: {
                 enable: false,


ojeytonwilliams (commented):
Could we fallback to the text if there's no html?

naomi-lgbt:
> Could we fallback to the text if there's no html?

Yes! I pushed a commit that uses the logical OR as a fallback - but this means the MIME type will still be HTML.

If we'd rather have a dynamic MIME type, let me know and I'll push a slightly more complex fallback.

ojeytonwilliams (approved):
No, no, that should be enough. As long as we send something, that's fine for now.

allella:
@naomi-lgbt is this the SendGrid API that was not taking both plaintext and HTML?

The sendgrid/mail package appears to shows both text and html fields without setting dynamic mime types.

Sites like eBay and GitHub  [previous chat on this topic](https://github.com/freeCodeCamp/chapter/issues/1263#issuecomment-1189407373). send a single email with a mixed body that wraps the text and HTML in their respective content type delimiters.

I'd expect Sendgrid to construct those content types and delimiters, so that's mainly why I'm curious if you're saying their API was rejecting both values. 

https://www.npmjs.com/package/@sendgrid/mail
https://github.com/sendgrid/sendgrid-nodejs/tree/main/packages/mail

naomi-lgbt:
> is this the SendGrid API that was not taking both plaintext and HTML?

There is some sort of abstraction happening in the npm package - the API expects a `content` key, the library magically handles that if you pass `text` or `html`, but not if you pass both?

We were seeing "content cannot be an empty string" errors - removing the text property resolved those.

There may be a way to send both through the API (likely using the `content` property, which is an array), but given that SendGrid is used here primarily as a temporary fix for our staging instance to function, it didn't seem worth the time investment.

allella:
Okay. Related to the staging, how would we see the test magic links?  Like I just tried to log in as foo@bar.com but can't see a magic link in the console or https://chapter.freecodecamp.dev:8025

naomi-lgbt:
I don't believe the links are logged anywhere in production.

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Tags:** Brittle

allella:
Okay, that makes sense for production, but I've been assuming https://chapter.freecodecamp.dev is more of a staging site.

If someone wants to log in to https://chapter.freecodecamp.dev then will they need to register an account and expect a real email? Would anybody be able to log in as an administrator or owner on that copy, or is that not the intention?

I could see allowing people to create accounts, or log in with elevated permissions being a bad idea in the long run, so I guess the best question is how are we expecting this https://chapter.freecodecamp.dev to be used?

naomi-lgbt:
Honestly, that might be a better question for @ojeytonwilliams - my work on Chapter has been relatively small in scope and I'm not nearly as up-to-date on the overall plan.

ojeytonwilliams:
> I could see allowing people to create accounts, or log in with elevated permissions being a bad idea in the long run, so I guess the best question is how are we expecting this https://chapter.freecodecamp.dev/ to be used?

The first goal was to test the deployment process and see what kind of surprises we run into.  That's now been done.  There's lots to improve, but everything seems to be working as expected (both the pipeline and the app).

As for what we do with it from now, there are a few possibilities.  From our experience with fcc, it's very helpful to have a staging site that mirrors production in every respect, except with a different db.  I think we'll want that for Chapter, too. It would not be for development, exactly, just for confirming that nothing is terribly broken before pushing a change to production. 

If we want a testing site where developers can log in with elevated permissions, that's possible, but there's a bit of work to be done before we can test everything.  For one thing, we would need to set up a mail server, since we can't spam random email addresses. They might be real and, even if they aren't, it would potentially trash our mail reputation (@naomi-lgbt may know for sure).

Also, we'll need some code changes before it's possible to login as, say, `admin@of.chapter.two`.  I'll be working on those soon.

So, to actually answer your question: the dev site will start it's life as a testing site that's lets you sign in as a member. Then the higher permission users will be available and finally email will be captured by, say, mail.chapter.freecodecamp.org (just running mailhog, I imagine).  Before we start deploying for real we'll want a staging site, so I'll set that up in the coming weeks.

naomi-lgbt:
> mail.chapter.freecodecamp.org

please can we call it chapter-mail.freecodecamp.org? seeing a fourth-level domain hurts my soul.

> For one thing, we would need to set up a mail server, since we can't spam random email addresses. They might be real and, even if they aren't, it would potentially trash our mail reputation

Unfortunately this is a bit complicated. There's two facets to this issue.

First, we need to be careful of our Sendgrid usage - admittedly, shoving SendGrid in this project was a quick hack to get the staging site *working* and is not meant to be long term. However, the benefit of SendGrid is that our DKIM/SPF/DMARC are all configured, and *more importantly* we have dedicated IPs that maintain a 99% positive sender reputation.

When we remove Sendgrid, which we will, our new approach will be a completely different IP address. We'll need to be careful here as we "warm it up". I don't imagine the volume of emails being sent on .dev is going to be high, BUT it's something to keep in mind. If we start sending 1000 emails a day from a new IP address, some mail providers will pick that up as spam and block the IP (Outlook is the worst at this, in my experience).

I agree that the best solution for the staging instance is something like a mailhog instance that's publicly viewable. This will allow our contributors who are looking at a production-like environment the ability to sign in as whichever spoofed email they need to test (based on things like permissions).

naomi-lgbt:
I think the root of this discussion, ultimately, is how do we want the .dev site to behave?

Should it follow our `/learn` paradigm, where it mimics production as much as possible and gives us an opportunity to double-check deployments before pushing forward?

If so, we'll want to address the email problem in a production-like manner.

If we're okay with it being a sandbox and not a production validation step, then we might be okay deviating from that. But IMHO it doesn't seem beneficial to host a staging site that isn't mirroring production - I could run it locally and get the same experience. We *want* things to work like they do on prod, so we can confirm they *do* work.

ojeytonwilliams:
> Should it follow our /learn paradigm, where it mimics production as much as possible and gives us an opportunity to double-check deployments before pushing forward?

We will need that, for sure.  My main concern is that we currently have a bunch of randomly generated email addresses that we would be spamming if we allowed that site to send emails.  I'm not sure how best to avoid that (for now I've just disabled Sendgrid on the vm).

> But IMHO it doesn't seem beneficial to host a staging site that isn't mirroring production 

I agree that a staging site should mirror production and that that's what we'll want before we launch.  I do think that a _testing_ site would be useful, at least for now, though.  How useful?  Hard to say, so I don't think tons of dev work should go into it.  However, as far as I know, all we're missing is the mailhog server.

naomi-lgbt:
> since this subdomain is public, and since other instance owners may do the same with dev / stage copies, shall we add a default <meta name="robots" content="noindex"> tag to the HTML head when the environment != production? I used to do this, and also a dynamic robots.txt, to block search engines from accidentally indexing domains. I actually was Googling for something yesterday and found a big organization that had their test domain inadvertently indexed.

Yes I think this is a very good idea.

ojeytonwilliams:
One thing that's been bothering me is that we want simple safeguards like "Is `NODE_ENV=production`? Alright, you need all the keys and all the dev-specific stuff (like dev-login) will be disabled".  However, we *also* want people to be able to test as different users, ideally without having to set up locally.

It's possible I'm missing something obvious, but I don't think we can do both.

If that's correct, then I propose we make .dev as production-like as we can (as Naomi said).  I did want to have an instance that was more 'testing' than 'staging', but the more I think about it, the more I don't like it.

> I could see it being useful in some testing cases, like a designer, UX, accessibility helper that's not likely to jump through the technical hoops of running a local instance.

I do agree, but if we put .dev in production mode *and* made it so that new users got, say, owner permissions the default, I can see that going very badly wrong.

It's not ideal, but I think GitPod will have to do as our simplest testing environment.

> I've seen some other test systems that are semi-open to the public that rebuild and wipe any data one or more times daily to avoid any real data persisting for spammers to harvest.

I think that's a good idea.  While we won't be intentionally exposing anything to untrusted users, it's best to be safe.  I'll set up a scheduled job to reset the db.

No-indexing is a good plan. Is the meta tag enough?  If so, I'll make sure it's added to .dev.

allella:
I'm fine with whatever and don't mean to open things up spam or hacking concerns.

If we have a need for someone, like Cecilia with the accessibility testing, to test then we could give them a temporary account on a dev or stage copy.

I seem to recall the systems that clear the database had a banner at the top of the page warning "THIS WILL RESET AT ##:##:## GST, or some such. I don't think that's necessary for now, but it helped avoid someone starting to test and getting unknowingly caught in a reset.

The meta tag noindex will be sufficient [as long as it's on all pages / URLs. Also, Google has a note that the robots.txt shouldn't be blocking the URL](https://developers.google.com/search/docs/advanced/crawling/block-indexing), which is unlikely to be the case since we don't have a robots.txt

You can do the same with HTTP headers or a robots.txt rule, but the meta tag is the easiest and avoids server-specifics.

ojeytonwilliams:
> I'm fine with whatever and don't mean to open things up spam or hacking concerns.

Sorry, I didn't mean to suggest you were.  I'm just trying to figure out how best to proceed.

> The meta tag noindex will be sufficient [as long as it's on all pages / URLs. Also, Google has a note that the robots.txt shouldn't be blocking the URL](https://developers.google.com/search/docs/advanced/crawling/block-indexing), which is unlikely to be the case since we don't have a robots.txt

Lovely, I'll do that tomorrow.

Re: the reset, it's surprisingly difficult once you start using managed dbs.  Since they don't like you connecting from any old ip, my plan to do it from an action doesn't work.  Even if we could allow whatever IPs actions use, we shouldn't!  I think resetting via the VM (which DO will allow) should be fine for now, but I'll sleep on it since it's not completely trivial.

PR #1361, opened 2022-08-09
fix: chapter presentation

Sboonny:

Clear the 7th task in #1339

----

This is basic prototype without refactoring, to get feedback if the layout is okay, and I can add tags type and more. Or we want different layout


gikf:
I mean it looks okay for me. But it doesn't look consistent with the rest of page.

Sboonny:
@gikf something like this?

image

gikf:
I'm talking about them too. It's standing out for me due to being different from the rest.

Sboonny:
here is how it looks currently

image

Note I didn't change the tags, because I will change the tags for Event and chapter later. I can do it in this pull request, if you think it doesn't need another pull request.

ojeytonwilliams:
@Sboonny the ChapterCard still needs updating to use the array of events, but I think it's okay otherwise.

I was going to *just* provide the event, with no tags, but I quickly realised this was not a small change.  I propose we remove the tags in a separate PR. That should make it easier to revert post-MVP when we want the tags back.

Sboonny:
Sure, I will open issue after this one is merged

client/src/components/ChapterCard.tsx
```diff
@@ -1,59 +1,51 @@
...
+        <GridItem
+          colStart={1}
+          colSpan={2}
+          marginInline={'1em'}
+          marginBlock={'.5em'}
+        >
+          <Text mt="2" as="p" fontWeight={400} fontSize={'md'}>
+            {chapter?.description}
+          </Text>
+        </GridItem>
```
Sboonny, line 47:
```suggestion
        </GridItem>
        {chapter.events && (
          <GridItem colSpan={3}>
          {chapter.events.map((_key, index) => (
            <Link href={`/events/${chapter?.events[ index ].id}`} _hover={{}}>
              <Flex paddingInline={"1em"} paddingBlock={".5em"} justifyContent={"space-between"}>
                <Text mt="2" fontWeight={600} fontSize={'md'} maxW={'10em'}>{chapter?.events[ index ].name}</Text>
                <Text mt="2" fontWeight={600} fontSize={'md'}>{chapter?.events[ index ].start_at}</Text>
              </Flex>
            </Link>
          ))}
          </GridItem>
          )
        }

**Review by Sboonny — COMMENTED** [review]
I couldn't add this, because of eslint aren't happy with it, but it doesn't give me errors.

I know it need to change the `_key`, but I don't know how to work around that
```ts
 {chapter.events.map((_key, index) => (
```

client/src/components/ChapterCard.tsx
Sboonny: Missing "key" prop for element in iterator error

Sboonny:
I am going to merge this tonight. If you think it needs improvement in any way, feel free to mention it.

If it merged and you have suggestions or issues, I will still implement it after you mention them

client/src/modules/chapters/graphql/queries.ts
```diff
@@ -80,6 +80,23 @@ export const CHAPTERS = gql`
...
+        canceled
+        image_url
+        tags {
+          tag {
+            id
+            name
+          }
+        }
+        invite_only
+        canceled
```
gikf, line 98:
```suggestion
        start_at
```
No need to have other fields in query if they aren't used

client/src/modules/home/graphql/queries.ts
```diff
@@ -28,6 +28,23 @@ export const HOME_PAGE_QUERY = gql`
...
+        canceled
+        image_url
+        tags {
+          tag {
+            id
+            name
+          }
+        }
+        invite_only
+        canceled
```
gikf, line 46:
```suggestion
        start_at
```
Same as the above

PR #1396, opened 2022-08-14
fix: wait for loading url parameter

gikf:

---

- Error occurs when adding venue to a specified chapter - i.e. `/dashboard/chapters/1/new-venue`.
- On the fresh load chapters dropdown appear for a short while.
- `useParam` don't differentiate when parameter isn't yet loaded and when it doesn't exist in the url.  Both of these give param as `-1`.
- At the same time, when param is provided in the url, couple renders happen before param is loaded, and rendering is as if param would be `-1`. When param from url is picked up there's a conflict resulting in warning about hooks order (see below).
- Change makes sure the form is rendered after parameter from url is loaded.

<details>
<summary>Warning</summary>

```
Warning: React has detected a change in the order of Hooks called by VenueForm. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://reactjs.org/link/rules-of-hooks

   Previous render            Next render
   ------------------------------------------------------
1. useContext                 useContext
2. useRef                     useRef
3. useState                   useState
4. useContext                 useContext
5. useMemo                    useMemo
6. useCallback                useCallback
7. useSyncExternalStore       useSyncExternalStore
8. useRef                     useRef
9. useState                   useState
10. useCallback               useCallback
11. useRef                    useRef
12. useEffect                 useEffect
13. useEffect                 useEffect
14. useContext                useEffect
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
</details>


Sboonny (approved):

ojeytonwilliams:
@gikf what do you think to expanding `useParam` to return both the value and `isReady`? I would have thought that almost every page that uses `useParam` could use `isReady` to avoid rendering something we don't want (when id is -1).

gikf:
That also would work. I thought at first about changing `useParam`, to return `undefined` when param is not ready.

To make sure we are on the same page - make `useParam` to return `{ param, isReady }`?

ojeytonwilliams:
It might be worth trying out returning `null` rather than `-1`, to cover the case where the id doesn't exist for some reason. That's even more of an edge case, though, so I think returning `{ param, isReady }` should everything we really care about.

gikf:
- I've changed `useParam` and places where it's used. Without expanding the usage of `isReady` at this time, except of the adding new venue for specified chapter, as originally in the PR.
- I was trying to make `useParam` return object with dynamic named property for param. Ie. `useParam('eventId')` -> `{ eventId: param, isReady }`. It was giving me type troubles. For some reason, in `{ [key]: param, isReady }`, the type was `{ [x: string]: number | boolean; isReady: boolean }`.
- I've replaced usage of `getId` and removed the util, as it was mirroring previous `useParam` functionality and after changes `useParam` should be preferred.

ojeytonwilliams (approved):
LGTM :+1: 

For the return type, the issue is that `{ [key]: param, isReady: boolean }` does mean that using the index 'key' could get either a `number` or a `boolean`, since `isReady` is just another string we could use as an index.

To work around this, we'd need to restrict the possible values that `key` could be and create specific objects like `{ chapterId: param, isReady }` so that TS knows exactly which properties it has.  Might be worth it at some point, but this is fine for now!

PR #1398, opened 2022-08-14
feat: Responsive header

Sboonny:

This pull request aims to 
- show the logo in small screens
- add menu, so it isn't hard for mobile users to interact with the page
- remove background `inherit` from search bar, so it can be readable


ojeytonwilliams:
@Sboonny I'm not sure this is quite correct.  The issue is that MenuItem is rendered as a button and that button contains a link, but the button does not react to clicks.  What this means is there's a clickable link inside the button, but the area around that link does not respond.

I don't entirely understand why the server rendered page doesn't match the hydration, but we can cross that bridge later (if at all).

Sboonny:
@ojeytonwilliams What browser are you using? I can't replicate it

ojeytonwilliams:
Peek 2022-08-17 14-35

I saw the same behaviour on Chrome, so it doesn't seem browser specific.

As I understand it, the buttons that MenuItem renders don't do anything `onClick`, so navigation only happens if you happen to click on the link part of the element. i.e. the bit that gets underlined on hover.  I believe it's standard practice to just use a Next Link (or `chakra-next-link`) for `a` tags and use buttons for other (non-link related) things.

client/src/components/PageLayout/Header.tsx
```diff
@@ -98,8 +98,8 @@ export const Header: React.FC = () => {
           />
         </Link>
         <Input
-          background="hsl(239, 100%, 97%)"
-          color="hsl(239, 100%, 13%)"
+          background={'gray.15'}
```
Sboonny, line 101:
```suggestion
          background={'gray.10'}
```

ojeytonwilliams:
@Sboonny I had a little look around and found this clever approach: https://github.com/chakra-ui/chakra-ui/issues/4892#issuecomment-1003213102 which seems to be the way to go right now.  Without doing something like this the result is that clicking on the login button collapses the menu while clicking on a link does not.

ojeytonwilliams (commented):
Thanks @Sboonny, this feels nice to use! LGTM :+1:

ojeytonwilliams (approved):

PR #1413, opened 2022-08-17
fix: restrict 'user ban' functionality to member with ChapterBanUser rule

shootermv:

Im adding here cypress test that should check - that admin of chapter1 cannot ban members of chapter2,
if he tries to do that - he should see 'access denied' notification
Closes part of #1217

### DIFF
--- a/cypress/e2e/dashboard/chapters/users.cy.ts
+++ b/cypress/e2e/dashboard/chapters/users.cy.ts
@@ -141,7 +141,6 @@ describe('Chapter Users dashboard', () => {
   }
 
   it('someone whos NOT an admin of a chapter, should NOT be able to ban a chapter users', () => {
-    cy.logout();
     cy.login('admin@of.chapter.one');
     cy.visit(`/dashboard/chapters/2/users`);
 
--- a/cypress/e2e/dashboard/chapters/users.cy.ts
+++ b/cypress/e2e/dashboard/chapters/users.cy.ts
@@ -155,6 +155,7 @@ describe('Chapter Users dashboard', () => {
       .findByRole('button', { name: 'Ban' })
       .click();
     cy.findByRole('button', { name: 'Confirm' }).click();
+    cy.findByRole('status').contains('access denied', { matchCase: false });
   });
 
   it('an admin cannot ban themselves', () => {
--- a/cypress/e2e/dashboard/chapters/users.cy.ts
+++ b/cypress/e2e/dashboard/chapters/users.cy.ts
@@ -140,6 +140,23 @@ describe('Chapter Users dashboard', () => {
       .as('firstUnbannedMember');
   }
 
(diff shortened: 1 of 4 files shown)


ojeytonwilliams (changes requested):
Hey @shootermv, thanks for working on this. This should be a `ChapterPermission`, since a chapter admin should be able to ban people from their chapter, but no others.

Also, could you create a test that confirms this, that would be awesome.  The `rejects chapter admin from changing chapter user role` spec should be a good starting point, but with `'admin@of.chapter.two'` instead.  It will also need a `banChaperUser` Cypress command, obviously!

shootermv:
as far as i understood, test for `admin@of.chpater.two` not bale to ban user of chapter one need to be added, im i right?

server/prisma/generator/factories/instanceRoles.factory.ts
```diff
@@ -20,6 +24,7 @@ const roles: InstanceRole[] = [
     permissions: [
       InstancePermission.ChapterJoin,
       InstancePermission.ChapterSubscriptionsManage,
+      ChapterPermission.ChapterBanUser,
```
ojeytonwilliams, line 27:
```suggestion
```
Members should not be able to ban anyone.

server/prisma/generator/factories/instanceRoles.factory.ts
```diff
@@ -1,6 +1,10 @@
 import { prisma } from '../../../src/prisma';
 
-import { InstancePermission, Permission } from '../../../../common/permissions';
+import {
+  ChapterPermission,
+  InstancePermission,
+  Permission,
+} from '../../../../common/permissions';
```
ojeytonwilliams, line 7:
```suggestion
import { InstancePermission, Permission } from '../../../../common/permissions';
```
See below

ojeytonwilliams (changes requested):
> as far as i understood, test for admin@of.chpater.two not able to ban user of chapter one need to be added, Am i right?

Yes, please!

cypress/e2e/dashboard/chapters/users.cy.ts
```diff
@@ -140,6 +140,24 @@ describe('Chapter Users dashboard', () => {
       .as('firstUnbannedMember');
   }
 
+  it('someone whos NOT an admin of a chapter, should NOT be able to ban a chapter users', () => {
+    cy.logout();
```
ojeytonwilliams, line 144:
```suggestion
```
`cy.login` changes the session, so there's no need to log out first.

ojeytonwilliams (changes requested):
This is looking good, just the one nitpick:

ojeytonwilliams (approved):

ojeytonwilliams:
Thanks @shootermv :rocket:

PR #1416, opened 2022-08-17
feat: add decorator for CanBeBanned

sirasitxp:

Closes part of #1217(canBeBanned)


server/prisma/generator/factories/instanceRoles.factory.ts
```diff
@@ -1,6 +1,10 @@
 import { prisma } from '../../../src/prisma';
 
-import { InstancePermission, Permission } from '../../../../common/permissions';
+import {
+  ChapterPermission,
+  InstancePermission,
+  Permission,
+} from '../../../../common/permissions';
```
ojeytonwilliams, line 7:
```suggestion
import { InstancePermission, Permission } from '../../../../common/permissions';
```
We members should not be able to ban users, so we do not need to pull out the chapter permissions here.

server/prisma/generator/factories/instanceRoles.factory.ts
```diff
@@ -20,6 +24,7 @@ const roles: InstanceRole[] = [
     permissions: [
       InstancePermission.ChapterJoin,
       InstancePermission.ChapterSubscriptionsManage,
+      ChapterPermission.ChapterBanUser,
```
ojeytonwilliams, line 27:
```suggestion
```
As mentioned, this isn't something members should be able to do.

ojeytonwilliams (changes requested):
Hi, @hisnameispum, thanks for working on this. Could you restore the chapter page? It seems to have been deleted by accident.

Also, to avoid regressions it would be great to have tests that confirm that the permissions work. Namely that users who should not be able to ban actually cannot.

https://github.com/freeCodeCamp/chapter/pull/1346 is a good example of this.

ojeytonwilliams:
Hi @hisnameispum if you've got a moment, could you take a look at the failing tests?

Sboonny:
@ojeytonwilliams, I am wondering how to find admin role in users because I can't find it.

image


I think the cypress test about finding admin role fails, because it doesn't exist in user dashboard

ojeytonwilliams:
> I am wondering how to find admin role in users because I can't find it.

There aren't instance admins (yet), since we figured they weren't MVP worthy.  The roles are instance owner, instance member, chapter admin and chapter member.

> after fixing status error, banned admin don't have permission to view the table at all

Yep, that's right. I'll update the tests.

ojeytonwilliams (approved):
LGTM :+1:

PR #1423, opened 2022-08-19
feat(dx): add cypress stuffs for gitpod

ShaunSHamilton:

There is the small chance we need to do the whole `echo 'export CYPRESS_BASE_URL...' | ~/.bashrc` thing.


ojeytonwilliams:
I'm seeing

> Your system is missing the dependency: Xvfb
>
> Install Xvfb and run Cypress again.

ShaunSHamilton:
Did you run `npm run cypress:install-build-tools`? because that is what it does

Oliver, do you think the cypress config will need to be changed?

ojeytonwilliams:
@ShaunSHamilton I think it will, but I'm not sure exactly what needs changing.  When I tried setting the `baseUrl` to the result of `gp url 3000` everything failed with 401 errors.

ShaunSHamilton:
> @ShaunSHamilton I think it will, but I'm not sure exactly what needs changing. When I tried setting the `baseUrl` to the result of `gp url 3000` everything failed with 401 errors.

You should just add `process.env.CYPRESS_BASE_URL || 'localnonesense'`

ojeytonwilliams:
@ShaunSHamilton yeah, that needs to be done, but it doesn't seem to be enough.

package.json
```diff
@@ -28,9 +28,13 @@
...
     "change-user:owner": "npm run change-user foo@bar.com",
-    "cypress:open": "cypress open",
-    "cypress:run": "cypress run",
+    "precypress": "npm run cypress:install",
+    "cypress": "cypress",
+    "cypress:open": "npm run cypress -- open",
+    "cypress:run": "npm run cypress -- run",
     "cypress:coverage:report": "npx nyc report",
+    "cypress:install": "cypress install",
+    "cypress:install-build-tools": "sh ./cypress-install.sh",
```
Sboonny, line 37: Random question, why do we need explicitly to have `npm run precypress`, while there is `npm run cypress:install`?

What the benefit over `"precypress": "cypress install"`

ojeytonwilliams: Good question! `cypress` isn't a global command, it's installed locally in the chapter repo (in node modules). That means we need to invoke it via npm.

The main point is to ensure that precypress gets called (since that's responsible for installing Cypress). Hence the other scripts call the cypress script.

Final piece of the puzzle: `pre<anything>` gets called before `<anything>` (cypress, in this case)

Sboonny (commented):
I don't know, if this ready for review but I have tested it and `cypress` after installing isn't working

image

----

`npm run cypress:open` run magically, but it can't find `tsconfig` file

image

Other aspects are working without issue so far, thank you for the hard work 💚

ojeytonwilliams:
Thanks for the effort, @ShaunSHamilton and @Sboonny, but this has become a little stale.

PR #1430, opened 2022-08-21
feat: chapters dropdown event create

### LINKED ISSUE #1215: Event and Venue creation pages

gikf:

Closes #1215 

---
- Adds general `Add new` button on the events dashboard, leading to `/dashboard/events/new` route, with add event form including chapters dropdown.
- Dropdown uses user `admined_chapters` array. This can be a bit weird, as permission required for _admining chapter_ - `ChapterEdit` is not the permission required for adding new events - `EventCreate`.
- `Add new` Button is visible only to those with permission to admin chapter.
- The bulk of the changed lines in `components/EventForm.tsx` is due to removal of blank dummy parent element and dedenting the whole form.

### DIFF
--- a/client/src/modules/dashboard/Events/components/EventForm.tsx
+++ b/client/src/modules/dashboard/Events/components/EventForm.tsx
@@ -44,14 +44,21 @@ import {
 import 'react-datepicker/dist/react-datepicker.css';
 
 const EventForm: React.FC<EventFormProps> = (props) => {
-  const { onSubmit, data, loading, submitText, chapterId, loadingText } = props;
-  const isChaptersDropdownNeeded = chapterId === -1;
+  const {
+    onSubmit,
+    data,
+    loading,
+    submitText,
+    chapterId: initialChapterId,
+    loadingText,
+  } = props;
+  const isChaptersDropdownNeeded = initialChapterId === -1;
   const {
     loading: loadingChapter,
     error: errorChapter,
     data: dataChapter,
   } = useChapterQuery({
-    variables: { chapterId },
+    variables: { chapterId: initialChapterId },
   });
 
   interface Chapter {
@@ -77,7 +84,7 @@ const EventForm: React.FC<EventFormProps> = (props) => {
         start_at: new Date(),
         ends_at: new Date(),
         venue_type: VenueType.PhysicalAndOnline,
-        chapter_id: chapterId,
+        chapter_id: initialChapterId,
       };
     }
     return {
@@ -94,7 +101,7 @@ const EventForm: React.FC<EventFormProps> = (props) => {
       venue_id: data.venue_id,
       image_url: data.image_url,
       invite_only: data.invite_only,
-      chapter_id: chapterId,
+      chapter_id: initialChapterId,
     };
   }, []);
 
@@ -124,13 +131,13 @@ const EventForm: React.FC<EventFormProps> = (props) => {
   const watchSponsorsArray = watch('sponsors');
   const inviteOnly = watch('invite_only');
   const venueType = watch('venue_type');
-  const chapter = watch('chapter_id');
+  const chapterId = watch('chapter_id');
 
   const {
     loading: loadingVenues,
     error: errorVenues,
     data: dataVenues,
-  } = useChapterVenuesQuery({ variables: { chapterId: chapter } });
+  } = useChapterVenuesQuery({ variables: { chapterId } });
 
   useEffect(() => {
     resetField('chapter_id', { defaultValue: adminedChapters[0]?.id ?? -1 });
@@ -187,12 +194,11 @@ const EventForm: React.FC<EventFormProps> = (props) => {
               })}
               isDisabled={loading}
             >
-              {adminedChapters?.length &&
-                adminedChapters.map(({ id, name }) => (
-                  <option key={id} value={id}>
-                    {name}
-                  </option>
-                ))}
+              {adminedChapters.map(({ id, name }) => (
+                <option key={id} value={id}>
+                  {name}
+                </option>
+              ))}
             </Select>
           </FormControl>
         )}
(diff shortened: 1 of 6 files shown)


client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -113,6 +124,24 @@ const EventForm: React.FC<EventFormProps> = (props) => {
...
+  useEffect(() => {
+    resetField('chapter_id', { defaultValue: adminedChapters[0]?.id ?? -1 });
+  }, [adminedChapters]);
+
+  useEffect(() => {
+    resetField('venue_id', {
+      defaultValue:
+        defaultValues.venue_id ?? dataVenues?.chapterVenues[0]?.id ?? -1,
+    });
+  }, [dataVenues]);
```
ojeytonwilliams, line 144: It's a bit out of scope of this PR, but this got me thinking: could we avoid resetting if we pulled the data fetching up into the parent compoenent? i.e. don't create the form until we have all the data and so avoid setting the wrong `defaultValue`.

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -113,6 +124,24 @@ const EventForm: React.FC<EventFormProps> = (props) => {
   const watchSponsorsArray = watch('sponsors');
   const inviteOnly = watch('invite_only');
   const venueType = watch('venue_type');
+  const chapter = watch('chapter_id');
```
ojeytonwilliams, line 127:
```suggestion
  const chapterId = watch('chapter_id');
```
Nit

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -113,6 +124,24 @@ const EventForm: React.FC<EventFormProps> = (props) => {
...
+
+  const {
+    loading: loadingVenues,
+    error: errorVenues,
+    data: dataVenues,
+  } = useChapterVenuesQuery({ variables: { chapterId: chapter } });
```
ojeytonwilliams, line 133:
```suggestion
  } = useChapterVenuesQuery({ variables: { chapterId } });
```
Nit

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -134,274 +163,288 @@ const EventForm: React.FC<EventFormProps> = (props) => {
...
+                valueAsNumber: true,
+              })}
+              isDisabled={loading}
+            >
+              {adminedChapters?.length &&
+                adminedChapters.map(({ id, name }) => (
+                  <option key={id} value={id}>
+                    {name}
+                  </option>
+                ))}
```
ojeytonwilliams, line 195:
```suggestion
              {adminedChapters.map(({ id, name }) => (
                <option key={id} value={id}>
                  {name}
                </option>
              ))}
```
Another nit, tbh. React seems to cope with an empty array, it just renders nothing.

ojeytonwilliams (changes requested):
I agree that `admined_chapters` is a bit odd, since it's the wrong permission, but I think it's fine for the MVP.  I've tracked it in https://github.com/freeCodeCamp/chapter/issues/1435

client/src/modules/dashboard/Events/components/EventForm.tsx
gikf: For venues problem would be fact venues belong to chapter. Unless all venues, for corresponding _possible_ admined chapters, are loaded before rendering form, there's bound to be setting default
when `dataVenues` changes, after selected chapter changed.

client/src/modules/dashboard/Events/components/EventForm.tsx
gikf: There's `chapterId` prop passed to the form. How does it sound renaming it in destructuring to ie. `initialChapterId` and using `chapterId` as in the suggestion?

ojeytonwilliams: Ah, good point. Going with `initialChapterId` sounds perfect.

client/src/modules/dashboard/Events/components/EventForm.tsx
ojeytonwilliams: That's a shame, but it sounds like some amount of resetting is inevitable.  Oh well.

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -43,19 +44,33 @@ import {
...
-  const { onSubmit, data, loading, submitText, chapterId, loadingText } = props;
+  const {
+    onSubmit,
+    data,
+    loading,
+    submitText,
+    chapterId: initialChapterId,
+    loadingText,
+  } = props;
+  const isChaptersDropdownNeeded = initialChapterId === -1;
```
Sboonny, line 55: We should change the variable name `isChaptersDropdownNeeded` because if someone doesn't know the issue and the pull request, they won't understand it.

What do you think of `associateChapters`

gikf: Some additional context (like usage) might be needed to fully understand it, but for how it's used I think the name is fairly accurate. `associatedChapters` is for me more vague, and indicates bigger scope than this variable has.

ojeytonwilliams (approved):
Alrighty, this LGTM :+1: Thanks @gikf!

PR #1441, opened 2022-08-23
refactor: DRY rsvpEvent

ojeytonwilliams:

This flows slightly better, I think.

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -281,6 +281,7 @@ export class EventResolver {
       },
     });
 
+    // TODO: find by permission, not role.
     const chapterAdministrators = await prisma.chapter_users.findMany({
       where: {
         chapter_id: chapterId,
@@ -290,9 +291,10 @@ export class EventResolver {
       ...chapterUserInclude,
     });
 
-    const userChapter = ctx.user.user_chapters.find(
-      (user_chapter) => user_chapter.chapter_id === chapterId,
-    );
+    const isSubscribedToChapter =
+      ctx.user.user_chapters.find(
+        (user_chapter) => user_chapter.chapter_id === chapterId,
+      )?.subscribed ?? true;
 
     const oldEventUser = await prisma.event_users.findUnique({
       include: { rsvp: true },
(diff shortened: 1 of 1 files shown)


gikf (approved):
Yes it does :+1: 

One small nitpick, I'd think about pulling part below in the `else` of `if (oldEventUser)`
https://github.com/freeCodeCamp/chapter/blob/97077cb3faf93974ba1a4a8697dc40e21aee25a4/server/src/controllers/Events/resolver.ts#L341-L347

It's not going to stretch it vertically much, and after moving having two conditions in `if` could make it a bit easier to follow.

If you don't think this change is worth it, feel free to merge.

ojeytonwilliams:
Cheers, that conditional was a bit ugly!

server/src/controllers/Events/resolver.ts
```diff
@@ -325,28 +324,25 @@ export class EventResolver {
...
+        user: { connect: { id: ctx.user.id } },
+        event: { connect: { id: eventId } },
+        rsvp: { connect: { name: newRsvpName } },
+        event_role: { connect: { name: 'member' } },
+        subscribed: isSubscribedToChapter,
+      };
+      eventUser = await prisma.event_users.create({
+        data: eventUserData,
+        include: eventUserIncludes,
       });
```
gikf, line 346:
```suggestion
      });
      
      if (newRsvpName !== 'waitlist' && isSubscribedToChapter) {
        await createReminder({
          eventId,
          remindAt: sub(event.start_at, { days: 1 }),
          userId: ctx.user.id,
        });
      }
```
It needs to be moved after the `event_user` is created

ojeytonwilliams:
Thanks @gikf it should all be okay now.  It bugs me that there's a (slightly) non-obvious ordering requirement, but I can live with it for now!

gikf:
Well... can do!
```
      const eventUserData: Prisma.event_usersCreateInput = {
        user: { connect: { id: ctx.user.id } },
        event: { connect: { id: eventId } },
        rsvp: { connect: { name: newRsvpName } },
        event_role: { connect: { name: 'member' } },
        subscribed: isSubscribedToChapter,
        ...(newRsvpName !== 'waitlist' &&
          isSubscribedToChapter && {
            event_reminder: {
              create: {
                remind_at: sub(event.start_at, { days: 1 }),
              },
            },
          }),
      };
```

ojeytonwilliams:
Yeaah, we could... I think the separation between reminders and events is useful, though.  Particularly given that we'll likely offload some of the reminding duties to Google.

gikf (approved):

Issue #1316, opened 2022-07-31
Flaky cypress tests


gikf:
Few cypress tests are lately getting more on the flaky side

- [ ] https://github.com/freeCodeCamp/chapter/blob/9fcd773c425df02e0c55b81594df6f7e4a811317/cypress/e2e/dashboard/events/events-index.cy.ts#L176-L213
- [ ] https://github.com/freeCodeCamp/chapter/blob/9fcd773c425df02e0c55b81594df6f7e4a811317/cypress/e2e/dashboard/events/events-index.cy.ts#L115-L174
- [ ] https://github.com/freeCodeCamp/chapter/blob/9fcd773c425df02e0c55b81594df6f7e4a811317/cypress/e2e/dashboard/events/%5BeventId%5D.cy.ts#L76-L91
- [ ] https://github.com/freeCodeCamp/chapter/blob/9fcd773c425df02e0c55b81594df6f7e4a811317/cypress/e2e/events/%5BeventId%5D.cy.ts#L74-L100
- [ ] https://github.com/freeCodeCamp/chapter/blob/9fcd773c425df02e0c55b81594df6f7e4a811317/cypress/e2e/dashboard/events/events-index.cy.ts#L24-L56

ojeytonwilliams:
@gikf this one was quite interesting.  Turns out the trick we used to make it 'necessary' to pass the `chapterId` to the `updateVenue` and `deleteVenue` resolvers has come back to bite us.

When you try and update a venue to have a physical venue, you get an error complaining that you're trying to `null` `events.chapter_id`:

```
return await prisma_1.prisma.events.update(
  Null constraint violation on the fields: (`chapter_id`)
```

...many hours later, I finally realised it's because disconnecting the venue requires you to null both `venue_id` and `chapter_id`, only one of which is allowed to be null!

Ooops.

Anyway, the short term fix is just to undo the db shenanigans and pass (but not use) the `chapterId` argument for those resolvers.  Ugly, but it should work.  The long term fix may be to live with it :man_shrugging: 

Oh, we should probably test each venue type, just prevent regressions, but that's not desperately urgent.

PR #1447, opened 2022-08-24
fix: allow venues to be changed to online

ojeytonwilliams:
Having event's fk for venue be composite meant that the query to remove
that reference would try and null both keys (chapter_id and venue_id),
but chapter_id is not nullable.

This was a bit of a hack to require chapterId to be sent as a
GraphQL variable (so that authChecker could see it), this just opts for
the lesser of two hacks. Namely, passing, but not using, chapterId.

Related to https://github.com/freeCodeCamp/chapter/issues/1316, but it would be best to close that after adding some tests to prevent regressions.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -83,7 +83,7 @@
             "deprecationReason": null
           },
           {
-            "name": "chatUrl",
+            "name": "chat_url",
             "description": null,
             "args": [],
             "type": {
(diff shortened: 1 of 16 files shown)


Sboonny (approved):
I didn't see or catch any issue 👍

ojeytonwilliams:
Cheers, @Sboonny. Hopefully this is a little more reliable.

PR #1451, opened 2022-08-24
feat: add csrf protection

### LINKED ISSUE #1375: CSRF protection
NextJS does not have CSRF protection in by default, so we should add it.

Since we have an express backend, we can make use of csurf to generate a token and have the client send it back in a CSRF-Token header.

gikf:

Closes #1375

---
- I'm still fairly confused with details here. I tried to follow the implementation from main repository.
- There's still issue with first try to send the post request, as neither cookie, nor header is set at that point.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -16,6 +16,7 @@
     "@chakra-ui/react": "2.2.8",
     "@emotion/react": "11.10.0",
     "@emotion/styled": "11.10.0",
+    "browser-cookies": "1.2.0",
     "chakra-confirm": "1.2.16",
     "chakra-data-table": "1.4.0",
     "chakra-next-link": "1.0.10",
--- a/client/package.json
+++ b/client/package.json
@@ -16,7 +16,6 @@
     "@chakra-ui/react": "2.2.8",
     "@emotion/react": "11.10.0",
     "@emotion/styled": "11.10.0",
-    "browser-cookies": "1.2.0",
     "chakra-confirm": "1.2.16",
     "chakra-data-table": "1.4.0",
     "chakra-next-link": "1.0.10",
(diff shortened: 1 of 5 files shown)


ojeytonwilliams:
This looks right to me.  The only issue I see is that the first request has to be GET, HEAD or OPTIONS so that the protection doesn't apply and the two cookies can get set.  However... I thought that was the case, so I'm not sure where the problem is coming from.

I'll do a bit of digging and see if I can help.

gikf:
@ojeytonwilliams Not sure if you have seen my comment on the issue, Apollo server seem to have built-in some CSRF protection. I don't know though if it would be enough for the needs here https://www.apollographql.com/docs/apollo-server/security/cors/#preventing-cross-site-request-forgery-csrf

ojeytonwilliams:
Thanks for pointing that out.  

What I'm getting from it is that express and Apollo need the same cors configuration, since Apollo overwrites express and may create interesting (read: bad) behaviour.  As well as that, turning on `csrfPrevention` for the Apollo server is a good idea.

Is it enough? Combine with the correct CORS config, I believe so.  My understanding is that it prevents any side-effects from happening from 'simple' requests (i.e. GET) that should not mutate anything... but might because developers could accidentally put a db update call in a query resolver.  It protects us from ourselves (and other libraries).

Sorry for sending of on a bit of an odyssey.  The one positive is that I doubt we'd have noticed the misconfigured cors otherwise!

gikf:
If I understand this correctly it's needed to:
- Add cors options to ApolloServer - `{ credentials: true, origin: allowedOrigins }`.
- Remove the manual cookie/token csrf.
- Add `csrfProtection` to ApolloServer.

Let me know if that's right

ojeytonwilliams:
@gikf yep, that's exactly right.  That should protect the /graphql endpoint and the rest we can take care of on a case-by-case basis.

ojeytonwilliams (approved):
Perfect, that should do the trick!

PR #1476, opened 2022-08-30
fix: events loading inconsistency

gikf:

---
- Fixes double load on opening `/events`, one coming from the normal query and one from `fetchMore` after `currentPage` is changed/set.
- Fetching more is now limited to not visited pages numbers. Otherwise page with events is loaded from cache (I assume it's nextjs' cache). This might be a little overkill for now, both in doing so in first place and in using set to keep track of visited pages.

### DIFF
--- a/client/src/modules/events/pages/eventsPage.tsx
+++ b/client/src/modules/events/pages/eventsPage.tsx
@@ -54,7 +54,7 @@ function Pagination({
 const pageSize = 5;
 export const EventsPage: NextPage = () => {
   const [currentPage, setCurrentPage] = useState(1);
-  const [visitedPages, setVisitedPages] = useState(new Set());
+  const [visitedPages, setVisitedPages] = useState(new Set([1]));
   const { loading, error, data, fetchMore } = usePaginatedEventsWithTotalQuery({
     variables: { offset: (currentPage - 1) * pageSize, limit: pageSize },
   });
--- a/client/src/modules/events/pages/eventsPage.tsx
+++ b/client/src/modules/events/pages/eventsPage.tsx
@@ -55,17 +55,15 @@ const pageSize = 5;
 export const EventsPage: NextPage = () => {
   const [currentPage, setCurrentPage] = useState(1);
   const [visitedPages, setVisitedPages] = useState(new Set([1]));
+  const offset = (currentPage - 1) * pageSize;
   const { loading, error, data, fetchMore } = usePaginatedEventsWithTotalQuery({
-    variables: { offset: (currentPage - 1) * pageSize, limit: pageSize },
+    variables: { offset, limit: pageSize },
   });
 
   useEffect(() => {
     if (visitedPages.has(currentPage)) return;
     fetchMore({
-      variables: {
-        offset: data?.paginatedEventsWithTotal.events.length || 0,
-        limit: pageSize,
-      },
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (commented):
Just to confirm: is the goal to prevent the client repeatedly requesting pages it has already fetched? e.g. if open events and click back and forth between pages 1 and 2 the browser should only make two GraphQL requests.

gikf:
> Just to confirm: is the goal to prevent the client repeatedly requesting pages it has already fetched? e.g. if open events and click back and forth between pages 1 and 2 the browser should only make two GraphQL requests.

As an additional effect - yes. The main goal was to stop loading, fetching and rendering twice when `/events` is opened for the first time.

ojeytonwilliams:
> As an additional effect - yes. The main goal was to stop loading, fetching and rendering twice when /events is opened for the first time.

I only see the one graphql request when I reload events.  Either way, I think this PR makes sense.

gikf:
Well, this is awkward. Scratch the fetching then please.

I've double checked how to reproduce the weird behavior that made us all gather here:
- Open home page.
- Navigate to _Events_ from _Menu_.
- Navigate to home page.
- Navigate to _Events_ from _Menu_. (Additionally this step often gets stuck with _Loading..._, in such case circling back home page -> Events would get it unstuck).
- List with events are displayed (that would be cache I believe now), but then _Loading..._ appears, and fresh list with events is fetched.

I've also found that's not fully resolved. I'll investigate further, but fresh eyes could spot something obvious.

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -54,17 +54,20 @@ function Pagination({
 const pageSize = 5;
 export const EventsPage: NextPage = () => {
   const [currentPage, setCurrentPage] = useState(1);
+  const [visitedPages, setVisitedPages] = useState(new Set([1]));
```
ojeytonwilliams, line 57:
```suggestion
  const [visitedPages, setVisitedPages] = useState(new Set([1]));
  const offset = (currentPage - 1) * pageSize;
```

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -54,17 +54,20 @@ function Pagination({
...
   const { loading, error, data, fetchMore } = usePaginatedEventsWithTotalQuery({
     variables: { offset: (currentPage - 1) * pageSize, limit: pageSize },
```
ojeytonwilliams, line 59:
```suggestion
    variables: { offset, limit: pageSize },
```

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -54,17 +54,20 @@ function Pagination({
...
   });
 
   useEffect(() => {
+    if (visitedPages.has(currentPage)) return;
     fetchMore({
       variables: {
         offset: data?.paginatedEventsWithTotal.events.length || 0,
```
ojeytonwilliams, line 66:
```suggestion
        offset,
```

ojeytonwilliams (changes requested):
I like what you've done to reduce queries, but it got me thinking about why it was getting stuck on loading.  It seems like the incorrect offset was confusing it.

ojeytonwilliams (approved):
LGTM. Thanks!

PR #1481, opened 2022-08-31
fix: delay rendering until url param is loaded

gikf:

---
- When page with url param is opened, url param isn't loaded instantly, but at least one render happens when param is `-1`. Query for something with `-1` usually ends up with error (or error-like state, when `!loading`, but also no proper `data` - `!data`). That change `loading -> !data/error` is usually reflected on page with some kind of _error_, which quickly disappears, when url param is finally loaded. Query goes back to the `loading`, and after getting the correct data rendering it.
- Change uses the previously added `isReady`, to render after url param is loaded, what limits the irrelevant quick state changes visible on the page.
- I hope I changed it in all relevant files.
- Unifying when and how `loading`, `error`, etc. are rendered could be a good idea, also on other pages. Currently almost every page does it in own, slightly different, way.

### DIFF
--- a/client/src/modules/chapters/pages/chapterPage.tsx
+++ b/client/src/modules/chapters/pages/chapterPage.tsx
@@ -27,7 +27,7 @@ import {
 import { useParam } from 'hooks/useParam';
 
 export const ChapterPage: NextPage = () => {
-  const { param: chapterId } = useParam('chapterId');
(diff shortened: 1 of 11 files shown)


Sboonny (approved):

ojeytonwilliams (approved):
LGTM :+1: 

Cheers, @gikf

Issue #1371, opened 2022-08-10
User friendly errors and logging


ojeytonwilliams:
###  User facing errors

Currently, when an error is thrown by the server, that error is sent to the client and displayed

On the client, this looks like:

![image](https://user-images.githubusercontent.com/15801806/183877249-1a5c48f5-a68d-4f25-97d5-797cd9b3387b.png)

when the following error is thrown

```
Invalid `prisma_1.prisma.chapters.findMany()` invocation in /usr/chapter/server/src/controllers/Chapter/resolver.js:23:47 20 const inputs_1 = require("./inputs"); 21 let ChapterResolver = class ChapterResolver { 22 async chapters() { → 23 return await prisma_1.prisma.chapters.findMany() The table `public.chapters` does not exist in the current database.
```

This is fine in development, but if `NODE_ENV=production` the user should see something more appropriate.  Ideally a genuinely useful message, but a generic "Something went wrong" message should be fine for the MVP.

### Developer facing errors

The other side of this is that no errors appear in the server logs, making debugging difficult.  It would make sense to integrate Sentry, but the main thing is that unexpected events (like the above failure) should be recorded somewhere.

MVP:

- [ ] Hide error details from users in production

Nice to have:

- [ ] Log server errors both in development and production
- [ ] Integrate with Sentry

Original discussion https://github.com/freeCodeCamp/chapter/issues/412

Sboonny:
Just wondering, about the possibility to experiment with new services like.

- For the backend error handling: https://www.axiom.co/
- For session replay in client side using: https://www.highlight.io/

ojeytonwilliams:
@Sboonny as much as possible I'd like to use the same tools as the main repo. We could discuss migrating with Mrugesh, but I don't think we need to experiment with unfamiliar tooling for the MVP.

Sboonny:
For documentation: one of the errors to handle in the middle is

- Google calendar integration 

An attempt to contact Google calendar's API fails (e.g. when creating an event), it should show a server error

PR #1490, opened 2022-09-04
chore: friendly message in production

Sboonny:

Ref #1371

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -16,7 +16,6 @@ import {
   user,
   events,
   handleError,
-  // handleAuthenticationError,
 } from './controllers/Auth/middleware';
 import { checkJwt } from './controllers/Auth/check-jwt';
 import { prisma } from './prisma';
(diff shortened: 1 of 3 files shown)


server/src/controllers/Auth/middleware.ts
```diff
@@ -130,3 +130,19 @@ export function handleAuthenticationError(
...
+  err: any,
+  _req: Request,
+  res: Response,
+  next: NextFunction,
+) {
+  if (res.headersSent) {
+    return next(err);
+  }
+  if (err) {
+    return res.status(401).send({
```
ojeytonwilliams, line 144:
```suggestion
    return res.status(500).send({
```
It's a server error after all.

ojeytonwilliams (changes requested):
Do you want to handle the client side in this PR?  It's totally fine to create a separate one.  Also, could you limit this to when `isProd()` is true? In development and testing we definitely want to see the full error messages.

Finally, I think this is a good PR to clear out the handleAuthenticationError code.  I should have deleted it in the first place, sorry!

server/src/app.ts
```diff
@@ -15,6 +15,7 @@ import { resolvers } from './controllers';
 import {
   user,
   events,
+  handleError,
   // handleAuthenticationError,
```
ojeytonwilliams, line 19:
```suggestion
```

ojeytonwilliams (approved):

PR #1491, opened 2022-09-04
Create/squence reminder

### LINKED ISSUE #275: UI: As a future participant after the event, I will automatically get emails notifying me of subsequent events.

Sboonny:

Closes #275 

I have opened this for draft, because I don't know if I am going the right track.

- I am trying to copy reminder, but instead of sending the email before the event, it will send the email after the event.

Because of labels on the issue, I am wondering if I am going the right track? or there is a simpler way to do this

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,8 @@
     "pretty": "prettier --write client/**/*.ts* server/**/*.ts",
     "prepare": "is-docker || husky install",
     "prisma": "npx -w=server dotenv -e ../.env -- prisma",
-    "reminders:send": "npm -w=server run reminders:send"
+    "reminders:send": "npm -w=server run reminders:send",
+    "subreminders:send": "npm -w=server run subreminders:send"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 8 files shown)


gikf:
Doesn't #275 just need adjusting text of the email send when new event is created?

Sboonny:
I thought, it was for reminder after attending an event, not when creating new Events. 

this makes things easier 🤣. Thank you so much

gikf:
It's written in a bit weird way, but I believe the main point is to get those who attended the event (that means chapter users) information about next events in chapter. And doing that at the event creation greatly simplifies things.

Sboonny:
I am having difficulty thinking of a logical way to pick the recently created event.

Should I create a copy of the code below, but it always picks the first event 
https://github.com/freeCodeCamp/chapter/blob/79a68f1aaf3382e56157dd5fa4fe2c99a7c1292e/server/src/services/Reminders.ts#L68-L80

Something like this [attempt](https://github.com/freeCodeCamp/chapter/compare/main...Sboonny:chapter:chore/reminder-on-event-creation)

gikf:
That's not really needed, when event is created emails are send with `useSendEventInviteMutation` hook, and its resolver - https://github.com/freeCodeCamp/chapter/blob/main/server/src/controllers/Events/resolver.ts#L743

PR #1509, opened 2022-09-07
fix: ensure event start is not after end

### LINKED ISSUE #1508: Events can have a nonsense period by ending before it starts
When you go to `/dashboard/events/1/edit`, you can edit the event to start and end in nonsense period. Example:

image

I don't think someone will actively try this while creating a realistic event, so we can discuss what do to Post MVP.

gikf:

Closes #1508

---
- Prevents event start and end times mismatch.
  - When _Start at_ is changed to be after _End at_, the _End at_ is set to _Start at_ as well.
  - When _End at_ is changed to be before _Start at_, the _Start at_ is set to _End at_ as well.


Sboonny (approved):

PR #1516, opened 2022-09-09
refactor: remove old auth (JWT) code

ojeytonwilliams:


Sboonny (approved):
is `express-oauth2-jwt-bearer` needed? 
reading the code it checks for Jwt, but there is no more JWT with this pull request

https://github.com/ojeytonwilliams/chapter/blob/2ec97d24c1909fa4fda7f3f71d50b33664a9a1dd/server/src/app.ts#L61-L96

gikf (approved):

ojeytonwilliams:
> is `express-oauth2-jwt-bearer` needed? reading the code it checks for Jwt, but there is no more JWT with this pull request

It is, yeah.  We don't directly provide the JWT, but one comes back from Auth0 after the user completes the authentication flow.  We send that token to the api to login/register the user. The api needs to know if this is a valid Auth0 token and that's where this library comes in.

PR #1517, opened 2022-09-09
chore: cancel event email content

Sboonny:

Ref #100

Currently, It isn't possible to run cypress locally, and it doesn't work on Gitpod, so I will delay adding test until then.

How it looks

image

the email content 

```md
The Upcoming Event for Heidenreich, Koch and Kreiger is canceled.

View Upcoming Events for Abshire - Barrows: [Abshire - Barrows chapter](https://3000-sboonny-chapter-0ge8coe9jt5.ws-eu64.gitpod.io/chapters/4).
----------------------------

- Stop receiving upcoming event notifications for Abshire - Barrows. you can do it here: https://3000-sboonny-chapter-0ge8coe9jt5.ws-eu64.gitpod.io/events/15?emaillink=true.
- More about Abshire - Barrows or to unfollow this chapter: https://3000-sboonny-chapter-0ge8coe9jt5.ws-eu64.gitpod.io/chapters/4.

----------------------------
You received this email because you subscribed to Heidenreich, Koch and Kreiger Event.

See the options above to change your notifications.
```


server/src/controllers/Events/resolver.ts
```diff
@@ -710,17 +711,31 @@ ${venueDetails}`;
...
+    const chapterURL = `${process.env.CLIENT_LOCATION}/chapters/${event.chapter.id}`;
+    const eventURL = `${process.env.CLIENT_LOCATION}/events/${event.id}?emaillink=true`;
     const notCanceledRsvps = event.event_users;
 
     if (notCanceledRsvps.length) {
       const emailList = notCanceledRsvps.map(({ user }) => user.email);
-      const subject = `Event ${event.name} canceled`;
-      const body = `The event ${event.name} was canceled`;
+      const subject = `Event ${event.name} is canceled`;
+      const cancelEventEmail = `The Upcoming Event for ${event.name} is canceled.<br />
```
gikf, line 721:
```suggestion
      const cancelEventEmail = `The Upcoming Event ${event.name} is canceled.<br />
```
With _for_ it looks like ${event.name} is supposed to be the chapter instead.

server/src/controllers/Events/resolver.ts
```diff
@@ -710,17 +711,31 @@ ${venueDetails}`;
...
+      <br />
+      View Upcoming Events for ${event.chapter.name}: <a href='${chapterURL}'>${event.chapter.name} chapter</a>.<br />
+      ----------------------------<br />
+      <br />
+      - Stop receiving upcoming event notifications for ${event.chapter.name}. you can do it here: <a href="${eventURL}">${eventURL}</a>.<br />
```
gikf, line 726: From event page, only further event notifications for that event can be changed. Seems that either wording or url needs to be changed.

Sboonny: I have updated it to this

```md
- Stop receiving event notifications for ${event.name}. You can do it here: <a href="${eventURL}">${eventURL}</a>.<br />
```

But it maybe redundant, should we delete that line?

gikf:
Out of curiosity, what's the _join and leave function_ mentioned in comment?

Sboonny:
In issue #274, there is this a mentioning of join button in this [comment](https://github.com/freeCodeCamp/chapter/issues/274#issuecomment-565740617). It should act as indicator that you are planning to attend the Event, as far as I know

So Event Card could have a join button

gikf:
That sounds like joining and leaving chapter. Joining chapter is implemented as a standalone action and also is implicitly performed when user is RSVPing for an event from not joined chapter.

Leaving chapter is not implemented.

server/src/controllers/Events/resolver.ts
```diff
@@ -854,17 +854,34 @@ ${unsubscribeOptions}`,
...
+      View Upcoming Events for ${event.chapter.name}: <a href='${chapterURL}'>${event.chapter.name} chapter</a>.<br />
+      ----------------------------<br />
+      <br />
+      - Stop receiving event notifications for ${event.name}. You can do it here: <a href="${eventURL}">${eventURL}</a>.<br />
+      - More about ${event.chapter.name} or to unfollow this chapter: <a href="${chapterURL}">${chapterURL}</a>.<br />
+      <br />
+      ----------------------------<br />
+      You received this email because you Subscribed to ${event.name} Event.<br />
+      <br />
+      See the options above to change your notifications.
```
ojeytonwilliams, line 878:
```suggestion
      const cancelEventEmail = `The upcoming event ${event.name} has been canceled.<br />
      <br />
      View upcoming events for ${event.chapter.name}: <a href='${chapterURL}'>${event.chapter.name} chapter</a>.<br />
      ----------------------------<br />
```
I think this flows better. I deleted the notifications text because it should be possible to DRY it out.  

We have two types of emails

- Ones sent in response to a user action
- Ones sent because a user is subscribed

All emails should have the same unsubscribe text

```
To manage notifications about this event, go to <a href="${eventURL}">${eventURL}</a>.<br />
To manage notifications from this chapter about new events, go to <a href="${chapterURL}">${chapterURL}</a>.<br />
```

Emails that are sent because the user is subscribed to an event should explain that.  Similarly emails that are sent because of a chapter subscription should explain that.

Finally emails that are sent in response to a user action need no explanation.

@Sboonny could you handle this in this PR?  I think makes sense to handle them all together, so we can make sure they're consistent.

ojeytonwilliams (changes requested):

server/src/controllers/Events/resolver.ts
Sboonny: I have tried to DRY it using `unsubscribeOption`, but the resolver didn't use `user.id`. 

So before I add them, do we need `user.id` in cancel events email? although it doesn't make sense to cancel cancel events, people may want to cancel their chapter membership.

So do we add it? or just link it?

ojeytonwilliams: It seems simplest to always provide both of the unsubscribe links. We don't need to customise them for the users current subscription state.

Unfollowing cancelled events should mostly be pointless, but conceivably events could be uncancelled. It's safer to let users unsubscribe and respect their wishes.

Sboonny: Sure, random question 😅, the https://github.com/freeCodeCamp/chapter/commit/06209830d3c450a4d4096bfdf2d3cbf61cb47e11 should have failed cypress because there were no `user.id`, but it succeeded should. should we worry about that?

ojeytonwilliams: Given that the tests didn't run, I'm not worried!

server/src/controllers/Events/resolver.ts
```diff
@@ -795,17 +772,32 @@ ${unsubscribeOptions}`,
     });
     await deleteEventReminders(id);
 
+    const chapterURL = `${process.env.CLIENT_LOCATION}/chapters/${event.chapter.id}`;
     const notCanceledRsvps = event.event_users;
-
+    const unsubScribeOptions = getUnsubscribeOptions({
+      chapterId: event.chapter_id,
+      eventId: event.id,
+      userId: ctx.user.id,
```
Sboonny, line 780: wait because the one who activating this function is instance owner, does that mean if someone tries to unsubscribe, they will use instance owner id?

server/src/controllers/Events/resolver.ts
Sboonny: sorry for the confusion, the function meant to run without user.id, which is interesting 🤔

server/src/controllers/Events/resolver.ts
ojeytonwilliams: If the instance owner makes the request, yes. Otherwise, no.

server/src/controllers/Events/resolver.ts
```diff
@@ -795,17 +770,32 @@ ${unsubscribeOptions}`,
     });
     await deleteEventReminders(id);
 
+    const chapterURL = `${process.env.CLIENT_LOCATION}/chapters/${event.chapter.id}`;
+    const eventURL = `${process.env.CLIENT_LOCATION}/events/${event.id}`;
```
Sboonny, line 774: the test is fallin because these two don't have unsubscribe in them, I think I need to add string that activate unsubscribe commands 🤔

gikf: :thinking: This doesn't seem related to that test.

I think the `getUnsubscribeOptions` is returning tokens wrapped in an `<a>` instead of proper links.

server/src/controllers/Events/resolver.ts
```diff
@@ -247,11 +223,10 @@ const rsvpNotifyAdministrators = async (
...
-      const chapterUnsubscribeToken = generateToken(
-        UnsubscribeType.Chapter,
-        chapter_id,
-        user.id,
-      );
+      const chapterUnsubscribeToken = getChapterUnsubscribeToken({
+        chapterId: chapter_id,
+        userId: user.id,
+      });
       const text = `${body}<br><a href="${process.env.CLIENT_LOCATION}/unsubscribe?token=${chapterUnsubscribeToken}Unsubscribe from chapter emails`;
```
ojeytonwilliams, line 230: This can use `getChapterUnsubscribeOptions`.

server/src/controllers/Events/resolver.ts
```diff
@@ -58,6 +54,11 @@ import {
 } from '../../util/calendar';
 import { updateWaitlistForUserRemoval } from '../../util/waitlist';
 import { redactSecrets } from '../../util/redact-secrets';
+import {
+  getChapterUnsubscribeToken,
+  getUnsubscribeOptions,
+  NotificationContextText,
```
ojeytonwilliams, line 60:
```suggestion
  getChapterUnsubscribeOptions,
  getEventUnsubscribeOptions,
```
We shouldn't be exporting helper functions from event email, just the final two functions that give us the text we want.

ojeytonwilliams (changes requested):
Hey @Sboonny I hope the suggestions make sense, but give me a shout if they don't.

server/src/controllers/Events/resolver.ts
```diff
@@ -823,17 +798,32 @@ ${unsubscribeOptions}`,
...
     await deleteEventReminders(id);
 
+    const chapterURL = `${process.env.CLIENT_LOCATION}/chapters/${event.chapter.id}`;
+    const eventURL = `${process.env.CLIENT_LOCATION}/events/${event.id}`;
     const notCanceledRsvps = event.event_users;
-
+    const unsubScribeOptions = NotificationContextText({
+      linkToEvent: eventURL,
+      linkToChapter: chapterURL,
+    });
```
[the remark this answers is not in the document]
Sboonny: This is complicated because this cancel RSVP without userId, so if we are using `getEventUnsubscibeOptions`, we will force user to log in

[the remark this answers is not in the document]
Sboonny: Should we move the text to the resolver with comment of why we are doing this?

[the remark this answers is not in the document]
ojeytonwilliams: We can't force users to login when they want to unsubscribe from email notifications.  Fortunately, there's no need to.  We have access to the userId via `event_users`.  Since each email is different (different unsubscribe tokens) you can use `batchSender` to send them all.  There's an example of that in `sendEventInvite`.

server/src/controllers/Events/resolver.ts
```diff
@@ -822,18 +796,30 @@ ${unsubscribeOptions}`,
       },
     });
     await deleteEventReminders(id);
-
     const notCanceledRsvps = event.event_users;
 
-    if (notCanceledRsvps.length) {
+    if (!notCanceledRsvps) return null;
```
Sboonny, line 801: I don't like the current logic, I think this won't cancel the event if there is no members in the events

Sboonny: Should we do something if there are no Members in Events

Sboonny: Nope event without members can be canceled, false alarm 👍

ojeytonwilliams: The fact you could delete an entire mutation and everything passed suggests we have a bit of a testing gap!

Sboonny: whelp, I have made it fail with `!notCanceledRsvps.length`, I am openning an issue for that 🤣

PR #1521, opened 2022-09-10
chore: policy page

Sboonny:


allella:
Thanks for creating these.

I would suggest removing the commented out HTML.  I plan to reword some things once it's merged in and we can reference the fCC pages if we need to see that content.

`/* commeted these because we don&apost have way to do them, yet.`

ojeytonwilliams (commented):
The redirect works just fine. When we have Chapter specific text, this should be fine for the MVP.

client/src/pages/policy/[policyId].tsx
```diff
@@ -0,0 +1,3 @@
+import { PolicyPage } from 'modules/policy';
```
Comment (Sboonny) [inline client/src/pages/policy/[policyId].tsx:1]:
```suggestion
import { PolicyPage } from '../../modules/policy';

File: client/src/pages/policy/index.tsx [hunk 2]
```diff
@@ -0,0 +1,3 @@
+import { PolicyPage } from 'modules/policy';
```
Sboonny, line 1:
```suggestion
import { PolicyPage } from '../../modules/policy';

File: client/src/pages/policy/[policyId].tsx [hunk 3]
```diff
@@ -0,0 +1,3 @@
+import { PolicyPage } from '../../modules/policy';
+
+export default PolicyPage;
```
Comment (gikf) [inline client/src/pages/policy/[policyId].tsx:3]: This is regarding the file itself, not the code. Is there need for a dynamic route `/policy/{whatever}`?

gikf (approved):
Unless text needs to be changed in this PR, this looks good to me :+1:

allella:
Go ahead and I'll take a look at the text once it's merged in and open a PR with any tweaks. Thanks

PR #1537, opened 2022-09-14
refactor: use single query in resolver

gikf:

---
- Use single query, instead of two.
- Remove not needed `ctx` argument. Being `@Authorized` by default means user is logged in.


ojeytonwilliams (approved):
Nice one, this is much cleaner.

PR #1541, opened 2022-09-15
feat: refresh google tokens automatically

ojeytonwilliams:

If the OAuth2Client is given the `access_token`, `refresh_token` and `expiry_date` it will automatically fetch new `access_token`s as they expire.

### DIFF
--- a/server/src/services/Google.ts
+++ b/server/src/services/Google.ts
@@ -82,8 +82,7 @@ async function onTokens(tokens: Credentials) {
 
   let userInfo;
   try {
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    userInfo = await createOAuth2Client().getTokenInfo(tokens.access_token!);
+    userInfo = await createOAuth2Client().getTokenInfo(access_token);
   } catch {
     throw new Error('Failed to get user info');
   }
--- a/server/src/services/Google.ts
+++ b/server/src/services/Google.ts
@@ -52,67 +52,77 @@ export function getGoogleAuthUrl(state: string) {
   });
 }
 
-// TODO: skip calling this if the user already has a valid, non-expired token
-export async function storeGoogleTokens(code: string) {
+export async function requestTokens(code: string) {
   const oauth2Client = createOAuth2Client();
 
-  let tokens, userInfo;
   try {
-    const tokenRes = await oauth2Client.getToken(code);
-    tokens = tokenRes.tokens;
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    userInfo = await oauth2Client.getTokenInfo(tokens.access_token!);
+    await oauth2Client.getToken(code);
   } catch {
     throw new Error('Failed to get tokens');
-  }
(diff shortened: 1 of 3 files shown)


server/src/services/Google.ts
```diff
@@ -97,35 +114,37 @@ export async function storeGoogleTokens(code: string) {
...
-  oauth2Client.setCredentials({ access_token });
+  const oauth2Client = createOAuth2Client().on('tokens', onTokens);
+
+  const tokenInfo = await prisma.google_tokens.findFirstOrThrow();
+  const tokens = {
+    access_token: tokenInfo.access_token,
+    refresh_token: tokenInfo.refresh_token,
+    // Awkwardly, Prisma is setting the type to bigint, but it's really an 8 byte
+    // integer in the database.
+    expiry_date: tokenInfo.expiry_date as unknown as number,
```
gikf, line 136: This might be weird question, but technically isn't that correct, since type in schema is changed to `BigInt`? Is schema change required?

server/src/services/Google.ts
```diff
@@ -52,25 +52,42 @@ export function getGoogleAuthUrl(state: string) {
...
+async function onTokens(tokens: Credentials) {
   // TODO: handle the case where the user rejects some or all of the scopes
   const { access_token, refresh_token, expiry_date } = tokens;
 
   if (!access_token || !expiry_date) throw new Error('Tokens invalid');
 
+  let userInfo;
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    userInfo = await createOAuth2Client().getTokenInfo(tokens.access_token!);
```
gikf, line 86:
```suggestion
    userInfo = await createOAuth2Client().getTokenInfo(access_token);
```

server/src/services/Google.ts
ojeytonwilliams: No, it's a good question.  Prisma's [BigInt](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#bigint) corresponds to a Postgres [bigint](https://www.postgresql.org/docs/current/datatype-numeric.html) which is an 8 byte integer.  JavaScript's integers only go up to `2^53 -1` while Postgres can handle `2^63-1`.

...and the `number` coming back from Google is genuinely a JS number, since it's a unix timestamp (but in milliseconds). So, that easily fits in `2^53-1` hence Google giving it type `number`.

If that makes sense, I'll update the comment to explain this better.

gikf: I somehow assumed timestamp will fit into the Postgres' _integer_. All is clear :+1:

ojeytonwilliams: You and me both. It didn't occur to me until I realised that the fact everything was expiring in 1970 was because of integer overflow...

ojeytonwilliams:
These tests really have gotten flaky :/  I'll see what I can do.

gikf (approved):
LGTM :+1:

As flakiness is not related to changes here, lets merge.

PR #1553, opened 2022-09-18
feat: truncate website text

gikf:

---
- Replaces website-link slicing with truncating text.
- `noOfLines === 1` adds the ellipsis when text exceeds the `maxWidth`.
- `maxWidth` is the attribute that can adjust how long text can be before truncating. `sm` value can be considered as a placeholder, if somebody has specific idea what put there instead.
- `wordBreak === 'break-all'` can break text on every character, while ie. `break-word` will break only on full words. In here that point will be the start of truncating.


Sboonny (approved):
I had no idea about `noOfLines`, thank you for this

PR #1554, opened 2022-09-18
fix: non-unique iteration key

gikf:

---
- Changes non-unique iteration key, which was causing console warning.

<details><summary>console warning</summary>

>Warning: Encountered two children with the same key, `no`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

</details>


Sboonny (approved):

PR #1556, opened 2022-09-18
test: change event venue deflaky

gikf:

---
- Test often failed because save button wasn't enabled when clicked. This suggests issue existing earlier, with dropdown not being recognized as dirtied.
- Change tries to ensure that dropdown is interacted with when venues are loaded, what would make save button enabled.

### DIFF
--- a/cypress/e2e/dashboard/events/events-index.cy.ts
+++ b/cypress/e2e/dashboard/events/events-index.cy.ts
@@ -153,7 +153,14 @@ describe('spec needing owner', () => {
     cy.createEvent(1, eventData).then((response) => {
       const eventId = response.body.data.createEvent.id;
       cy.visit(`/dashboard/events/${eventId}/edit`);
-      cy.findByRole('combobox', { name: 'Venue' })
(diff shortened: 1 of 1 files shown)


cypress/e2e/dashboard/events/events-index.cy.ts
```diff
@@ -163,6 +170,7 @@ describe('spec needing owner', () => {
         .findByRole('button', {
           name: 'Save Event Changes',
         })
+        .should('be.enabled')
```
ojeytonwilliams, line 173: Could it be enough to just check if it's enabled?  It's possible that it's failing because Cypress is clicking too quickly, before the button has a chance to become enabled.

gikf: I dug a little more and initially it looked like that was _unfortunately no_. I've even added check what was the actual value selected at the end. Then an anomaly caught my eye - selected value was `3`. All this time we were being fooled by `.select` method, when passed integer, it will select option by the index, not the value.

ojeytonwilliams (approved):
Nice catch!

LGTM :+1:

PR #1548, opened 2022-09-16
fix: stop dev login happening automatically

ojeytonwilliams:
- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `main` branch of Chapter.

Closes https://github.com/freeCodeCamp/chapter/issues/1547

PR #1557, opened 2022-09-19
test: use deletable venue for deletion

gikf:

---
- Changes venue used in test to one that can be deleted (not used in any event).
- Technically the venue shouldn't matter, as test should be stopped at the authorization. When that doesn't happen though, test can result with confusing error (see below), if the venue is still used in some event. This is to reduce that confusion when things don't work.
- Relevant test still can fail until #1548 makes logout more consistent.

```
Invalid prisma_1.prisma.venues.delete() invocation in
/usr/chapter/server/src/controllers/Venue/resolver.js:61:51
return await prisma_1.prisma.venues.delete(
Foreign key constraint failed on the field: events_venue_id_fkey (index) to equal **Access denied! You don't have permission for this action!**
```

### DIFF
--- a/cypress/e2e/dashboard/venues.cy.ts
+++ b/cypress/e2e/dashboard/venues.cy.ts
@@ -89,50 +89,66 @@ describe('venues dashboard', () => {
     const venueCreateVariables = {
       chapterId: 1,
     };
-    const venueUpdateDeleteVariables = {
+    const venueUpdateVariables = {
       chapterId: 1,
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (approved):
LGTM :+1: 

At some point we'll have to handle deletion of a venues which are used by events, but not today!

gikf:
Lets do it after deleting venues is added to UI :grin:.

PR #1559, opened 2022-09-19
chore: add custom error message

gikf:

---
- Adds custom error message for when (rejected) response doesn't contain one error. It gives a bit of context when test fails, without the need to digging deeper into test. Goal is to improve over the default message:
>Target cannot be null or undefined.


Sboonny (approved):

PR #1561, opened 2022-09-19
fix: use stable unique key

ojeytonwilliams:

https://github.com/freeCodeCamp/chapter/pull/1554 does suppress the warning, but using an index as a key can result in weirdness and is considered to be an anti-pattern: https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/


gikf (approved):
That completely makes sense :+1: 

Also thanks for the article.

PR #1562, opened 2022-09-19
chore: remove unrequired inputs and polish some style

Sboonny:

I created a branch to upstream by mistake, do I close this, open a new pull request and delete the upstream branch?

New: 
image

Old: 
image

--- 

In small screen there is an overflow because of the buttons, I have no Idea how to fix, I am open for suggestion 

Note: breaking the words (if possible) will create spacing issues

image

---

Create Event Page 

image


client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -260,7 +265,7 @@ const EventForm: React.FC<EventFormProps> = (props) => {
           Invite only
         </Checkbox>
 
-        <FormControl isRequired>
+        <FormControl marginBlock={'1em'}>
```
ojeytonwilliams, line 268:
```suggestion
        <FormControl isRequired marginBlock={'1em'}>
```
I know it doesn't make much difference, since one is always selected, but since this is required, we should keep that explicit.

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -283,7 +288,7 @@ const EventForm: React.FC<EventFormProps> = (props) => {
             <h1>Error loading venues</h1>
           ) : (
             isPhysical(getValues('venue_type')) && (
-              <FormControl isRequired>
+              <FormControl marginBlock={'1em'}>
```
ojeytonwilliams, line 291:
```suggestion
              <FormControl isRequired marginBlock={'1em'}>
```
Same principle as above - this one is required.

ojeytonwilliams (changes requested):
Looking good, just a couple of minor things:

ojeytonwilliams (approved):
LGTM 👍

gikf:
A bit late, but... for event required fields changes:
- server throws when there's not `url`
- `streaming_url` should be probably always required when venue has online type.
- `image_url` - is there intended to be some default image? Lack of image (as not required now) ends up with blank broken image

PR #1563, opened 2022-09-19
feat: display toast after sending email

gikf:

---
- Adds toast after email(s) are send. This also changes to `await` for emails in client.
- Tries to lower flakiness of the relevant test.
- Changes test to check sent emails only after toast is displayed. Test was often failing with incorrect number of emails - `cy.waitUntilMail` was too fast and was getting lower number of emails than expected.
- Some alternative could be changing `cy.waitUntilMail` to keep waiting when number of emails caught by mailhog changes. Or more explicitly wait for a specific number of emails.
- Slightly unrelated, disables button when emails are being sent.


ojeytonwilliams (approved):
LGTM :+1: 

Regarding the flakiness of `waitUntilMail`, one thing that could work is to wait until the expected number of emails has been sent (if we know it).  Otherwise we're always going to have one race condition or another.

PR #1567, opened 2022-09-20
chore: create user profile

Sboonny:

In this attempt, I won't try to add the ability to get more data, but will focus on deletion function.


client/src/modules/profiles/pages/userProfile.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { Flex, Heading, Text, Link } from '@chakra-ui/layout';
+import { useConfirmDelete } from 'chakra-confirm';
+import { useRouter } from 'next/router';
+import { Button } from '@chakra-ui/button';
+import { useDeleteMeMutation } from 'generated/graphql';
+import { useAuth } from 'modules/auth/store';
```
ojeytonwilliams, line 7:
```suggestion
import { useAuthStore } from 'modules/auth/store';
```
By using the store we can update the stored user after deleteMe resolves.

client/src/modules/profiles/pages/userProfile.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Users } from 'modules/dashboard/Users/graphql/queries';
+
+export const UserProfilePage = () => {
+  const { user } = useAuth();
```
ojeytonwilliams, line 11:
```suggestion
  const {
    data: { user },
    setData,
  } = useAuthStore();
```

client/src/modules/profiles/pages/userProfile.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const confirmDelete = useConfirmDelete({ doubleConfirm: true });
+  const [deleteMe] = useDeleteMeMutation({
+    refetchQueries: [{ query: Users }],
+  });
+  const clickDelete = async () => {
+    const ok = await confirmDelete();
+    if (!ok) return;
+    deleteMe({ variables: { userId } });
+    router.push('/');
```
ojeytonwilliams, line 27:
```suggestion
    setData({});
    router.push('/');
```
Purge the stored user.

ojeytonwilliams (changes requested):

client/src/modules/auth/graphql/mutations.ts
```diff
@@ -0,0 +1,9 @@
+import { gql } from '@apollo/client';
+
+export const deleteMe = gql`
+  mutation deleteMe($userId: Int!) {
+    deleteMe(id: $userId) {
```
ojeytonwilliams, line 5:
```suggestion
  mutation deleteMe {
    deleteMe {
```
By giving this an id parameter it implies that you can delete a user with that id.  That's not the case, so we should remove the parameter.

server/src/controllers/Auth/resolver.ts
```diff
@@ -1,12 +1,21 @@
...
 @Resolver()
 export class AuthResolver {
   @Query(() => UserWithInstanceRole, { nullable: true })
   async me(@Ctx() ctx: ResolverCtx): Promise<UserWithInstanceRole | null> {
     return ctx.user ?? null;
   }
+
+  @Mutation(() => User)
+  async deleteMe(
+    @Arg('id', () => Int) id: number,
```
ojeytonwilliams, line 16:
```suggestion
```
The queries should not be passing this id.

server/src/controllers/Auth/resolver.ts
```diff
@@ -1,12 +1,21 @@
...
+    @Ctx() ctx: Required<ResolverCtx>,
+  ): Promise<User | undefined> {
+    if (id === ctx.user.id) return await prisma.users.delete({ where: { id } });
```
ojeytonwilliams, line 19:
```suggestion
    return await prisma.users.delete({ where: { id: ctx.user.id } });
```

ojeytonwilliams (changes requested):
I noticed, by chance, that banned users can't delete their accounts.  For the MVP I think we can allow that (even knowing that they can create new accounts to evade bans) and consider a proper solution later down the line. Could you make the `user_bans` cascade like the sessions?

gikf (commented):
Self note for future: remember to apply schema changes before wondering why updated relations do not cascade on deletion :eyes:

ojeytonwilliams (approved):
LGTM :+1:

Issue #1534, opened 2022-09-14
Users should not be able to view dashboard pages without permission


shootermv:
### Describe the bug:

currently, admin of some chapter can see users of different chapter
after navigate there from browser url:
`/dashboard/chapters/2/users`
it will be better to show 'Error: Access denied! You don't have permission for this action!' text instead

PR #1572, opened 2022-09-20
feat: hide dashboards

ojeytonwilliams:

Ref: https://github.com/freeCodeCamp/chapter/issues/1534

Partial progress on requiring permissions to see the dashboard. The main goal was to hide the set of dashboards shown in `all.cy.ts` from the banned user.

Currently the privacy protections are a sham, because the GraphQL server will still return data from the unprotected query resolvers.  We should audit all public routes to make sure nothing private is accessible through them.  However, protecting the views like this does reduce frustration (since the mutations *are* protected and it would be annoying to find that out after filling in the form!)

### DIFF
--- a/client/src/modules/dashboard/Chapters/graphql/queries.ts
+++ b/client/src/modules/dashboard/Chapters/graphql/queries.ts
@@ -8,3 +8,36 @@ export const ChapterRoles = gql`
     }
   }
 `;
+
+export const DASHBOARD_CHAPTER = gql`
+  query dashboardChapter($chapterId: Int!) {
+    dashboardChapter(id: $chapterId) {
+      id
+      name
+      description
+      category
+      city
+      region
+      country
+      image_url
+      chat_url
+      events {
+        id
+        name
+        description
+        start_at
+        invite_only
+        canceled
+        image_url
+        tags {
+          tag {
+            id
+            name
+          }
+        }
+        invite_only
+        canceled
+      }
+    }
+  }
+`;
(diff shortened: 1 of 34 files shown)


Sboonny (approved):
I tried to access without log in and was getting `Access denied! You don't have permission for this action!`, so this is good to go for me.

I won't merge, because I am not skilled enough to review the code 👍

client/src/modules/dashboard/Chapters/graphql/queries.ts
```diff
@@ -8,3 +8,36 @@ export const ChapterRoles = gql`
...
+        canceled
+        image_url
+        tags {
+          tag {
+            id
+            name
+          }
+        }
+        invite_only
+        canceled
```
gikf, line 39:
```suggestion
```
Duplicated

common/permissions.ts
```diff
@@ -1,6 +1,7 @@
 export enum ChapterPermission {
   ChapterEdit = 'chapter-edit',
-  ChapterBanUser = 'ban-user',
+  ChapterBanUser = 'chapter-ban-user',
+  ChapterViewUsers = 'chapter-view-users',
```
gikf, line 4: This added permission isn't used anywhere.

gikf (commented):
I know _Sponsors_ dashboard is hidden from navigating to it without permission, but I've cheated and entered there directly. To see if individual pages (`/dashboard/sponsors/1`, `/dashboard/sponsors/1/edit`) will display access denied like other dashboards. It would be nice to make it similar to others.

PR #1577, opened 2022-09-22
fix: use node 16 (not alpine) in dev for M1

### LINKED ISSUE #1571: MacOS Support
We used to require the docker platform be linux/amd64 for compatibility with Apple Silicon Mac, but this has recently started causing problems in CI. Ref: https://github.com/freeCodeCamp/chapter/pull/1570 and https://github.com/freeCodeCamp/chapter/actions/runs/3089347523/jobs/4996858566

It would be good if we could keep CI reliable and have Chapter work on Macs.

@c-ehrlich any idea how we can keep both running?

ojeytonwilliams:
Until prisma provide more binaries, this seems necessary to support M1
processors

Closes https://github.com/freeCodeCamp/chapter/issues/1571 (hopefully)

### DIFF
--- a/client.Dockerfile
+++ b/client.Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16-alpine as development
+FROM node:16 as development
 WORKDIR /usr/chapter/
 
 FROM development as build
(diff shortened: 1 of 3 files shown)


shootermv:
i tested this pr on my m1 machine. i can see the site running after `docker-compose up` command.
i think it important to also mention in `how to contribute` docs , that you must run `npm run db:sync` and `npm run db:seed` commands to have the data set

PR #1581, opened 2022-09-22
create relation between sponser and its events

Sboonny:


gikf:
Events and sponsors already have many-to-many relation - each event can have multiple sponsors, and each sponsor can be assigned to multiple events. What's the goal here?

Sboonny:
I want to show the event sponsored by a specific sponsor.

for example: 

I sponsored

- hanging out Friday
- movie night Wednesday
- clean energy month

gikf:
That doesn't need any deeper schema changes (maybe just rename of the field in `sponsors` table). The key for it is the `event_sponsors` field in `sponsors` table. On the server side it needs to be added to the `Sponsor` graphql-type, and then the resolver will need adjusting.

If you'd need some example, take a look how `sponsors` or `tags` are included in the `Event` graphql-type. Barring the different names used, I think it will be very similar to what is required to get events sponsored by x.

client/src/modules/dashboard/Sponsors/graphql/queries.ts
```diff
@@ -11,3 +11,22 @@ export const SPONSOR = gql`
...
+    logo_path
+    type
+    event_sponsors {
+      event {
+        id
+        name
+        invite_only
+        canceled
+      }
+    }
```
Sboonny, line 30:
```suggestion
    }
  }

File: client/src/modules/dashboard/Sponsors/pages/SponsorPage.tsx [hunk 2]
```diff
@@ -43,14 +44,20 @@ export const SponsorPage: NextPage = () => {
...
-
-      <h3>Placeholder for events ....</h3>
+      {data.sponsorWithEvents && (
+        <EventsList
+          title={'Sponser'}
+          events={data.sponsorWithEvents.event_sponsors
+            .map((e) => e.events)
+            .flat()}
+        />
+      )}
```
Sboonny, line 60: this is a workaround typescript error, there are more info in discord

https://discord.com/channels/692816967895220344/1024341059989409844/1024343219682689146

ojeytonwilliams:
Hey, @Sboonny, I think I see why this needs workarounds.  Part of the problem stems from the fact `event_sponsors` has `events` as a field, not `event`, making it look like it should be an array.  I think everything should be a bit clearer once that's fixed, so I'll push a commit with the change.

Sboonny:
@ojeytonwilliams, go ahead 👍, I have tried to make Events an array in here https://github.com/freeCodeCamp/chapter/pull/1581/commits/147fc8f4892436a822a6227fae61230b712df986

But that caused events to be unique, so I like your approach more 💚

ojeytonwilliams:
@Sboonny I didn't do much in the end, you and @gikf had nailed the implementation.  Hopefully it's a little more obvious why the flattening was necessary, though.

ojeytonwilliams (approved):
LGTM :+1:

PR #1584, opened 2022-09-22
chore: consistent color scheme

Sboonny:

 
 Although this makes it uglier, it's consistent and not confusing.
 
 - cancel and close are colorless, everywhere
 - everything that can be changed easily are `blue` and without risky affects.
 
 I wish it wasn't blue 😞

### DIFF
--- a/client/src/components/Modal.tsx
+++ b/client/src/components/Modal.tsx
@@ -63,12 +63,10 @@ export const Modal = (props: ModalProps) => {
           <ModalFooter>
             {buttonsLeft}
             <HStack>
-              <Button colorScheme="blue" onClick={modalProps.onClose}>
-                Close
-              </Button>
+              <Button onClick={modalProps.onClose}>Close</Button>
               {buttons}
               {formButtonText && (
-                <Button colorScheme={'green'} type="submit">
+                <Button colorScheme={'blue'} type="submit">
                   {formButtonText}
                 </Button>
               )}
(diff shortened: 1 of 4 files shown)


ojeytonwilliams:
While grey seems good for canceling an action (i.e. on confirm modals) shouldn't initiation actions have a stronger colour?  To my mind RSVP and Cancel should be coloured as should Subscribe and Unsubscribe.  They're all buttons to click if you want to do something, rather than buttons you're clicking because you've changed your mind.

Sboonny:
@ojeytonwilliams, I can see what you mean, but I may channel ahmad spirit by saying this.

The standard is that unsubscribe and similar buttons work as a button when you change your mind. For example, in YouTube UX the subscribe button is the brand color.  
image

And when the YouTuber change his mind, he will be presented and somewhat colorless button.
image

This will indicate to the user, unless you're canceling the subscription, the area is no longer of use to you.

---

Now that I used ahmad spirit, I will tell you the real reason of the change, I can't deal with hover and coloring without spending long time. I want to make it the same color as blue but with `opacity` or something like `.2` but adding more work isn't worth it.

---

All that being said, I think it's ugly and I am okay with reverting it back, if you want. moving this to post post mvp.

ojeytonwilliams:
I believe YouTube does that because they're trying to nudge users into staying subscribed, rather than to convey what the button does.

That said this is a complicated discussion (one which should involve Ahmad!), so we can revisit this down the line. 

I think the consistency is a plus, so let's get this in.

ojeytonwilliams (approved):

PR #1585, opened 2022-09-22
feat: render 404 when not found

gikf:

---

- Adds rendering 404 page when in database is not found... what should be found.

### DIFF
--- a/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
@@ -21,9 +21,6 @@ export const ChaptersPage: NextPage = () => {
   if (isLoading || error)
     return <DashboardLoading loading={isLoading} error={error} />;
 
-  // TODO: render something nicer if this happens. A 404 page?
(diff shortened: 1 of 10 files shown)


client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -21,9 +21,6 @@ export const ChaptersPage: NextPage = () => {
   if (isLoading || error)
     return <DashboardLoading loading={isLoading} error={error} />;
 
-  // TODO: render something nicer if this happens. A 404 page?
-  if (!data.chapters) return <div> No Chapters found</div>;
-
```
gikf, line 26: This is not needed. When there's no chapters in the instance, `data.chapter` exists, but it's an empty array. In such case, the `DataTable` will display _Empty_ text. For `data.chapter` not exist I assume there would need to be some kind of error, which would be caught either as `error` or event earlier.

ojeytonwilliams (approved):
Less code than the TODO comment :cold_sweat: Nice!

Sboonny (approved):

PR #1593, opened 2022-09-23
feat: auto subscribe switch

gikf:

---
- Adds option in profile to change the _Automatic chapter subscription_.
- When `true`:
  - Subscribes to chapter when joining it.
  - ~Subscribes to event when RSVPing. (Currently regardless of the individual chapter subscription).~
- When `false`:
  - Doesn't subscribe to chapter when joining it.
  - ~Subscribing to event depends on chapter subscription. If chapter subscription was manually changed to `true`, user will be subscribed to event when RSVPing.~
- Subscription to chapters already joined and RSVPed events is not affected.


client/src/modules/profiles/pages/userProfile.tsx
```diff
@@ -59,6 +85,24 @@ export const UserProfilePage = () => {
...
+              <FormLabel htmlFor="subscription">
+                Automatic subscription
+              </FormLabel>
+              <Switch
+                id="subscription"
+                isChecked={autoSubscribe}
+                onChange={onSwitch}
+                isDisabled={isLoading}
+              />
+              {isLoading && <Spinner size="sm" />}
```
Sboonny, line 99: I think the spinner is redundant with `isDisabled={isLoading}` because the switch color is changed subtly when it's disabled, and the cursor changes as well

```suggestion
```

client/src/modules/profiles/pages/userProfile.tsx
```diff
@@ -1,16 +1,27 @@
...
-import { Flex, Heading, Text, Link } from '@chakra-ui/layout';
+import React, { useEffect, useState } from 'react';
+import {
+  Flex,
+  FormControl,
+  FormHelperText,
+  FormLabel,
+  Heading,
+  Link,
+  Spinner,
```
Sboonny, line 9:
```suggestion
```

Sboonny (approved):
When reviewing, I noticed that profiles isn't sectioned, so it's hard to distinguish what something is for, and it feels chaotic.

But this is for a different pull request, I didn't find issues, only small nitpick which I am okay with this going through with dismissing as well.

Edit: I meant this as a comment, not as approval sorry for the confusion 🤦 😅

client/src/modules/profiles/pages/userProfile.tsx
gikf: I found the difference between disabled and unchecked to not be that distinct. It can be also disabled just for fractions of second, at which point if somebody isn't specifically looking for it, it might not be clear that switch was disabled. Appearing spinner is more likely to catch eye.

I'd consider disabling it only as a way to prevent from multiple clicks before setting is changed, and the spinner (or something else) as an indicator that something is happening/happened.

What do you think?

server/src/controllers/Events/resolver.ts
```diff
@@ -339,7 +339,7 @@ export class EventResolver {
     const isSubscribedToChapter =
       ctx.user.user_chapters.find(
         (user_chapter) => user_chapter.chapter_id === chapterId,
-      )?.subscribed ?? true;
+      )?.subscribed || ctx.user.auto_subscribe;
```
ojeytonwilliams, line 342:
```suggestion
      )?.subscribed ?? ctx.user.auto_subscribe;
```
Otherwise if `subscribed` is `false` then this will fall back to `auto_subscribe`.

server/src/controllers/Events/resolver.ts
```diff
@@ -576,7 +576,8 @@ ${unsubscribeOptions}`,
         sponsor_id,
       }));
 
-    const isSubscribedToEvent = userChapter ? userChapter.subscribed : true; // TODO add default event subscription setting override
+    const isSubscribedToEvent =
+      userChapter?.subscribed || ctx.user.auto_subscribe;
```
ojeytonwilliams, line 580:
```suggestion
      userChapter?.subscribed ?? ctx.user.auto_subscribe;
```
Same principle.

ojeytonwilliams (changes requested):
More of a question than a suggestion, but what do you think to including `auto_subscribe` updates in `updateMe`?  It cuts down on the number of resolvers, hooks and so on.

client/src/modules/profiles/pages/userProfile.tsx
Sboonny: Some people will find getting email, annoying, and it will be more annoying to think you are disabled it, but it actually didn't because the request wasn't sent to the server, and you had no idea of known that. 🤔 👍 

I am convinced

ojeytonwilliams:
@Sboonny yeah, that's a very good point.  People need to opt-in to receiving email, so we'll need to have this default to false and do something similar to the main repo.  That is to say, when people sign up for the first time, we should give them the option to receive emails from Chapter.

> Edit: maybe that what RSVP means, if so you can ignore me.

Nope, you will get notified (by email) about new events so that you can decide to go (which you do by RSVPing).

server/src/controllers/Events/resolver.ts
gikf: Generally that was the idea, when `auto_subscribe === true`, it subscribes when joining chapters and RSVPing to event. Regardless of the `chapterUser.subscribed` of the event's chapter.

But it can be changed to give `chapterUser.subscribed` precedence. In such case actually `auto_subscribe` won't be needed here at all, as `chapterUser` is always created before `eventUser`.

gikf:
I'm going to pull it into `updateMe` and flip the default to `false`.

server/src/controllers/Events/resolver.ts
ojeytonwilliams: Ah, right. So, this is a tricky one because it's important to avoid sending emails when a user thinks they've unsubscribed.  But it's also important that we have sensible defaults.

As a user, when I unsubscribe from a Chapter, I expect that Chapter to stop communicating with me.  However, if I then RSVP to an Event, do I want notifications about that specific Event, even though I don't want them from the Chapter generally?  Maybe, but I think we need to let people opt-in.

Another issue with the existing logic is how we communicate what it does.  We'd have to say something like 
> Automatically enable email notifications when you join a chapter or RSVP to an event (even if you are not subscribed to the chapter hosting that event).

vs.

> Automatically enable email notifications from a chapter when you join. 

Since we really want to encourage users to turn on notifications, I think it's important that the setting that does is transparent.

allella: I feel like the _RSVP_ notifications and _chapter_ notifications shouldn't impact each other.

If someone turns off _chapter.subscribed_, then they are turning off new event announcements, general announcements, and anything else an organizer might send to everyone.  I don't think that would disable RSVP related email notifications just like it wouldn't un-RSVP a person from an event. 

@ojeytonwilliams 's suggestion to labels clear is probably the only way we can avoid confusion.

Just a first pass at it

Chapter Page
_Would you like to receive email updates for new event and general announcements for this Chapter?_ 

RSVP Page / Form
_Would you like to stop all email updates for this event, including changes and cancellations?_

allella: Actually, I updated the suggestions so both are asking about disabling emails, like "turn off" and "stop".  We could very well make them both affirmative, like "enable" and "turn on", but it's probably good to do things in a similar format.

I guess it depends on the form element being used and even the current state of the element.  Like, could we change the language of the label based on the state? For example, if it's on, then ask "Would you like to turn off". If it's off, then ask "Would you like to turn on?".

Since we're talking about "Switches" I wanted to repost the earlier conversation about form element selection and about toggling notifications on and off as it was related to emails.  It looks like we even explored the use of the word "Follow {Chapter Name}" instead of "Subscribe" vs "Unsubscribe". I'm not suggesting we change the schema, just that we might use words for the user that help add to the context.

Facebook uses "Follow" and Meetup uses "Join" and the current UI has "Subscribe" vs "Unsubscribe".

One somewhat confusing thing is when you toggle "Subscribe" on it says "member of the chapter". So, does that mean if you unsubscribe that you're no longer a member? I'd imagine you can be a member of the group, and just not get notifications, but that may open a can of worms for the MVP.  

https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-562374895
https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-562423675
https://mui.com/material-ui/react-switch/#switches-with-formcontrollabel
https://github.com/freeCodeCamp/chapter/issues/107#issuecomment-931996286

allella:
Not that I'm suggesting this is MVP, but here's how other sites deal with the complexity of "belonging to a group" vs "getting notifications from the group" (via a series of switches)

## Meetup Example
image
and clicking "Group Notifications" goes to
image


## Facebook Example
image

image

server/src/controllers/Events/resolver.ts
allella: Also, given the types of custom switch designs where you have "On" and "Off", perhaps using a question format for a label, like I suggested above, isn't good with switches.

I know we're using [Chakra](https://chakra-ui.com/docs/components/switch) not using [Material UI, but their custom switches section](https://mui.com/material-ui/react-switch/#customization) shows some of the less-basic things like having a checkmark or the "On" and "Off" labels.

ojeytonwilliams: @allella that's a good point about the separation between Chapter and Event notifications.  If we make it clear what turning off Chapter notifications does, then it's probably fine to have clicking on RSVP to tacitly accept that you're going to get emails about that Event.

We will need to make this configurable sooner rather than later, though not necessarily for the MVP.

Also, though I've not tested this properly, certain notifications are sure to overlap with the notifications coming from Google Calendar.  We'll have to experiment a bit to make sure users can get the emails they want.

> I know we're using Chakra, but the Material UI docs show what you can do with custom switches like having a checkmark in the switch, or the "On" and "Off" labels.

We should be careful we don't customize switches too much, but something like what you shared below looks good.  I don't know how straightforward that would be to implement, though.

Finally, I'm not sure which word (follow or subscribe) works best.  That might be something we discover through user testing.

ojeytonwilliams: So, for this PR, I think the consensus is the following. 

- Chapter subscription should control if a user gets new event notifications
- Events should be subscribed to when a user clicks RSVP, regardless of Chapter subscription

We can handle further notification configuration separately (maybe post MVP).

@gikf @allella does that sound about right?

allella: For MVP, I'd agree new event notifications are the minimum. If / when organizers & admin can send an email to all Chapter members for a general announcements, then that would seem to fit under "subscribed to chapter notifications", at least until we break the controls down to smaller bits.

Part of our challenge with the switches for MVP is there may be more than one thing it turns on or off, as with the new events & general announcements example for a _chapter_. 

Eventually, we can have more controls to allow a user / member to be more granular about their notification preferences, as in the screenshots of Facebook and Meetup examples.

For the RSVP, it looks like we don't give the _member_ the option to turn off RSVP-related notifications until after they have hit the RSPV confirmation button.
image
That's fine, but eventually it may make sense to have the RSVP form with other options (media waiver, food preferences, and such, and that interface might also make sense to toggle on/off the RSVP notifications.

gikf: To be honest I'm still not sure what to do with how auto subscribe works for events.

What I'm sure is that currently, when unsubscribing from chapter, user is also unsubscribed from events in that chapter. Because that should be slightly outside of this PR scope, I've created issue - #1639

---

> * Chapter subscription should control if a user gets new event notifications

That's how it is, I believe. User has to be subscribed to chapter to receive emails about newly created events.

> * Events should be subscribed to when a user clicks RSVP, regardless of Chapter subscription

As I understand this, it means always subscribing to events when RSVPing.

ojeytonwilliams (changes requested):
Just some nitpicks:

ojeytonwilliams:
I blocked this because it requires a db migration and I've not set up the workflows for that yet.

client/src/modules/profiles/component/ProfileForm.tsx
```diff
@@ -83,6 +92,23 @@ export const ProfileForm: React.FC<ProfileFormProps> = (props) => {
...
+          <Switch
+            id="auto_subscribe"
+            {...register('auto_subscribe')}
+            isDisabled={loading}
+          />
+        </Flex>
+        <FormHelperText>
+          {hasAutoSubscribe
+            ? 'Automatically subscribe to chapter when joining it and receive emails about new events in chapter.'
+            : "Don't automatically subscribe to joined chapters."}
```
ojeytonwilliams, line 109:
```suggestion
            ? '(After joining a new chapter, you will be notified when new events are created.)'
            : '(After joining a new chapter, you will not be notified about new events unless you subscribe.)'}
```
This way the information is describing the current state, rather than what will be the case once you click the button.  I believe this is the most common pattern

ojeytonwilliams (changes requested):

client/src/modules/profiles/component/ProfileForm.tsx
```diff
@@ -83,6 +92,23 @@ export const ProfileForm: React.FC<ProfileFormProps> = (props) => {
           />
         ),
       )}
+      <FormControl>
+        <Flex>
+          <FormLabel htmlFor="auto_subscribe">
+            Automatic chapter subscription
```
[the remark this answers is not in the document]
gikf: I have two ideas going in a slightly different direction:
- Default chapter subscription
- Subscribe to chapters by default

What do you think? The helper text suggestion would work with these too.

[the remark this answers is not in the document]
ojeytonwilliams: You know what, I think "Subscribe to chapters when joining them" might be the way to go.  It's a bit verbose, but it at least gets across the point that 'joining' and 'subscribing' are two different things.

I've spent far too long thinking about this, though!

ojeytonwilliams (approved):

gikf:
@ojeytonwilliams Dunno if this was expected on the first run, but I had to use `migrate reset` initially, otherwise the `migrate dev` was finding dozens of changes.

PR #1594, opened 2022-09-23
feat: login when rsvping

ojeytonwilliams:
- fix: logout by refetching, not setting data
- refactor: make auth/store responsible for login

- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `main` branch of Chapter.

Closes #XXXXX

PR #1601, opened 2022-09-24
feat: login when rsvping (take two)

ojeytonwilliams:
- Revert "Revert "feat: login when rsvping (#1594)" (#1597)"
- fix: try to create session if no user data

The last attempt introduced a bug because the AuthStore only tried to create a session when no data at all came back from the `meQuery`. Instead it should try to create a session whenever it sees `null` user data.  Since it also needs `isAuthenticated` to be true, it will not attempt to login again after the user logs out of Auth0.


ojeytonwilliams:
I've already tested this a bit on staging, so I'm going to get this in.

PR #1604, opened 2022-09-25
feat: add required inputs

Sboonny:

Based on gikf comment here: https://github.com/freeCodeCamp/chapter/pull/1562#issuecomment-1256922451

I am wondering if it's possible to convince you to use https://www.pexels.com/photo/closeup-photo-of-brown-and-black-dog-face-406014/ as chapter default image?

If not, I will add https://www.pexels.com/photo/brown-window-curtain-532557/ and https://www.pexels.com/photo/orange-graphics-design-1154738/ as default values in primsa schema.

if you have a specific pictures in mind or hopefully library, we can use them instead


gikf:
If it would be up to me then pictures of dogs could be used as defaults, and it wouldn't be possible to change them :stuck_out_tongue:

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -299,6 +299,7 @@ const EventForm: React.FC<EventFormProps> = (props) => {
               label="Streaming URL"
               placeholder=""
               isDisabled={loading}
+              isRequired
```
ojeytonwilliams, line 302: An event is quite a generic thing, so it may be streamed or it may not.  It's quite possible someone could just plan to meet on Google meet, so streaming url would not be necessary.
```suggestion
```

ojeytonwilliams (changes requested):
Regarding the images, the main things are that they're inoffensive and we have the right to use them.  With Pexels, I'm not sure if we can link to them.  That said, if we can't, we should be able host whatever we want on our CDN.

ojeytonwilliams:
Sorry, @Sboonny, I wasn't thinking straight.  The form is fine, the issue is that the server is rejecting requests where the url string is not a valid URL. It shouldn't do that, because an event may not have a URL.

The fix is to modify `EditEventPage`'s submit function to send `null` if `data.url.trim()` is falsy, server/src/controllers/Events/inputs.ts needs the `@IsOptional` decorator for url and finally the `updateEvent` resolver should not fall back to `event.url`, it should allow `url` to be nullish if `data.url` is.

client/src/modules/dashboard/Events/pages/EditEventPage.tsx
```diff
@@ -58,6 +58,7 @@ export const EditEventPage: NextPage = () => {
         capacity: parseInt(String(data.capacity)),
         start_at: data.start_at,
         ends_at: data.ends_at,
+        url: data.url?.trim() ?? null,
```
ojeytonwilliams, line 61:
```suggestion
        url: data.url?.trim() || null,
```
If the `url` is just `''` then `??` will return it and then the validation will fail.

ojeytonwilliams (changes requested):

ojeytonwilliams (approved):
LGTM 👍

PR #1611, opened 2022-09-26
fix: wait for specific number of emails

gikf:

---

- Uses waiting for specific number of emails.
- It's kind of implied that after waiting for x emails, emails should have length of x. So _should_ might not be necessarily needed anymore.


ojeytonwilliams (commented):
Yeah, this is fine, but the should is redundant now.

ojeytonwilliams (approved):
LGTM :+1:

Issue #1608, opened 2022-09-26
I know _Sponsors_ dashboard is hidden from navigating to it without permission, but I've cheated and entered there directly. To see if individual pages (`/dashboard/sponsors/1`, `/dashboard/sponsors/1/edit`) will display access denied like other dashboards. It would be nice to make it similar to others.


ojeytonwilliams:
I know _Sponsors_ dashboard is hidden from navigating to it without permission, but I've cheated and entered there directly. To see if individual pages (`/dashboard/sponsors/1`, `/dashboard/sponsors/1/edit`) will display access denied like other dashboards. It would be nice to make it similar to others.

_Originally posted by @gikf in https://github.com/freeCodeCamp/chapter/pull/1572#pullrequestreview-1119322825_

ojeytonwilliams:
This isn't completely trivial because Sponsors occupy a weird place in our data model.  They're not attached to Chapters so it doesn't make sense to allow Chapter administrators to modify them, but they need to be able to add them to events.

We can either handle them like Venues (i.e. attach them to Chapters) or, if we need them to be instance wide, we could grant Chapter administrators a new SponsorView InstancePermission.

ojeytonwilliams:
Granting administrators permission is the simplest approach, so should be fine for the MVP.

However, post-MVP an instance-wide pool of Sponsors is probably the way to go, assuming we want instance owners to be the only ones capable of managing Sponsors.  They would then need to be able to attach sponsors to Chapters and at that point administrators would be able to view those sponsors and attach them to events.

PR #1620, opened 2022-09-28
feat: require permission for (most of) sponsor dashboard

ojeytonwilliams:
- fix: make add sponsor button more explicit
- feat(test): try to add sponsor when creating event
- test: add more pages needing permission
- refactor: rename permissions
- feat: add chapter_administrator instance role
- feat: require permission for sponsors dash
- feat: protect dashboard/sponsor/[id]

Ref: https://github.com/freeCodeCamp/chapter/issues/1608

It's "most of" because the query resolver to get a sponsor is used both to provide data for /dashboard/sponsor/[id] and /dashboard/sponsor/[id]/edit, but an admin is allowed to view the first, but not the second.

The only fix I know of is to create yet another resolver, but I've left it for now.


server/prisma/generator/factories/instanceRoles.factory.ts
```diff
@@ -36,18 +49,20 @@ const createRole = async ({ name, permissions }: InstanceRole) => {
       instance_role_permissions: { create: permissionsData },
     },
   });
-  return { [name]: { name: createdRole.name, id: createdRole.id } };
+  const role: RoleMap = {};
+  role[name] = createdRole;
+  return role as RoleMap;
```
gikf, line 54: I'm curious about this part. Is there some specific reason why `role` is both declared as `RoleMap` and then returned `as RoleMap`?

ojeytonwilliams: Uh, not a great reason. I think I was just trying a few things to get the types to play nicely and forgot about this. I'll try dropping it since it seems redundant.

gikf (approved):
Seems to be working as intended :+1:

PR #1622, opened 2022-09-28
fix: stop displaying emails

ojeytonwilliams:
- fix: never show emails in the app
- test: update to use names not emails

To protect user privacy this stops the api from returning email addresses.  They're still used by the app, but there's no need for any user (up to and including the owner) from seeing them.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -5579,22 +5579,6 @@
         "description": null,
         "fields": [
           {
-            "name": "email",
-            "description": null,
-            "args": [],
-            "type": {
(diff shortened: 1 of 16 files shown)


Sboonny (approved):

gikf:
Reminders are printing emails to console, does this count? https://github.com/freeCodeCamp/chapter/blob/main/server/reminders/reminders.ts

ojeytonwilliams:
Nice catch, @gikf. It's probably fine, but let's err on the side of caution.  After all, I'm not 100% convinced we'll need to log that info.

PR #1624, opened 2022-09-28
chore: add colors to call to actions buttons in menu

Sboonny:

Add coloring to `login` and `auth with google`

image

image

hover state
image


Sboonny:
I have reverted my DRY attempt, because the two components are different in structure. 

`Login is a button`, while `Auth with google` is a link. I should create a `className` that carry the style for both of them, but currently it isn't possible on my end.

ojeytonwilliams (approved):
LGTM :+1:

PR #1628, opened 2022-09-28
feat: add links from events/chapters pages to their dashboards

Sboonny:

while testing today, people may try to go to the dashboard through chapters, so I have added a button to navigate them through it, if they do want to navigate to the dashboard through `/chapters`

image


client/src/modules/chapters/pages/chaptersPage.tsx
```diff
@@ -14,7 +15,12 @@ export const ChaptersPage: NextPage = () => {
 
   return (
     <Stack mt={10} mb={5} display={'block'}>
-      <Heading marginBlock={'1em'}>Chapters: </Heading>
+      <Flex alignContent={'center'}>
+        <Heading marginBlock={'1em'}>Chapters: </Heading>
+        <LinkButton href="/dashboard/chapters" colorScheme={'blue'}>
+          Chapter Dashboard
+        </LinkButton>
```
ojeytonwilliams, line 22: It's a little misaligned
image

Also, can we hide this for logged out users?

ojeytonwilliams (changes requested):

client/src/modules/chapters/pages/chaptersPage.tsx
Sboonny: Do we need to hide them? we can disable them if they aren't logged in

at least this way people won't think, we are hiding stuff 🤔. but this maybe me overthinking it

Sboonny: I have added user import so hiding them is easy, what do we want to hide or disable the buttons?

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -71,11 +76,25 @@ export const EventsPage: NextPage = () => {
 
   const isLoading = loading || !data;
   if (isLoading || error) return <Loading loading={isLoading} error={error} />;
+  const canAuthenticateWithGoogle = useCheckPermission(
+    Permission.GoogleAuthenticate,
+  );
```
ojeytonwilliams, line 81: This can't be after a conditional, since every hook has to be called on every render: [rules of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

client/src/modules/policy/pages/policyPage.tsx
```diff
@@ -1,97 +1,136 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import { NextPage } from 'next';
-import { Flex, Heading, Text } from '@chakra-ui/react';
+import { Flex, Grid, Heading, Text, GridItem } from '@chakra-ui/react';
+import { LinkButton } from 'chakra-next-link';
+import { useAuth } from '../../../modules/auth/store';
```
ojeytonwilliams, line 5:
```suggestion
```

client/src/modules/policy/pages/policyPage.tsx
```diff
@@ -1,97 +1,136 @@
...
+import { useCheckPermission } from '../../../hooks/useCheckPermission';
+import { Permission } from '../../../../../common/permissions';
+
+interface Props {
+  link: string;
+  text: string;
+}
 
 export const PolicyPage: NextPage = () => {
+  const { user } = useAuth();
```
ojeytonwilliams, line 15:
```suggestion
```

client/src/modules/policy/pages/policyPage.tsx
```diff
@@ -1,97 +1,136 @@
...
+  );
+
+  const HeaderItem = forwardRef<HTMLDivElement, Props>(({ link, text }) => {
+    return (
+      <LinkButton
+        width={'18em'}
+        href={link}
+        colorScheme={'blue'}
+        margin={'1em'}
+        isDisabled={user?.id === undefined}
```
ojeytonwilliams, line 28:
```suggestion
```

client/src/modules/chapters/pages/chaptersPage.tsx
```diff
@@ -1,20 +1,38 @@
...
 export const ChaptersPage: NextPage = () => {
   const { loading, error, data } = useChaptersQuery();
-
+  const { user } = useAuth();
   const isLoading = loading || !data;
   if (isLoading || error) return <Loading loading={isLoading} error={error} />;
 
+  const canAuthenticateWithGoogle = useCheckPermission(
+    Permission.GoogleAuthenticate,
+  );
```
ojeytonwilliams, line 21: Similarly to eventsPage, this has to be called before the conditional.  Otherwise, if it returns early (with the `Loading` component) this hook won't be reached.

ojeytonwilliams (changes requested):

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -75,7 +81,14 @@ export const EventsPage: NextPage = () => {
   return (
     <VStack>
       <Stack w={['90%', '90%', '60%']} maxW="600px" spacing={6} mt={10} mb={5}>
-        <Heading>Events: </Heading>
+        <Flex justifyContent={'space-between'} alignItems={'center'}>
+          <Heading>Events: </Heading>
+          {canAuthenticateWithGoogle && (
```
gikf, line 86: Using the `GoogleAuthenticate` permission will show the buttons only to instance owners, what is probably not the intent. `useAuth()` can return now `isLoggedIn` property which would be more appropriate in these places. Or some new instance permission for accessing dashboard.

This applies to other pages too.

After updating branch with the changes from `main` couple things will need to be changed, as `useCheckPermission` is no longer there.

client/src/modules/policy/pages/policyPage.tsx
```diff
@@ -1,97 +1,132 @@
...
-    </Flex>
+        <Text>
+          That&apos;s all, folks. Know your privacy rights, and stay safe out
+          there!
+        </Text>
+      </Flex>
+      {canAuthenticateWithGoogle && (
+        <GridItem margin={'1em'}>
+          <HeaderItem link="/profile" text="Edit Your Profile" />
+          <HeaderItem link="/chapters" text="View More Chapters" />
```
gikf, line 126:
```suggestion
          <HeaderItem link="/chapters" text="View All Chapters" />
```

gikf:
Last thing that's bugging me is the position of button on chapters page:
image

I think it will look better if either, button is moved to the left, to line up with the end of chapter card. Or cards with chapters would be centered.

Sboonny:
I will center the card in different pr because, I ran out of my Gitpod plan 👍

client/src/modules/policy/pages/policyPage.tsx
```diff
@@ -1,97 +1,129 @@
...
+
+interface Props {
+  link: string;
+  text: string;
+}
 
 export const PolicyPage: NextPage = () => {
+  const { isLoggedIn } = useAuth();
+
+  const HeaderItem = forwardRef<HTMLDivElement, Props>(({ link, text }) => {
```
ojeytonwilliams, line 15:
```suggestion
  const HeaderItem = ({ link, text }: Props) => {
```

ojeytonwilliams (changes requested):
Looking good. Just minor nitpicks:

ojeytonwilliams (approved):

PR #1630, opened 2022-09-29
feat: add total of interested people in the event

Sboonny:

while discussing with ahmad, he mentioned the max number of RSVPs, because people are social and want to attend the meeting other people are attending, as well

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -41,8 +41,8 @@
   "devDependencies": {
     "@babel/core": "7.20.12",
     "@babel/plugin-proposal-decorators": "7.20.13",
-    "@codesee/babel-plugin-instrument": "0.493.0",
-    "@codesee/tracker": "0.493.0",
+    "@codesee/babel-plugin-instrument": "0.494.0",
+    "@codesee/tracker": "0.494.0",
     "@graphql-codegen/cli": "2.16.5",
     "@graphql-codegen/fragment-matcher": "3.3.3",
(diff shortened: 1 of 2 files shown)


client/src/components/EventCard.tsx
```diff
@@ -9,26 +9,17 @@ import {
...
-    | 'description'
-    | 'start_at'
-    | 'image_url'
-    | 'invite_only'
-    | 'canceled'
-  > & {
-    chapter: Pick<Chapter, 'id' | 'name'>;
-    tags?: EventTag[] | null;
-  };
+  event: EventWithRelations;
```
ojeytonwilliams, line 16: How about this. Use 
```suggestion
 event: ChapterQuery['chapter']['events'];
```
since that's what you're getting from the chapterPage and an optional `eventUsers` prop that you only pass in when you want to show this info?  Unless the plan is to show it everywhere, in which case it's fine to update the type to include `event_users`.

ojeytonwilliams (changes requested):

client/src/components/EventCard.tsx
Sboonny: [b05355a](https://github.com/freeCodeCamp/chapter/pull/1630/commits/b05355ab5858b9ad5db7c0c803859b94df5d3573) there is a change in types that make the solution not good enough, I will come back to this after fixing emails feedback, and I want to display users profiles as well so it was going to take a while anymore 👍

PR #1632, opened 2022-09-29
Refactor: remove events tags requirements from fields and resolvers

Sboonny:

There is no reason to use eventTags every time we try to use Event, so this aims to separate them.


server/prisma/generator/factories/events.factory.ts
```diff
@@ -83,52 +83,6 @@ const createEvents = async (
...
-    await prisma.events.update({
-      where: { id: event.id },
-      data: {
-        tags: {
-          create: connectOrCreateTags,
-        },
-      },
-    });
-
-    events.push(event.id);
```
Sboonny, line 131: I don't know why I have deleted factory, I guess I was on frenzy. anyway should they be brought back or when we try to use EventTags in the future we will use them?

server/src/controllers/Events/resolver.ts
```diff
@@ -719,22 +700,11 @@ ${unsubscribeOptions}`,
...
-              tag: {
-                connectOrCreate: {
-                  create: { name: tagName },
-                  where: { name: tagName },
-                },
-              },
-            })),
-          },
-        },
+        data: {},
```
Sboonny, line 707: Shouldn't these be deleted as well? I amn't sure, but now they don't update anything

server/src/graphql-types/Event.ts
```diff
@@ -44,10 +44,14 @@ export class Event extends BaseObject {
...
   @Field(() => String)
   image_url: string;
+}
 
+@ObjectType()
+export class EventWithTag extends Event {
   @Field(() => [EventTag])
   tags: EventTag[];
 }
+
```
ojeytonwilliams, line 54:
```suggestion
```
This doesn't get used, so we can remove it.

ojeytonwilliams (changes requested):
Looking good, just one minor thing:

ojeytonwilliams (approved):
LGTM :+1:

PR #1633, opened 2022-09-29
feat: allow users to leave chapters

### LINKED ISSUE #1631: Leaving and unsubscribing from chapters
A user needs the option to leave chapters entirely, not simply unsubscribe from notifications.

As @allella pointed out in https://github.com/freeCodeCamp/chapter/pull/1593#discussion_r983040671 it currently says has the text 'member of the chapter' next to the 'Unsubscribe' button:

image

This makes it appear that clicking 'Unsubscribe' will cause you to leave the group, but it does not - it just stops notifications.

TODO:
- [x] Create a public `leaveChapter` mutation
- [x] Add a 'Leave Chapter' button to chapters/[id] (client/src/modules/chapters/pages/chapterPage.tsx) to trigger that mutation
- [x] Improve the clarity of the UI so it's obvious that 'Unsubscribe' stops notifications while 'Leave Chapter' also removes you from every event you're currently attending.

Sboonny:

Ref #1631 

I have opened it, before doing anymore to the UI aspect, like making buttons sizes consistent, because the changes should apply to all buttons not only two, so I will dry them or add classNames with consistent styles

So for MVP, we can have leave function, and in post MVP I will try to make them looks better 👍


cypress/e2e/chapters/[chapterId].cy.ts
```diff
@@ -8,39 +7,24 @@ describe('chapter page', () => {
     cy.task('seedDb');
   });
 
+  //ToDo: check for if UI components are functioning as intended
```
Comment (gikf) [inline cypress/e2e/chapters/[chapterId].cy.ts:10]: That's what tests were doing before. Why remove them, instead of updating?

cypress/e2e/chapters/[chapterId].cy.ts
```diff
@@ -8,39 +7,24 @@ describe('chapter page', () => {
...
-      (chapter_users) => {
-        expect(
-          chapter_users.findIndex(
-            ({ user: { email }, subscribed }) =>
-              email === 'test@user.org' && subscribed,
-          ),
-        ).to.not.equal(-1);
+    cy.login();
+
+    cy.joinChapter(chapterId, { withAuth: true }).then((response) => {
```
Comment (gikf) [inline cypress/e2e/chapters/[chapterId].cy.ts:14]: This logins as owner - `foo@bar.com`, who is seeded as a member of chapter with id `1`.

cypress/support/commands.ts
```diff
@@ -545,6 +545,28 @@ const joinChapter = (chapterId: number, options = { withAuth: true }) => {
     : cy.request(requestOptions);
 };
 Cypress.Commands.add('joinChapter', joinChapter);
+/**
+ * Join chapter using GQL mutation
```
gikf, line 549:
```suggestion
 * Leave chapter using GQL mutation
```

gikf (commented):
General question - should something happen with events from the left chapter, if user RSVPed to them?

cypress/e2e/chapters/[chapterId].cy.ts
Comment (Sboonny) [reply cypress/e2e/chapters/[chapterId].cy.ts:10]: This UI is for MVP, we can have them back but I am going to style them in post MVP, and create switch as described in the subscribe issue in the profile, and I want the file to test both layer.

cypress/e2e/chapters/[chapterId].cy.ts
Comment (Sboonny) [reply cypress/e2e/chapters/[chapterId].cy.ts:14]: Owner can leave chapter 🤔🤦

cypress/e2e/chapters/[chapterId].cy.ts
Comment (gikf) [reply cypress/e2e/chapters/[chapterId].cy.ts:10]: Could you elaborate a bit? This sounds like tests were removed, because in future style will be changed

client/src/modules/chapters/pages/chapterPage.tsx
```diff
@@ -136,34 +151,56 @@ export const ChapterPage: NextPage = () => {
...
+                      colorScheme="blue"
+                      onClick={() => chapterSubscribe(true)}
+                      size="md"
+                    >
+                      Subscribe
+                    </Button>
+                  </HStack>
+                )}
+              </HStack>
+            )
```
gikf, line 181:
```suggestion
            <HStack>
              {dataChapterUser?.chapterUser?.subscribed ? (
                <HStack justifyContent={'space-between'} width={'100%'}>
                  <Text fontWeight={500}>
                    Unfollow upcoming chapter&apos;s events
                  </Text>
                  <Button onClick={() => chapterSubscribe(false)} size="md">
                    Unsubscribe
                  </Button>
                </HStack>
              ) : (
                <HStack justifyContent={'space-between'} width={'100%'}>
                  <Text fontWeight={500}>
                    Follow upcoming chapter&apos;s events
                  </Text>
                  <Button
                    colorScheme="blue"
                    onClick={() => chapterSubscribe(true)}
                    size="md"
                  >
                    Subscribe
                  </Button>
                </HStack>
              )}
            </HStack>
```

`dataChapterUser?.chapterUser?.subscribed && (` wraps the whole `HStack`, what makes impossible to reach the false of ternary operator - button to subscribe will not be displayed.

gikf:
```suggestion
            dataChapterUser?.chapterUser && (
              <HStack>
                {dataChapterUser.chapterUser.subscribed ? (
                  <HStack justifyContent={'space-between'} width={'100%'}>
                    <Text fontWeight={500}>
                      Unfollow upcoming chapter&apos;s events
                    </Text>
                    <Button onClick={() => chapterSubscribe(false)} size="md">
                      Unsubscribe
                    </Button>
                  </HStack>
                ) : (
                  <HStack justifyContent={'space-between'} width={'100%'}>
                    <Text fontWeight={500}>
                      Follow upcoming chapter&apos;s events
                    </Text>
                    <Button
                      colorScheme="blue"
                      onClick={() => chapterSubscribe(true)}
                      size="md"
                    >
                      Subscribe
                    </Button>
                  </HStack>
                )}
              </HStack>
            )
```

Sorry, just the top condition needs changing.

client/src/modules/chapters/pages/chapterPage.tsx
```diff
@@ -136,34 +151,56 @@ export const ChapterPage: NextPage = () => {
...
+                <CheckIcon marginRight={1} />
+                {dataChapterUser.chapterUser.chapter_role.name} of the chapter
+              </Text>
+              <Button onClick={leaveChapter}>Leave</Button>
             </HStack>
           ) : (
-            <Button colorScheme="blue" onClick={joinChapter}>
-              Join chapter
-            </Button>
+            <HStack>
```
gikf, line 196:
```suggestion
            <HStack justifyContent="space-between">
```
For consistency. After joining chapter, buttons are spaced similarly.

cypress/e2e/chapters/[chapterId].cy.ts
Comment (gikf) [reply cypress/e2e/chapters/[chapterId].cy.ts:14]: I mean, changed code checks if when user being already part of chapter, and tries to join it again, doesn't end up with error.

Comment (Sboonny) [reply cypress/e2e/chapters/[chapterId].cy.ts:14]: Well, time to update the function to

- stop joined member from joining
- stop owner from leaving
- update cypress tests to check for these

client/src/modules/chapters/pages/chapterPage.tsx
Sboonny: No worries and thank you 💚, there is a similar condition, I will update it as well.

cypress/e2e/chapters/[chapterId].cy.ts
Comment (Sboonny) [reply cypress/e2e/chapters/[chapterId].cy.ts:10]: I will update the comment with more details 👍

Sboonny:
About the general question, to be honest I find it weird that RSVPs aren't stored in chapterUser or there is a field for it that cascade on deletion.

to answer, I would want my RSVPs to be deleted, when I leave chapter 🤔. What do you think?

allella:
If we're going to delete RSVPs when leaving a chapter, then I think we should include that as a warning and/or trigger an email letting the person know they've cancelled their RSVP to events too.

There are situations where someone may sign up for an event and then get annoyed by the chapter's notifications and decide to leave the chapter but still plan to attend the event. For instance, there are some local chapters in town that do annual events that I might attend, but I typically won't go to other events they do throughout the year. If I signed up for a conference and then "muted" the chapter's other notifications by leaving, then I may not assume I'm also loosing my RSVPs for events.

This is an edge case, but I think cancelling RSVPs without a warning or notice is too big of an assumption for the "leave chapter" button.

Sboonny:
I am keeping it as is (not RSVPs from events), but we should have both options, maybe adding a checkbox asking if you want to remove RSVPs as well.

gikf:
> If I signed up for a conference and then "muted" the chapter's other notifications by leaving, then I may not assume I'm also loosing my RSVPs for events.

For that distinction between being part of chapter and being subscribed to it is vital. Because there are no event roles and event permissions, keeping RSVP without being part of chapter is very problematic. Chapter role has permissions for RSVPing (and changing it) and managing event subscription. Not being part of chapter means changing RSVP status or subscription to event will not be authorized.

cypress/e2e/chapters/[chapterId].cy.ts
Comment (gikf) [reply cypress/e2e/chapters/[chapterId].cy.ts]: - In general I think that's somewhat convenient, to kind of ignore on the server side further attempts to join chapter. I'm not sure if it's worth having it tested explicitly or changing.
- Why instance owner shouldn't be allowed to leave chapter?

allella:
@gikf I'm not familiar with the authorization side of things, but from a schema standpoint it appears these are not tied together since both event_users and chapter_users are separate and both have _subscribed_ fields.

https://opensource.freecodecamp.org/chapter/tables/event_users.html
https://opensource.freecodecamp.org/chapter/tables/chapter_users.html

If it's going to complicate things, then we can at least warn people that leaving the chapter will remove their RSVPs too.

cypress/e2e/chapters/[chapterId].cy.ts
Comment (Sboonny) [reply cypress/e2e/chapters/[chapterId].cy.ts]: > In general I think that's somewhat convenient, to kind of ignore on the server side further attempts to join chapter. I'm not sure if it's worth having it tested explicitly or changing.

I have no idea, I like the calls to server to be tested by cypress because I don't have that much experience with dealing with server side issues, so I feel comfortable with safe nets around the server-side.

> Why instance owner shouldn't be allowed to leave chapter?

The word owner confused me, and I thought he was owner of chapter not instance anyway, you are right it doesn't matter 👍 

---

> I don't have that much experience with dealing with server side issues

This does count chapter, which is scary 😨. And I may lack experience in client side as well. 🤷

Sboonny:
have you started working on adding https://github.com/freeCodeCamp/chapter/pull/1633#issuecomment-1266440628? @gikf 

Can I pick it up or should I clear some work on progress PRs?

gikf:
> have you started working on adding [#1633 (comment)](https://github.com/freeCodeCamp/chapter/pull/1633#issuecomment-1266440628)? @gikf

I'm not sure to what you refer in that comment...

cypress/e2e/chapters/[chapterId].cy.ts
Comment (gikf) [reply cypress/e2e/chapters/[chapterId].cy.ts]: I see the changed comment, but that doesn't explain it for me. Most of the UI, if not all, can be considered as not final (or maybe it is?), that doesn't mean all tests should be deleted.

The new tests are checking only if request to server doesn't result in error. The lack of error is not very strong proof that functionality is working. Another thing is by using only direct server request tests don't check the user side of things, user don't communicate with server directly.

Sboonny:
- Adding events_roles and event_permession

---

- event_permession is responsible for receiving emails
- link subscription to event_permession
- when someone unsubscribe the lose event_permission, so they lose ability to receive emails 

---

- event_roles links to event_users
- event_roles links to chapter_role
- chapter_role links to chapter_users
- when someone leave they lose their entry in chapter_user
- this cascade and delete their entries in event_users 

---

Wow I thought this was simpler, 🤔🤯

ojeytonwilliams:
> Not being part of chapter means changing RSVP status or subscription to event will not be authorized.

Yeah, good point. It would be a fairly rubbish UX to leave a chapter, realise you were still attending events and be unable to cancel.

> There are situations where someone may sign up for an event and then get annoyed by the chapter's notifications and decide to leave the chapter but still plan to attend the event.

Absolutely. I think the main thing we need to do is be clear that you can mute notifications without having to leave the Chapter. Also that if you do leave the chapter, then you'll leave its events in the process.

We can revisit this whole thing when we have individual permissions for an event, making it possible to leave a chapter while still retaining permissions for that event. It's a bit of a task, though, so I'd rather wait until the MVP is out and somewhat stable before attempting it.

Also, as discussed, it might not be necessary. If we communicate how subscription works well enough, then people may be happy to stay members.

> Wow I thought this was simpler

I wish! It's a challenge trying to keep the flexibility we want without the complexity stopping progress.

server/prisma/schema.prisma
```diff
@@ -172,7 +172,7 @@ model event_users {
   rsvp_id        Int
   subscribed     Boolean
   title          String[]
-  user           users            @relation(fields: [user_id], references: [id], onDelete: Cascade, onUpdate: NoAction)
+  user           users            @relation(fields: [user_id], references: [id], onDelete: NoAction, onUpdate: NoAction)
```
Sboonny, line 175: there is an issue that when I delete event_user to delete RSVP, way more is deleted with it.

- owner lose their permission when they leave any chapter

```suggestion
  user           users            @relation(fields: [user_id], references: [id], onDelete: Cascade, onUpdate: NoAction)

Comment (Sboonny) [reply server/prisma/schema.prisma]: Nvm this, just bad logical statements from my part

File: server/src/controllers/ChapterUser/resolver.ts [hunk 7]
```diff
@@ -54,7 +54,6 @@ export class ChapterUserResolver {
     });
   }
 
-  @Authorized(Permission.ChapterJoin)
```
Sboonny, line 46: I have removed it because as long as you are logged in, there shouldn't be a blocked for you joining and leaving a chapter, as pointed out by Oliver.

That being said, I am wondering about RSVPs, because people need to be Auth for chapter to book events for them. Do we need to add a blocker for RSVP, checking if they auth us to update their Google calendar?

ojeytonwilliams: Thing is, all members have the `JoinChapter` permission.  With the exception of banned users (they're restricted and maybe we should kick them out of the chapter, too, I don't think that's been settled yet) there are no restrictions imposed by this decorator.

> That being said, I am wondering about RSVPs, because people need to be Auth for chapter to book events for them.

If someone leaves a chapter, they're leaving the events, too.  This means their RSVPs will cease to exist and the Google Calendar Event needs updating to reflect the fact they're not longer attending.

> Do we need to add a blocker for RSVP, checking if they auth chapter to update their Google calendar?

No, because it's not their calendar.  They're invited to an event on the Chapter's Calendar (technically the calendar of the owner that first authenticated with Google), so they don't have to authorize anything.  It's the same thing as if I invited you to an event on my calendar - you don't have to authorize that, you just get an invite in the mail.

server/src/controllers/ChapterUser/resolver.ts
```diff
@@ -83,7 +83,41 @@ export class ChapterUserResolver {
...
+      include: chapterIncludes,
+    });
+  }
+
+  @Mutation(() => ChapterUser)
+  async leaveChapter(
+    @Arg('chapterId', () => Int) chapterId: number,
+    @Ctx() ctx: Required<ResolverCtx>,
+  ): Promise<ChapterUser | null> {
+    await prisma.event_users.deleteMany({
```
ojeytonwilliams, line 95: As mentioned above, this resolver will need to cancel the Google Calendar invites for these events.

ojeytonwilliams (changes requested):

server/src/controllers/ChapterUser/resolver.ts
Sboonny: I have no idea how to update google calendar 😅

Sboonny: does deleting google calendar need to be added to removing RSVP from Event, as well?

ojeytonwilliams: The resolver has a helper `updateCalendarEventAttendees` which handles this.  Once the _event_users_ have been updated, this can be called to update the Google calendar with those changes.

Have a look at the logic for `deleteRsvp`.  It's quite similar to `leaveChapter` at least as far as calendar event attendees.

Sboonny: Copying deleteRsvp lead to this, which didn't create issue. but here is an issue with the logic because it delete one calendar instead of all of them

```ts
await prisma.event_users.delete({
      where: { user_id_event_id: { user_id: ctx.user.id, event_id: { chapter_id: chapterId } } },
      select: {
        event: {
          select: {
            calendar_event_id: true,
            chapter: {
              select: { calendar_id: true },
            },
          },
        },
      },
    });
```

I tried to include, calendar_id and a like this. but getting type error

```ts
     await prisma.event_users.deleteMany({
       where: {
         user_id: ctx.user.id,
         event: { chapter_id: chapterId },
         calendar_event_id: { chapter_id: chapterId },
         chapter: {
          select: { calendar_id: true },
        } 
       },       
     });
```

here is the type error.

>   Object literal may only specify known properties, and 'calendar_event_id' does not exist in type 'event_usersWhereInput'

Wondering how to include the `calendar_event_id` without adding field to `event_user`?

gikf: What is `calendar_event_id` expected to do here?

Sboonny: find events in the calendars which has a relation with chapterId

gikf: `calendar_event_id` doesn't have relation with chapter

cypress/e2e/chapters/[chapterId].cy.ts
Comment (Sboonny) [reply cypress/e2e/chapters/[chapterId].cy.ts]: Sorry for the lack or reply, didn't notice the comment.

I don't remember specific reason of why I didn't implement cypress tests, and not update them. But I can tell you how I currently feel about adding test to these component.

I don't think there is need to cypress test the component, because the time spend going around the test and making the test work isn't worth it. 

We are using variables like `chapterUser?.subscribed`, which indicates if the user is joined or not to. 

So unless someone changes the variables' name, create pull request that passes enlist and build errors, pr is merged it, create conflict with someone else, then the 2nd person fix the conflict but doesn't account for changed variables, and passes enlist and build errors, then the broken pr is merged with flawed logic, and cause the button to not work.

Therefor, the variables like `chapterUser?.subscribed` should be acting as our static tests.

The users should be using join and leave chapter regularly, if this doesn't work we will know quickly about it.

Server issues are harder to fix, than client side, if this breaks in the client, I can't think of a scenario in which we can't revert the breaking change, fix it and implements new change in small period of time. (this may be me lacking experience in these parts)

---


Saying all this, although I find test around specific sequence or UI function annoying, I see it's usage. I am totally okay with update them and added them back if deemed needed

Comment (Sboonny) [reply cypress/e2e/chapters/[chapterId].cy.ts]: nvm that oliver implement a test 🤣

Sboonny:
Apparently stopping the push will stop the lint, but not the push commit itself 🤷, I have added gikf to the wrong commit as co-author.

Anyway this should be it, unless there is something missed, meanwhile I am going to mess around with the code. 👍

ojeytonwilliams:
@Sboonny I'm just trying to see if the error Cypress is reporting is real, hence the merge. Hopefully not!

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -87,10 +87,14 @@ export const EventPage: NextPage = () => {
     const confirmOptions = options?.invited
       ? {
           title: 'You have been invited to this event',
-          body: 'Would you like to attend?',
+          body: `
+          Would you like to attend?<br/>
+          Note: if you haven't joined event's chapter, you will become member when RSVPing
```
ojeytonwilliams, line 92:
```suggestion
          Note: joining this event will make you a member of the event's chapter.
```
Let's go with this for now.

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -87,10 +87,14 @@ export const EventPage: NextPage = () => {
...
+          `,
         }
       : {
           title: 'Join this event?',
+          body: `Note: if you haven't joined event's chapter, you will become member when RSVPing`,
```
ojeytonwilliams, line 97:
```suggestion
          body: `Note: joining this event will make you a member of the event's chapter.`
```

ojeytonwilliams (changes requested):
This is looking great.  Just some minor nitpicks:

client/src/modules/dashboard/Events/graphql/mutations.ts
```diff
@@ -68,3 +68,17 @@ export const sendEventInvite = gql`
...
+      id
+      name
+      canceled
+      description
+      url
+      streaming_url
+      capacity
+    }
+  }
+`;
```
Sboonny, line 84:
```suggestion
```

cypress/e2e/chapters/[chapterId].cy.ts
```diff
@@ -1,61 +1,64 @@
...
+    cy.toggleChapterSubscription(chapterId, { withAuth: true }).then(
+      (response) => {
+        expect(response.status).to.eq(200);
+        expect(response.body.errors).not.to.exist;
       },
     );
+    cy.leaveChapter(chapterId, { withAuth: true }).then((response) => {
+      expect(response.status).to.eq(200);
+      expect(response.body.errors).not.to.exist;
+    });
```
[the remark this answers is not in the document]
Comment (Sboonny) [reply cypress/e2e/chapters/[chapterId].cy.ts:28]: sure, didn't know it was merged 👍

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -90,10 +90,14 @@ export const EventPage: NextPage = () => {
...
           title: 'You have been invited to this event',
-          body: 'Would you like to attend?',
+          body: `
+          Would you like to attend?<br/>
+          Note: joining this event will make you a member of the event's chapter.
+          `,
         }
       : {
           title: 'Join this event?',
+          body: `Note: joining this event will make you a member of the event's chapter.`
```
Sboonny, line 100:
```suggestion
          body: `Note: joining this event will make you a member of the event's chapter.`,

File: cypress/e2e/chapters/[chapterId].cy.ts [hunk 14]
```diff
@@ -47,21 +49,42 @@ describe('chapter page', () => {
         ).to.not.equal(-1);
       },
     );
+    cy.leaveChapter(chapterId, { withAuth: true }).then((response) => {
+      expectNoErrors(response);
+    });
```
Comment (ojeytonwilliams) [inline cypress/e2e/chapters/[chapterId].cy.ts]:
```suggestion
    cy.leaveChapter(chapterId, { withAuth: true }).then(expectNoErrors);
```

ojeytonwilliams (approved):
LGTM :+1: 

Nice work, @Sboonny :rocket:

PR #1634, opened 2022-09-29
fix: populate rsvp table during init

ojeytonwilliams:

The majority of changes were to the test code, to ensure that a fresh start is tested at least once.  Their organisation is a little messy now, but I'm not sure there's an obviously better approach.

The issue itself was simply that the rsvp table had no entries causing event creation to fail (it tries to connect and fails)

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -5,6 +5,7 @@ import { config } from 'dotenv';
 import coverage from '@cypress/code-coverage/task';
 
 import { prisma } from './server/src/prisma';
+import { InstanceRole } from './server/prisma/generator/factories/instanceRoles.factory';
 
 const getChapterMembers = (chapterId: number) =>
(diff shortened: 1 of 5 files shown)


ojeytonwilliams:
Sorry for the noise on this PR. I'm at a loss to why it's failing in CI, since it's fine locally.

ojeytonwilliams:
Alright.  The main issue seems to be that two instances of the prisma client are clashing with each other postgres is throwing 

> ERROR:  cached plan must not change result type

quite why, I'm not sure, but once this happens it causes issues with subsequent tests - `seedDb` starts to fail.

ojeytonwilliams:
This *might* just be a problem in testing (since we're not going to be reseting/seeding/syncing the db on the fly in production), but I'm not quite convinced, yet.

Sboonny (approved):

PR #1644, opened 2022-10-02
feat: event dashboard improvements

Sboonny:

TD;LR, admins shouldn't be able to control if someone else will attend the meeting, and it isn't needed to be controlled if someone will do so

While trying to make `kick` clearing for UX, I had the idea why do we need to allow admins, to confirm or kick other people from the Events while there is `ban`. 

So the admin shouldn't be able to book the events for a random person who may not want to attend, if someone is harassing the event managers then they would want to ban, not kick them.

About the average user, they can leave and confirm with the calendar normally, and they don't need to kick themselves because there is a leave button, and they can confirm through the email


Sboonny:
I haven't commented the cypress test, or deleted it because I am not sure 100%, that I am not missing something out and these button does something else that isn't related to RSVP

ojeytonwilliams (changes requested):
I agree that kicking a user that you're not planning to ban is going to be rare, it's still useful.  If someone has joined something inappropriate (e.g. a beginner's only course, when they're experienced) then kicking them seems reasonable, but banning does not.

Confirmation is required for events with a waiting list, so that administrators can decide who gets to go.

If there was a lot to fix, it might well be quicker to the MVP to scrap these features, but I don't think there is.

How about we change the text, so it's more descriptive?  Kick -> Remove from event and Confirm -> Add to attendees, for example.

Sboonny:
I like `Remove`, good call 👍 

We can change `Confirm` to `Add to Attendees` when we decided to change the `RSVPs` 👍, although I don't think we have problem with `Confirm`

ojeytonwilliams:
@Sboonny just to check: this removes the button to send emails and renames 'kick' to 'remove'?

Sboonny:
Yeab, I have altered the styles too, because having button in middle of the text was weird.

image

But as functionality goes, rename the remove event and delete the `Event Attendees`

gikf (approved):
LGTM :+1:

ojeytonwilliams (approved):
LGTM :+1: Thanks, @Sboonny!

PR #1645, opened 2022-10-02
feat: use nested layout for dashboard

gikf:

---
- Uses https://nextjs.org/docs/basic-features/layouts#with-typescript pattern to create nested layout for dashboard pages.
- Currently on every dashboard page the dashboard layout (_Chapters_, _Events_, (...) blue/grey buttons below header) are loaded freshly along the page.
- This change makes the dashboard layout to behave similarly to the page header - _dashboard layout state_ is preserved between dashboard pages and only necessary changes are rendered.
- I haven't actually looked if the difference between the two can be observed right now. It's part that exploded when I was making another change and with that the difference would be painfully obvious.


Sboonny (approved):
With the new change

image

----

the old ways: 

image

ojeytonwilliams (approved):
LGTM :+1: 

I couldn't see an obvious performance increase, but this does look to be the right approach overall.

PR #1646, opened 2022-10-02
feat: ability to update personal profile picture

Sboonny:

this aims to create, profile picture field which will be used by `Avatar` chakra component to display personal characteristics into profile picture. 

Will I attempt to block any type of images? nope, just adding the ability to upload an image, more than can be dealt with post MVP

the outcome 

image


client/src/components/PageLayout/Header.tsx
```diff
@@ -78,6 +78,7 @@ export const Header: React.FC = () => {
 
   const goHome = () => router.push('/');
 
+  console.log(user?.image_url);
```
gikf, line 81:
```suggestion
```

gikf (commented):
I don't know if that's intended or not. When using transparent image (ie. https://cdn-icons-png.flaticon.com/512/149/149157.png), the background color that's set by the user name will still be used.

image image

Sboonny:
It isn't intended, it's chakra doing good work.

That being said, do we want transparent image? I think it's harmful for UX, but we don't have users yet so we can have it 🤔 and test it

Sboonny:
Stop it as 100%? I don't think so. 

But we can create an `::after` content (maybe a nice trending glass effect) that stay behind the image, if the image is transparent, then that `::after` content will show up instead (well technically the image will disappear and what behind it (::after) will show up).

Should I give that a go and try to override chakra background color?

ojeytonwilliams:
I could be underthinking it, but how about setting the background to white if there's an image?

Sboonny:
There are pages that use `avatar` component in them, mainly ChapterPage and EventPage, although this isn't affecting them yet. it should be DRYed and used on these pages. if we're going to DRY it, using background white won't work in UX aspect but the idea is on the mark 👍

Sboonny:
I have reread this, and I was confused. I though we didn't want a transparent Image, but we do want transparent photo, I am adding them 👍

server/prisma/schema.prisma
```diff
@@ -264,6 +264,7 @@ model users {
   updated_at       DateTime        @updatedAt
   id               Int             @id @default(autoincrement())
   name             String
+  image_url        String
```
ojeytonwilliams, line 267:
```suggestion
  image_url        String?
```
Having this be nullable better reflects the fact that the image_url is optional. It also means that `image_url: ''` can be deleted everywhere.

ojeytonwilliams (changes requested):

server/prisma/generator/factories/user.factory.ts
```diff
@@ -19,27 +19,31 @@ const createUsers = async (
     email: 'foo@bar.com',
     name: 'The Owner',
     instance_role: { connect: { id: instanceRoles.owner.id } },
+    image_url: '',
```
ojeytonwilliams, line 22:
```suggestion
```

server/prisma/generator/factories/user.factory.ts
```diff
@@ -19,27 +19,31 @@ const createUsers = async (
...
   };
   const owner = await prisma.users.create({ data: ownerData });
 
   const chapter1AdminData: Prisma.usersCreateInput = {
     email: 'admin@of.chapter.one',
     name: 'Chapter One Admin',
     instance_role: { connect: { id: instanceRoles.chapter_administrator.id } },
+    image_url: '',
```
ojeytonwilliams, line 30:
```suggestion
```

server/prisma/generator/factories/user.factory.ts
```diff
@@ -19,27 +19,31 @@ const createUsers = async (
...
   };
   const chapter1Admin = await prisma.users.create({ data: chapter1AdminData });
 
   const chapter2AdminData: Prisma.usersCreateInput = {
     email: 'admin@of.chapter.two',
     name: 'Chapter Two Admin',
     instance_role: { connect: { id: instanceRoles.chapter_administrator.id } },
+    image_url: '',
```
ojeytonwilliams, line 38:
```suggestion
```

server/prisma/generator/factories/user.factory.ts
```diff
@@ -19,27 +19,31 @@ const createUsers = async (
...
   };
   const chapter2Admin = await prisma.users.create({ data: chapter2AdminData });
 
   const bannedAdminData: Prisma.usersCreateInput = {
     email: 'banned@chapter.admin',
     name: 'Banned Chapter Admin',
     instance_role: { connect: { id: instanceRoles.chapter_administrator.id } },
+    image_url: '',
```
ojeytonwilliams, line 46:
```suggestion
```

server/prisma/generator/factories/user.factory.ts
```diff
@@ -48,6 +52,7 @@ const createUsers = async (
     email: 'test@user.org',
     name: 'Test User',
     instance_role: { connect: { id: instanceRoles.member.id } },
+    image_url: '',
```
ojeytonwilliams, line 55:
```suggestion
```

server/prisma/generator/factories/user.factory.ts
```diff
@@ -58,6 +63,7 @@ const createUsers = async (
       email: internet.email(),
       name: `${name.firstName()} ${name.lastName()}`,
       instance_role: { connect: { id: instanceRoles.member.id } },
+      image_url: '',
```
ojeytonwilliams, line 66:
```suggestion
```

server/src/app.ts
```diff
@@ -120,6 +120,7 @@ export const main = async (app: Express) => {
             name: 'member',
           },
         },
+        image_url: '',
```
ojeytonwilliams, line 123:
```suggestion
```

server/tests/fixtures/users.ts
```diff
@@ -38,6 +38,7 @@ const baseUser: User = {
   created_at: new Date(),
   updated_at: new Date(),
   email: 'an@add.ress',
+  image_url: '',
```
ojeytonwilliams, line 41:
```suggestion
```

ojeytonwilliams (changes requested):
Pretty sure we don't need any of these.

server/prisma/generator/factories/user.factory.ts
Sboonny: Oh 😅, they were an attempt to please typescript, is there a better way around that type error?🤔

ojeytonwilliams:
@Sboonny I seem to have upset Cypress. Sorry about that!  I'll fix it shortly.

Sboonny:
no worries, feel free to 👍

ojeytonwilliams (approved):
LGTM :+1: Assuming the CI passes, :crossed_fingers:, this seems ready to me.

PR #1659, opened 2022-10-05
feat: ability to wait for client permission check

gikf:

---
- Adds ability to wait for the client permission check in the `useCheckPermission`. This allows avoiding the _incorrect_ initial render, when permission is not checked, and visible update to the _correct_ state, once permission is loaded.
- It might be good to also add such ability when using `useAuth` hook, to get some more flexibility.
- Using the change, delays rendering in couple of places. The most visible is the _Menu_/_Profile_ in the header.
- Updates the dashboard header to not display _Users_ link, if user doesn't have appropriate permission.
- Adds displaying `403` error when user visits directly `/dashboard/sponsors/new` and doesn't have permission to add new sponsor - slight experiment how that might look like.


ojeytonwilliams:
The functionality is looking good, but I think `useCheckPermission` has ended up taking on too many responsibilities.  Fundamentally, it's job should be to check that the current user has a specific permission.  Checking to see if the user is loading is really the responsibility of the `auth/store` module.

I poked around a bit with this locally, so if you're okay with this change of direction (and me pushing a commit) I'll clean up what I have and push it.

gikf:
@ojeytonwilliams Sure, feel free to do that.

---

I don't know why those two tests are having issues, I wasn't able to reproduce fails at those points locally.

ojeytonwilliams:
All done.  Let me know if these changes make sense.

> I don't know why those two tests are having issues, I wasn't able to reproduce fails at those points locally.

Me either.  Hopefully they've magically resolved themselves.

gikf:
> All done. Let me know if these changes make sense.

I like it, it's way simpler than what I contemplated when thinking how to put loading into `useAuth` :rocket: 

Magically resolving test is not a bad name :eyes:

ojeytonwilliams:
Thanks. I'm not sure why it took so long to realise that was the way to go, but it seems right. And, yeah, the magic happened, so let's get this in :)

ojeytonwilliams (approved):

PR #1669, opened 2022-10-06
feat: refactor menu layout and home page layout

Sboonny:

The menu is cluttered, so I am going to link `homePage` to `/chapters` and `/events` through buttons/links (haven't decided yet, and we can update it later), this will remove two `menuItems`

I will swap `profile` and `auth with google`, because people will use it once, so there is no reason to draw attention to it


client/src/modules/policy/pages/policyPage.tsx
```diff
@@ -1,15 +1,44 @@
 import React from 'react';
 import { NextPage } from 'next';
 import { Flex, Heading, Text } from '@chakra-ui/react';
+import { LinkButton } from 'chakra-next-link';
 
 export const PolicyPage: NextPage = () => {
+  const serverUrl =
+    process.env.NEXT_PUBLIC_SERVER_URL || 'http://localhost:5000';
+
```
Sboonny, line 9: I am getting hydration error, asking me to check.  https://nextjs.org/docs/messages/react-hydration-error

Does adding these lines change how the type of the page? That's why I am getting hydration error?

And do I add `useEffect`, know how `useEffect` works, it feels overkill for something that should be update once 🤔

cypress/e2e/dashboard/events/events-index.cy.ts
```diff
@@ -194,9 +195,10 @@ describe('spec needing owner', () => {
   });
 
   it('editing event updates cached events on home page', () => {
-    cy.visit('');
+    cy.login('').visit('');
+
     cy.get('button[aria-label="Options"]').click();
-    cy.findByRole('menuitem', { name: 'Events' }).click();
+    cy.findByRole('a[href="/events/"]').click();
```
Sboonny, line 201:
```suggestion
```

client/src/modules/policy/pages/policyPage.tsx
gikf: I don't know why exactly hydration error gives troubles here. But after putting the _Authenticate with Google_ behind waiting for loading user and checking permission for authentication with google, the issue doesn't appear anymore.

gikf:
I also noticed console error caused by `LinkButton` 
```
Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.
```

client/src/components/PageLayout/Header.tsx
```diff
@@ -92,7 +95,7 @@ export const Header: React.FC = () => {
           <HStack as="nav">
             <Box>
               {!user ? (
-                <LoginButton login={login()} text="Log in" />
+                <LoginButton login={login} text="Log in" />
```
Sboonny, line 98:
```suggestion
                <LoginButton login={login()} text="Log in" />

File: client/src/components/PageLayout/Header.tsx [hunk 4]
```diff
@@ -136,7 +139,7 @@ export const Header: React.FC = () => {
                       </Flex>
                     </MenuList>
                   </Menu>
-                  <LoginButton login={logout().then(goHome)} text="Log out" />
+                  <LoginButton login={logoutGoHome} text="Log out" />
```
Sboonny, line 142:
```suggestion
                  <LoginButton login={logout().then(goHome)} text="Log out" />

**Review by Sboonny — COMMENTED** [review]
I don't know how this is possible, but if you used the code below you will create infinite loop

File: client/src/modules/home/index.tsx [hunk 5]
```diff
@@ -44,6 +45,13 @@ const Home = () => {
       <GridItem colSpan={{ base: 2, xl: 1 }}>
         <VStack align="flex-start">
           <Heading>Upcoming events</Heading>
+          <Link
+            href={`/events/`}
+            _hover={{ color: 'blue' }}
+            textDecoration={'underline'}
+          >
+            View events list
+          </Link>
```
ojeytonwilliams, line 54: I like the extra way to get to `/events`, it's always felt like a chore to have to click 'click for more'.  What do you think to making the heading above into the link?  It's neater, but I'm wondering if it's obvious enough that that's what it does.

Same principle applies to the `/chapters` link.

client/src/modules/chapters/pages/chaptersPage.tsx
```diff
@@ -24,6 +25,13 @@ export const ChaptersPage: NextPage = () => {
           </LinkButton>
         )}
       </Flex>
+      <Link
+        href={`/events/`}
+        _hover={{ color: 'blue' }}
+        textDecoration={'underline'}
+      >
+        View events list
+      </Link>
```
ojeytonwilliams, line 34: I'm wondering what user journey this helps enable.  I totally get the desire to have links from the home page, but if a user has arrived on the chapters page, I would assume they're just interested in seeing the chapters.

ojeytonwilliams (changes requested):
I'm not sure the menu is *that* cluttered, yet. We only have 4 or 5 items, depending on which user is logged in.  freeCodeCamp proper has 10 (one of which is a sub-menu), so I think we've got a bit of room to grow.

That said, maybe some grouping would help it feel less cluttered?  i.e. public stuff (events, chapters, profile), admin stuff (dashboard, auth with google) and logout.  What do you reckon?

PR #1674, opened 2022-10-07
refactor: simplify updateEvent resolver

ojeytonwilliams:
- refactor: simplify updateEvent using helpers
- refactor: break down updateEvent some more

`updateEvent` had got a bit too unwieldy, so this PR breaks it into more manageable chunks.


gikf (approved):
Oh yes, this is nice :+1:

ojeytonwilliams:
@gikf today I discovered https://eslint.org/docs/latest/rules/complexity and I've been using it to find the most convoluted functions. It's pretty sweet!

gikf:
@ojeytonwilliams You made me curious to check how the scores look like. I honestly thought there will be more files affected by high complexity.

ojeytonwilliams:
Me too. Turns out we're not *terrible* on that front, which is nice. However, it's a pretty simple metric: how branched is a function? That doesn't catch anything like massive components should be broken up, unless they also have a lot of conditionals.

Issue #1621, opened 2022-09-28
Modifying a user's roles is laborious


ojeytonwilliams:
This has always been the case, but it's made more explicit by https://github.com/freeCodeCamp/chapter/pull/1620 which differentiates between the instance roles `chapter_administrator` and `member`. Previously admins just made do with `member` instance permissions.

While we only have three distinct _user roles_  `instance owner`, `chapter administrator` and `member`, those are defined by their combination of instance and chapter roles:

| User   |      Instance      |  Chapter |
|----------|:-------------:|------:|
| member |  member | member (when they join) |
| admin |    chapter_administrator   |   administrator |
| owner | owner | nothing |
    
This means that an `owner` wishing to promote a `member` to a `chapter administrator` must change the user's roles twice and those changes have to be executed on two separate pages.  Since a `chapter_administrator` that is also an instance `member` is not a real role, it should not be possible to make that happen.  Instead the api should only allow changes between the three _user roles_, and the UI should reflect that.

gikf:
The basic, messy way I imagined so far:

As chapter administrator needs to be connected to specific (or multiple chapters) it's logical to still set it from chapter users.

Chapter roles in UI:
- member
- administrator - resolver sets also instance role to chapter_administrator

Instance roles in UI:
- member
- owner
- chapter_administrator - It probably should be visible on the list, but should not be possible to be set directly as the instance role. Changing from it to other instance roles also can be spotty.

ojeytonwilliams:
> As chapter administrator needs to be connected to specific (or multiple chapters) it's logical to still set it from chapter users.

Agreed. I can't think of a better place.

> chapter_administrator - It probably should be visible on the list...

I'm not sure about that. We, as maintainers, know that the chapter_administrator role is a necessary evil, but do users need to know or care about it?

gikf:
That's a good question. I'm not too set on it. Currently it's a bit easier for me to think of them as visible, rather than looking for ways to keep them hidden.

ojeytonwilliams:
Yeah, its awkward.  I could be missing something, but if we have the logic to handle making it impossible to change to a `chapter_administrator`, we're most of the way to making them hidden, too.  Probably... I've not thought too deeply about this yet!

PR #1679, opened 2022-10-08
feat: internal chapter_administrator instance role

gikf:

Reference #1621

---
- Makes chapter_administrator instance role an internal role.
- For convenience changing roles relies now on the role names, instead of ids.
- Client doesn't send request if role isn't actually changed.
- Cypress
  - Adds few fixtures.
  - Adds task for getting user.
  - Adds custom command changing instance role.
  - Adds util to expect no errors in response. I was confused for a while when one request was quietly erroring.
- Uses chapter and instance roles enums imported from factories.

### DIFF
--- a/cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
+++ b/cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
@@ -25,6 +25,23 @@ describe('Chapter Administrator', () => {
     cy.joinChapter(firstChapterId);
   });
 
+  function confirmInstanceRole(email: string, expectedRole: string) {
+    cy.task<User>('getUser', email).then(({ instance_role: { name } }) => {
+      expect(name).to.eq(expectedRole);
+    });
+  }
+
+  function applyChapterRoleChanges(
+    userId: number,
+    changes: { chapterId: number; roleName: string }[],
+  ) {
+    for (const { chapterId, roleName } of changes) {
+      cy.changeChapterUserRole({ chapterId, userId, roleName }).then(
+        expectNoErrors,
+      );
+    }
+  }
+
   function changeChapterRole({
     userEmail,
     initialInstanceRole,
@@ -61,23 +78,6 @@ describe('Chapter Administrator', () => {
       );
       confirmInstanceRole(userEmail, expectedRole);
     });
-  }
-
-  function confirmInstanceRole(email: string, expectedRole: string) {
-    cy.task<User>('getUser', email).then(({ instance_role: { name } }) => {
-      expect(name).to.eq(expectedRole);
-    });
-  }
-
-  function applyChapterRoleChanges(
-    userId: number,
-    changes: { chapterId: number; roleName: string }[],
-  ) {
-    for (const { chapterId, roleName } of changes) {
-      cy.changeChapterUserRole({ chapterId, userId, roleName }).then(
-        expectNoErrors,
-      );
-    }
   }
 
   describe('when promoted to administrator of chapter', () => {
--- a/cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
+++ b/cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
@@ -29,17 +29,17 @@ describe('Chapter Administrator', () => {
     userEmail,
     initialInstanceRole,
     changesToApply,
-    expectedRole,
+    expectedInstanceRole,
   }: {
     userEmail: string;
     initialInstanceRole: string;
     changesToApply: { chapterId: number; roleName: string }[];
-    expectedRole: string;
+    expectedInstanceRole: string;
   }) {
     cy.task<User>('getUser', userEmail).then(({ id, instance_role }) => {
       expect(instance_role.name).to.eq(initialInstanceRole);
       applyChapterRoleChanges(id, changesToApply);
-      confirmInstanceRole(userEmail, expectedRole);
+      confirmInstanceRole(userEmail, expectedInstanceRole);
     });
   }
 
@@ -89,7 +89,7 @@ describe('Chapter Administrator', () => {
         changesToApply: [
           { chapterId: firstChapterId, roleName: chapterRoles.ADMINISTRATOR },
         ],
-        expectedRole: instanceRoles.CHAPTER_ADMINISTRATOR,
+        expectedInstanceRole: instanceRoles.CHAPTER_ADMINISTRATOR,
       });
     });
 
@@ -101,7 +101,7 @@ describe('Chapter Administrator', () => {
         changesToApply: [
           { chapterId: firstChapterId, roleName: chapterRoles.ADMINISTRATOR },
         ],
-        expectedRole: instanceRoles.OWNER,
+        expectedInstanceRole: instanceRoles.OWNER,
       });
     });
   });
@@ -115,7 +115,7 @@ describe('Chapter Administrator', () => {
         changesToApply: [
           { chapterId: firstChapterId, roleName: chapterRoles.ADMINISTRATOR },
         ],
-        expectedRole: instanceRoles.CHAPTER_ADMINISTRATOR,
+        expectedInstanceRole: instanceRoles.CHAPTER_ADMINISTRATOR,
       });
       changeChapterRole({
         userEmail: users.testUser.email,
@@ -123,7 +123,7 @@ describe('Chapter Administrator', () => {
         changesToApply: [
           { chapterId: firstChapterId, roleName: chapterRoles.MEMBER },
         ],
-        expectedRole: instanceRoles.MEMBER,
+        expectedInstanceRole: instanceRoles.MEMBER,
       });
     });
 
@@ -137,7 +137,7 @@ describe('Chapter Administrator', () => {
           { chapterId: firstChapterId, roleName: chapterRoles.ADMINISTRATOR },
           { chapterId: secondChapterId, roleName: chapterRoles.ADMINISTRATOR },
         ],
-        expectedRole: instanceRoles.CHAPTER_ADMINISTRATOR,
+        expectedInstanceRole: instanceRoles.CHAPTER_ADMINISTRATOR,
       });
       changeChapterRole({
         userEmail: users.testUser.email,
@@ -145,7 +145,7 @@ describe('Chapter Administrator', () => {
         changesToApply: [
           { chapterId: firstChapterId, roleName: chapterRoles.MEMBER },
         ],
-        expectedRole: instanceRoles.CHAPTER_ADMINISTRATOR,
+        expectedInstanceRole: instanceRoles.CHAPTER_ADMINISTRATOR,
       });
     });
 
@@ -159,7 +159,7 @@ describe('Chapter Administrator', () => {
           { chapterId: firstChapterId, roleName: chapterRoles.ADMINISTRATOR },
           { chapterId: secondChapterId, roleName: chapterRoles.ADMINISTRATOR },
         ],
-        expectedRole: instanceRoles.CHAPTER_ADMINISTRATOR,
+        expectedInstanceRole: instanceRoles.CHAPTER_ADMINISTRATOR,
       });
       changeChapterRole({
         userEmail: users.testUser.email,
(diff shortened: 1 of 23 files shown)


cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
```diff
@@ -0,0 +1,237 @@
...
+  before(() => {
+    cy.fixture('chapterRoles').then((r) => {
+      chapterRoles = r;
+    });
+    cy.fixture('instanceRoles').then((r) => {
+      instanceRoles = r;
+    });
+    cy.fixture('users').then((u) => {
+      users = u;
+    });
```
Sboonny, line 20: Do `r` and `u` stand for the JSON files?

I'm wondering if it's good idea to change them to something descriptive

Sboonny (approved):
well, https://docs.cypress.io/api/commands/fixture#Syntax is descriptive, so I am just nitpicking, which I am okay with it being ignored.

cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
gikf: In a way yes. It's what cypress passes from fixture file, to the `.then` method. I didn't spend long on thinking about their names. The most obvious `chapterRoles`, `instanceRoles`, `users` can't be used to be able to assign to these variables from the outer scope.

I guess I could do `.then((fixture) => {(...)`.

Sboonny: I am happy with `fixture`, it hints that people should look up `.fixture` to know what is going on, and it's simple 👍

gikf: Could you confirm whether to change it, or not? I'm slightly confused now.

Sboonny: Sure, my intention was to change it, but I was worried that it would be over complicate something simple, so I resolved it. (something weird like `userRoleInJson` or my old suggestions)

I am happy with simple `.then((fixture) => {(...)` as it isn't complicated, and it's a good hint, if you want to add it, please do.

```suggestion
    cy.fixture('chapterRoles').then((fixture) => {
      chapterRoles = fixture;
    });
    cy.fixture('instanceRoles').then((fixture) => {
      instanceRoles = fixture;
    });
    cy.fixture('users').then((fixture) => {
      users = fixture;
    });
```

ojeytonwilliams:
@gikf I probably won't have a chance to properly go over this until Wednesday, but at first glance this looks great.  Thanks a ton for putting this together!

gikf:
@ojeytonwilliams Sure, no rush!

cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
```diff
@@ -0,0 +1,237 @@
...
+    cy.fixture('chapterRoles').then((fixture) => {
+      chapterRoles = fixture;
+    });
+    cy.fixture('instanceRoles').then((fixture) => {
+      instanceRoles = fixture;
+    });
+    cy.fixture('users').then((fixture) => {
+      users = fixture;
+    });
+  });
```
ojeytonwilliams, line 21: I'm not sure this will necessarily cause issues, but the Cypress docs discourage use of variables like this.  The reasoning is that `cy.whatever` calls are generally asynchronous and so can cause fun bugs if you access them in places they're not guaranteed to exist.

Which is all to say: this code is probably fine, but not a best practice.  Could you refactor it to use aliases instead?  https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Aliases covers this particular use-case quite well.

cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  function confirmInstanceRole(email: string, expectedRole: string) {
+    cy.task<User>('getUser', email).then(({ instance_role: { name } }) => {
+      expect(name).to.eq(expectedRole);
+    });
+  }
+
+  function sequenceChangeChapterUserRole(
```
ojeytonwilliams, line 227: How about just `applyChapterRoleChanges`?   I wasn't sure what 'sequence' meant in this content and it looks like it's just 'do this' :)

server/src/controllers/ChapterUser/resolver.ts
```diff
@@ -14,9 +14,71 @@ import { ResolverCtx } from '../../common-types/gql';
...
+    const isStillAdmin = chapterUser.user.user_chapters.some(
+      (chapter_user) =>
+        chapter_user.chapter_id !== chapterUser.chapter_id &&
+        chapter_user.chapter_role.name === ChapterRoles.administrator,
+    );
+
+    if (!isStillAdmin) return InstanceRoles.member;
+  }
+  return oldInstanceRole;
+};
```
ojeytonwilliams, line 80: I think this deserves one or two unit tests.  It looks solid, but there's enough complexity that I think some simple checks would help.  Plus it's a nice, well-behaved function with no hidden dependencies, so it's easy to test!

server/src/controllers/Users/resolver.ts
```diff
@@ -1,47 +1,102 @@
...
+  if (oldRole === InstanceRoles.owner) {
+    const isAdmin = user.user_chapters.some(
+      (chapter_user) =>
+        chapter_user.chapter_role.name === ChapterRoles.administrator,
+    );
+
+    if (isAdmin) return InstanceRoles.chapter_administrator;
+  }
+  return newRole;
+};
```
ojeytonwilliams, line 49: Similarly, this might be worth a unit test or two.  Mostly to document what this does and why.

server/src/controllers/Users/resolver.ts
```diff
@@ -1,47 +1,102 @@
...
+  newRole: string;
+  user: UserWithUserChapters;
+}
+
+const getRoleName = ({ oldRole, newRole, user }: ChangeRoleNameData) => {
+  if (
+    oldRole === InstanceRoles.chapter_administrator &&
+    newRole === InstanceRoles.member
+  )
+    return InstanceRoles.chapter_administrator;
```
ojeytonwilliams, line 38:
```suggestion
  // The client should not request this change, but if it does it's ignored.
  if (
    oldRole === InstanceRoles.chapter_administrator &&
    newRole === InstanceRoles.member
  )
    return InstanceRoles.chapter_administrator;
```
Some context would help

server/src/controllers/Users/resolver.ts
```diff
@@ -1,47 +1,102 @@
...
+
```
ojeytonwilliams, line 39:
```suggestion
  // since the former owner may still be an administator, we check if they will
  // need the chapter_administrator role
```

server/src/controllers/Users/resolver.ts
```diff
@@ -1,47 +1,102 @@
...
-            instance_role_permissions: {
-              include: { instance_permission: true },
-            },
-          },
-        },
-      },
+      include: instanceRoleInclude,
+    });
+
+    const usersWithReplacedAdministrator = users.map((user) => {
```
ojeytonwilliams, line 61:
```suggestion
    // The chapter_administrator role is internal, so should not be displayed
    const usersWithReplacedAdministrator = users.map((user) => {
```

ojeytonwilliams (changes requested):
Amazing work, as always @gikf.

I made a few nitpicks and general suggestions below, but this is already looking great.

cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
gikf: Thanks for the idea, I wasn't happy with it and your version is better.

server/src/controllers/ChapterUser/resolver.ts
gikf: I've most likely went a bit overboard with the unit tests...

I had issues with making appropriate `chapterUser` objects for tests, so I've refactored it to use simpler type as argument. Names of both these functions are a bit iffy, but so far I haven't found anything short for it.

cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
gikf: I'm not entirely sure, but this might be slight exception https://docs.cypress.io/api/commands/fixture is linking to two pages using this approach https://glebbahmutov.com/blog/fixtures-in-custom-commands/ and https://github.com/cypress-io/testing-workshop-cypress/tree/master/cypress/integration/14-fixtures

I've started to change for using aliases and to be honest that's going to be painful. There's couple reasons for that:
- Resetting of aliases after each test, so fixtures would need to be loaded in the `beforeEach`. 
- As we have three fixtures here, that's three `cy.get('@alias').then(...)` in each test.
- Additionally to using `cy.get`, object with fixture passed to `.then` method doesn't have correct type by default.

What do you think?

server/src/controllers/ChapterUser/resolver.ts
ojeytonwilliams: > I've most likely went a bit overboard with the unit tests...

Not necessarily.  I generally aim for 100% coverage when I'm creating a new function, just so I know when to stop.  If what you've added is necessary to get to 100% that's totally fine. 

If you can cut a few tests/fixtures and still be at 100%, that's helpful (since there's less overall to maintain), but it's not too crucial.

> I had issues with making appropriate chapterUser objects for tests, so I've refactored it to use simpler type as argument.

As long as there's no need to assert types, it's fine.

cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
```diff
@@ -0,0 +1,237 @@
...
+          expect(instance_role.name).to.eq(instanceRoles.MEMBER);
+          applyChapterRoleChanges(id, [
+            { chapterId: firstChapterId, roleName: chapterRoles.ADMINISTRATOR },
+          ]);
+        },
+      );
+      confirmInstanceRole(
+        users.testUser.email,
+        instanceRoles.CHAPTER_ADMINISTRATOR,
+      );
```
ojeytonwilliams, line 59: This is a common pattern.  They mostly take in an email, initial role, list of changes and expected roles.  It feels like it should be a function and then these tests should all become one or two calls to that function.

What do you reckon?  Apologies if I underestimated the complexity!

server/src/controllers/Users/resolver.ts
```diff
@@ -4,44 +4,73 @@ import { prisma } from '../../prisma';
...
+      return userWithReplacedRoleName;
     });
+    return usersWithReplacedAdministrator;
   }
 
   @Authorized(Permission.UserInstanceRoleChange)
   @Mutation(() => UserWithInstanceRole)
   async changeInstanceUserRole(
-    @Arg('roleId', () => Int) roleId: number,
+    @Arg('roleName', () => String) roleName: string,
```
ojeytonwilliams, line 45:
```suggestion
    @Arg('roleName', () => String) newRole: string,
```
Same nitpick as before

server/src/controllers/Users/resolver.ts
```diff
@@ -4,44 +4,73 @@ import { prisma } from '../../prisma';
...
+    const user = await prisma.users.findUniqueOrThrow({
       where: { id: userId },
       include: {
+        ...instanceRoleInclude,
+        user_chapters: { include: { chapter_role: true } },
+      },
+    });
+
+    const oldRole = user.instance_role.name;
+    const newRole = roleName;
```
ojeytonwilliams, line 57:
```suggestion
```

ojeytonwilliams (changes requested):

cypress/e2e/dashboard/chapters/chapter-administrator.cy.ts
gikf: You are right! It was hiding from me before.

ojeytonwilliams (approved):
Thanks @gikf this looks great.

I think we will (at some point) need to prune our tests a bit. They're still running reasonably quickly at the moment, but I suspect we're over-testing and they're a little brittle.

I'm totally okay with these tests, mind, it's just that some spring cleaning will be needed periodically!

PR #1686, opened 2022-10-10
refactor: tidy and separate concerns of venueform

ojeytonwilliams:
- refactor: DRY venue update and creation
- refactor: clean up default fallbacks
- refactor: use returned chapter_id
- refactor: handle multiple loading errors
- refactor: move data-fetching into page components

The forms are still a bit scary, but this teases things apart slightly.

### DIFF
--- a/client/src/modules/dashboard/Venues/pages/ChapterNewVenuePage.tsx
+++ b/client/src/modules/dashboard/Venues/pages/ChapterNewVenuePage.tsx
@@ -1,25 +1,16 @@
-import { useRouter } from 'next/router';
 import React, { ReactElement, useEffect, useState } from 'react';
 
-import {
-  useChapterLazyQuery,
-  useCreateVenueMutation,
-} from '../../../../generated/graphql';
+import { useChapterLazyQuery } from '../../../../generated/graphql';
 import { DashboardLoading } from '../../shared/components/DashboardLoading';
 import { Layout } from '../../shared/components/Layout';
-import VenueForm, { VenueFormData } from '../components/VenueForm';
-import { VENUES } from '../graphql/queries';
+import VenueForm from '../components/VenueForm';
 import { useParam } from '../../../../hooks/useParam';
 import { NextPageWithLayout } from '../../../../pages/_app';
+import { useSubmitVenue } from '../utils';
 
 export const ChapterNewVenuePage: NextPageWithLayout = () => {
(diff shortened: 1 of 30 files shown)


client/src/modules/dashboard/Venues/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+        },
+      });
+      if (venue.data) {
+        router.replace(`/dashboard/venues/${venue.data.createVenue.id}`);
+      }
+    } catch (err) {
+      console.error(err);
+    } finally {
+      setLoading(false);
+    }
```
gikf, line 32: A bit side note, from the observing form behavior when it's submitted, for the other comment. When submission is successful there's a moment before redirect, when loading is set back to `false`, this can be seen in the form and submit button (not using loading text anymore).

That made me wonder if setting loading always back to `false` is desired, after all if there's redirection it shouldn't matter. Only if we are staying on the page enabling form again is needed. I've looked on some other pages and few `await`s for the `router.replace` what makes the issue not relevant.

gikf (commented):
Things are much simpler :+1: 

There's few details touched in specific comments.

client/src/modules/dashboard/Venues/pages/ChapterNewVenuePage.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <VenueForm
+      loading={loading}
+      chapterData={data}
+      onSubmit={onSubmit}
+      submitText={'Add venue'}
+      chapterId={chapterId}
+      loadingText={'Adding venue'}
+    />
```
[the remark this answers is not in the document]
ojeytonwilliams: Very good point.  I just saw loading and didn't think enough about what it was for!

Since `setLoading` just affects the VenueForm, it seems pointless to pass it around.

client/src/modules/dashboard/Venues/utils.ts
ojeytonwilliams: I see what you mean and the current behaviour doesn't seem quite right.

It's a touch better if it back to `false` if there's an error, but leave it otherwise (as you say, it's pointless if we're redirecting).  At least they can change something and try again. It's still bad UX, but slightly less bad than if the form becomes obviously stuck.

gikf (approved):
Thanks for updates :+1:

PR #1687, opened 2022-10-10
chore: add a problem and a solution for sync .env file

Sboonny:

As mentioned in discord, this aims to hint about the changes in `.env` file

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -592,6 +592,9 @@ If your problem isn't resolved in the sections below, then visit our [chat](http
   > *Invalid'prisma_1.prisma.chapters.findMany()* </br>
 
   **Solution:** The [database needs to be initialized](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md#initializing-the-database). Run `npm run db:reset` to clear and re-create the database tables.
+
+  * **Problem:**  I have come back after a long period of time, and chapter isn't working. </br>
+    **Solution:** In chapter development, `.env.example` has changed quite a bit. Be sure to sync up your .env with .env.example or things won't work until you do.
 </details>
 
 <details>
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -592,6 +592,9 @@ If your problem isn't resolved in the sections below, then visit our [chat](http
   > *Invalid'prisma_1.prisma.chapters.findMany()* </br>
 
   **Solution:** The [database needs to be initialized](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md#initializing-the-database). Run `npm run db:reset` to clear and re-create the database tables.
+
+  * **Problem:**  I have come back after a long period of time, and chapter isn't working. </br>
+    **Solution:** In chapter development, `.env.example` has changed quite a bit. Be sure to sync up your .env with .env.example or things won't work until you do.
```
gikf, line 597:
```suggestion
* **Problem:**  I have come back after a long period of time, and chapter isn't working. </br>
  **Solution:** In chapter development, `.env.example` has changed quite a bit. Be sure to sync up your .env with .env.example or things won't work until you do.
```
Indentation makes it appear as a subsection of the previous solution.

Wording could be slightly more pointing this as a _possible_ solution, but I don't have any specific idea how to change that.

Sboonny: We can change `or things won't work until you do`, to something like `so the codebase could run smoothly`.

```suggestion
  **Solution:** In chapter development, .env.example has changed quite a bit. Be sure to sync up your .env with .env.example so the codebase could run smoothly.
```

ojeytonwilliams: It's more that we don't know if the example has changed or if that's the problem they're facing.

How about this: "As we develop Chapter it is often necessary to change the environment variables we use. In case your .env is out of date, try replacing it with .env.example"?

gikf (approved):

PR #1696, opened 2022-10-11
fix: rsvp when clicking on invitation

ojeytonwilliams:
- refactor: replace useAuthStore with useAuth
- refactor: slight tweaks to eventPage
- test: add tests for rsvp/cancel via email
- refactor: use functions so they're hoisted
- fix(test): target the right text
- fix: rsvp automatically when using invitation link

This started off life as a refactor, but I realised that the rsvp flow was broken when you arrived via the link in the invitation.  And that there was a regression (probably caused by me!) at some point. Variable hoisting turns out to be confusing! 

Anyways, all should be well now.

### DIFF
--- a/client/src/components/PageLayout/Header.tsx
+++ b/client/src/components/PageLayout/Header.tsx
@@ -7,7 +7,7 @@ import { useRouter } from 'next/router';
 import React, { forwardRef } from 'react';
 import NextLink from 'next/link';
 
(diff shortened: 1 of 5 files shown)


gikf (approved):
Seems to be all good :+1:

I wasn't sure if this should be working for not logged in user as well, but then I found open issue #796 that this part is not expected yet.

PR #1708, opened 2022-10-13
fix: change confirm button text to 'yes, im sure'

### LINKED ISSUE #1705: better to improve text of are-you-sure-to-delete-your-data  modal
### Describe the bug:
Currently modal has too many repetitive text:

Screen Shot 2022-10-13 at 12 41 06

I think 'confirm' button better to have 'Yes, im sure!' text

- Browser(s) name and version: chrome
- Operating System: mac

shootermv:

Closes #1705

### DIFF
--- a/client/src/modules/profiles/pages/userProfile.tsx
+++ b/client/src/modules/profiles/pages/userProfile.tsx
@@ -21,7 +21,7 @@ export const UserProfilePage = () => {
   const logout = useLogout();
   const router = useRouter();
 
-  const confirmDelete = useConfirmDelete({ doubleConfirm: true });
+  const confirmDelete = useConfirmDelete({ buttonText: 'Yes, im sure' });
   const [deleteMe] = useDeleteMeMutation();
   const [updateMe] = useUpdateMeMutation({
     refetchQueries: [{ query: meQuery }],
--- a/client/src/modules/profiles/pages/userProfile.tsx
+++ b/client/src/modules/profiles/pages/userProfile.tsx
@@ -21,7 +21,10 @@ export const UserProfilePage = () => {
   const logout = useLogout();
   const router = useRouter();
 
-  const confirmDelete = useConfirmDelete({ buttonText: 'Yes, im sure' });
+  const confirmDelete = useConfirmDelete({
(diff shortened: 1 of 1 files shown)


client/src/modules/profiles/pages/userProfile.tsx
```diff
@@ -21,7 +21,7 @@ export const UserProfilePage = () => {
   const logout = useLogout();
   const router = useRouter();
 
-  const confirmDelete = useConfirmDelete({ doubleConfirm: true });
+  const confirmDelete = useConfirmDelete({ buttonText: 'Yes, im sure' });
```
Sboonny, line 24: I don't know what happens but after deleting the using, 

auth with google is deleted

image

---

I have no issue with it, just wondering if we okay with single confirm?

Why did we have double confirm before?

Sboonny (changes requested):
this isn't about the file itself, but there is similar case in ChaptePage in the dashboard

here is the code

```ts
const clickDelete = async () => {
    const ok = await confirmDelete({ doubleConfirm: true });
    if (!ok) return;
    deleteChapter({ variables: { chapterId } });
    router.push('/dashboard/chapters');
  };
  ```
  if you change it too, that would be lovely 👍

File: client/src/modules/profiles/pages/userProfile.tsx (hunk 1 above)
Comment (gikf) [inline client/src/modules/profiles/pages/userProfile.tsx:24]:
```suggestion
  const confirmDelete = useConfirmDelete({
    body: 'Are you sure you want to delete your account? Account deletion cannot be reversed.',
    buttonText: 'Delete account',
  });
```
Could we change at the same time the default text? And maybe also make explicit that is account what is being deleted.

gikf: Is that instance owner logged in?

Sboonny: nope I deleted the instance owner 🤣

gikf: Exactly :slightly_smiling_face:

Sboonny (approved):

PR #1711, opened 2022-10-13
feat: polish chapter card

### LINKED ISSUE #1719: Event status does not appear on Chapter cards
Since Chapter cards show all events, it should be obvious when those events are already cancelled.

For example

image

Batz Group is cancelled, but that can't be seen on the Chapter card.

Considerations: we could hide canceled events by default, since most users will not be interested in cancelled events. That's a little more involved, though.

### What we need

A visual indication that a particular event is canceled.  Probably just the word 'Canceled' in red and in the right place.

Sboonny:

Closes: #1719 

This PR centers the chapterPage and I wanted to add image back to test how it looks, but I can't make them look good which annoying me 

---

I have tried for long while to make this looks good, I can't wrap around how to make it good

image

should I revert adding the image and wait for later?


CONTRIBUTING.md
```diff
@@ -399,11 +399,6 @@ The database is seeded with several types of user.  To experiment with different
 - `npm run change-user foo@bar.com` (an _owner_ with full permissions)
 - `npm run change-user admin@of.chapter.one` (an _administrator_ of chapter 1)
 
-To experiment with different roles using the terminal, you can run these commands into your terminal:
-
-- `npm run change-user foo@bar.com` (an _owner_ with full permissions)
-- `npm run change-user admin@of.chapter.one` (an _administrator_ of chapter 1)
-
```
Sboonny, line 406: after 114 PR and I make this silly mistake 🤦

Sboonny:
while testing the style, I refreshed the page, and this cased a new image to be generated 

before refreshing 

image

After refreshing

image

---

is this something to worry about, or is it intentional?

ojeytonwilliams:
Aren't we inevitably going to run into a11y issues if we overlay an arbitrary image with text?  How do we guarantee there is enough contrast between the text and the image?

Sboonny:
And I have no idea how to fix it, because I have no control over the images.

Well I do have an idea, [ImageData](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/colorSpace) of Canvas API can open the door for checking the image colors, and displaying background gradient which is accessible with the image.

But I don't know if that's worth it, I think it's better to display less data and put the data in single small section, which make the context easier to control.

So the question is the image worth it? we will fetch less data from the database, and it may look more stylish.

or keep the same layout? we will fetch same data from the database, but we still have more option to make it more stylish. (I haven't manipulated fonts yet)

ojeytonwilliams:
It's more important that it's accessible than stylish.  If we can do both, awesome, but if we have to choose then it's got to be a11y.

> But I don't know if that's worth it, I think it's better to display less data and put the data in single small section, which make the context easier to control.

Sure, sounds reasonable.  I don't think we want the cards to be too busy, after all. It's just a summary, so it can get away with missing out some details.

If there's a strong enough gradient such that we can be confident the text is easy to read, that could work.  If not, how about just giving all text a plain, monochrome background?

allella:
Do we need the most recent six events? Perhaps top half the card is an image and the bottom half is the three most recent events?

Also, does "Passed" make sense for past event status?  I feel like it would be "Past" but I'm not a language geek.

ojeytonwilliams:
> Do we need the most recent six events? Perhaps top half the card is an image and the bottom half is the three most recent events on top of a solid background?

Sounds reasonable, yeah. @Sboonny what do you reckon?

> I feel like it would be "Past" but I'm not a language geek.

Neither am I, but I think you're right.

Sboonny:
I wanted to pull user data and show subscribed user, with avatars and stuff alike. Because people like know if this chapter has people who interested in it. 
 
This didn't give the needed info for the subscribed users, because they are already subscribed to that chapter and may need different info. So I was going to create logic to display the most upcoming event, if the current user is subscribed to said chapter.

But I like @allella idea more because people want to know if the chapter is still active. 🤔 

I will start implementing it. 

About `passed`, I will opened 2nd PR for it later 👍

Sboonny:
I can't find a change that could be made into a separate PR anymore, so I am closing this

PR #1714, opened 2022-10-14
refactor: always define before use

ojeytonwilliams:
This avoids an annoying class of bugs:

f()
var a = 'defined'
function f() { // does something with a }

f looks as if it should be able to use a, but hoisting means that's not always the case.

This is quite relevant for our code, since we often need to useEffect before a conditional return, but rely on variables that are only guaranteed to be defined after it.


gikf (approved):
Great :+1:

Issue #1707, opened 2022-10-13
Logo for the chapter page and refactor the image_url field


Sboonny:
Currently, the image_url act as banner, knowing that chapter are non-profit organizations, they may need logo as well.

This will make us change three things 

- [x] refactor image_url field in chapter to banner_url
- [x] create logo_url field 
- [ ] use logo_url to display chapter logo

----

(Optional) 

Images can be any height and any width, should we optimize them using services like Amazon s3 https://aws.amazon.com/blogs/compute/resize-images-on-the-fly-with-amazon-s3-aws-lambda-and-amazon-api-gateway/

Or limit the size the user can submit?

ojeytonwilliams:
There's no need to host images for the MVP. It's fine to use links for now.

Sboonny:
should the logo field be added for mvp?

ojeytonwilliams:
Absolutely.  As long as not too much work is involved, small tweaks like this seem fine to me.

PR #1717, opened 2022-10-14
feat: add chapter logo image

Sboonny:

ref #1707


server/prisma/schema.prisma
```diff
@@ -21,6 +21,7 @@ model chapters {
   city          String
   region        String
   country       String
+  logo_url      String
   banner_url    String
```
ojeytonwilliams, line 25: Thinking about it, shouldn't these be optional?  Also I'm wondering if we should be providing default images at all.  It's probably better to fallback to displaying nothing rather than showing an image that the user wasn't expecting to be there.

ojeytonwilliams (commented):
This looks good.  It did get me thinking about the default images though:

server/prisma/schema.prisma
Sboonny: GitHub provides a default image creating the organization, here is my new organization, so the main stream is to provide default image. 
https://github.com/Messy-land

but we don't have library, that provided generic images. So I am leaning toward it being optional. Post MVP we can discuss this in a meeting to see how the others feel

ojeytonwilliams: That makes sense.  A random abstract image should work well, but, yeah, let's discuss that later.

I think, for now, it would be good if we made the various images optional.  Not showing the image should be an okay fallback for the banners and maybe the Chakra Avatar as a logo fallback?

Sboonny: I have added the optional schema. I am planning to add the logo after this is merged, because locally it's started to act weird.

I am getting error in `\server\src\controllers\Events\resolver.ts` and `server\src\controllers\Chapters\resolver.ts` in fields related to UserBan. 

Reseting the db doesn't fix it, so I am wondering if you can check these files and see if they are getting errors while reviewing it again.

Sboonny: Random thoughts 💭🤔, maybe a merged pr is changing the required dB. But that doesn't explain why it fails locally

ojeytonwilliams: Making the field optional changes what comes back from the database. That field can then be null. The return types of the prisma queries change to reflect this.

This causes a problem because that returned data is then supposed to be sent back to the client in response to a graphql query. However there's a mismatch between what we're claiming to send (Chapter, ChapterWithRelations and so on) and what we're actually sending (the result of the prisma query).

TL;DR: the return types of the resolvers need changing. If you run `npx tsc` inside `server` it should tell you which ones and why.

It might be necessary to run `npm run clean` from root first, though, so you might as well do that anyway.

Sboonny: > TL;DR: the return types of the resolvers need changing. If you run npx tsc inside server it should tell you which ones and why.

You have told me before, I think. 🤔 

Anyway the issue was there before I had changed the prisma schema, I have a guess of what happened. 


---

After refactoring the `image_url` to `banner_url`, I foolishly didn't update my main. So whenever I try to create new PR, update old one, I would need to reset my db.

I think while resetting my db multiple times, I got in a situation in which I had the old `image_url` and the new `banner_url` at the same time.

I think `npm run clean` cleared my schema, at least that what the terminal say, which deleted the corrupted database.


But this is 5 days ago, I may be misremembering, and it was human error all along 👍

ojeytonwilliams: Yeah, it's a pain. The db:reset isn't *too* slow, but it's a nuisance to have to remember it each time.

ojeytonwilliams (approved):
LGTM :+1:

Issue #1639, opened 2022-09-30
How chapter subscription affects event subscription


gikf:
Issue to discuss and iron out the if/how changing subscription to chapter should affect subscription to events from chapter.

#### Chapter subscription
- Subscribing to chapter doesn't change event subscriptions.
- Unsubscribing from chapter unsubscribes from all events in that chapter as well.
  - > Unsubscribing from this chapter will affect subscriptions of all existing events, and new events in the chapter.
  - > Unsubscribe Options
[Attend this event, but only turn off future notifications for this event]
Or, [stop receiving all notifications by unfollowing ${event.chapter.name}]
  - User can still subscribe (again, or first time) to individual events.

- Determines if user gets emails:
  - About new events.

#### Event subscription
- When RSVPing to an event, chapter subscription is used as initial event subscription. 
- Determines if user gets emails:
  - Event reminder.
  - When event is cancelled.
  - When event changes venue.
  - (administrators) When there's new RSVP.

If I missed something, feel free to edit it.

allella:
Post MVP, I could see unsubscribing from a _chapter_ impacting if an _organizer_ can send a general announcement to the _members_.

I'm not sure unsubscribing from a _chapter_ should impact any existing reservations / RSVP.  The _member_ should be able to easily disable those messages from their email's footer.

Just as an example, some chapters may have a larger annual event that someone wants to attend, but they may not want to get all the other event notifications.  If they RSVP to an event and then unsubscribe to stop getting new event notes, or general announcements, then they may miss important RSVP notes.

If we're going to disable event notifications for RSVP'ed members, then we'd want to remind them and perhaps even do a query and list the future RSVPs so they can see what events they are being removed from.

gikf:
I've added part that after unsubscribing from chapter, user still is able to subscribe to individual events, again or first time.

ojeytonwilliams:
While I think it's probably fine for the MVP, we will be sending quite a lot of notifications. Reason being, Google Calendar sends notifications independently.  These will be sent for the initial invitation as well as any updates to the event, including cancellations.

I imagine users will either want one or the other. We can experiment with their API, but I haven't had chance to see if we can control notifications at the user level or if it's set for the entire event. I don't think this has to be addressed immediately, but it's something to consider.

Anyway, back to Chapter notifications.

> I'm not sure unsubscribing from a chapter should impact any existing reservations / RSVP. The member will be able to easily disable those event / RSVP-related messages from their email's footer.

[...]

PR #1727, opened 2022-10-16
feat: add share button for events

### LINKED ISSUE #1721: Add links that let users join chapters and attend events
Currently, if I follow a link like http://localhost:3000/events/22?emaillink=true I will see a modal asking me to confirm that would like to join the event.

If would be helpful for administrators to be able to easily share this after creating a new event.  For example, the dashboard for that event could have a share button that shows the link and gives the user an option to copy it to their clipboard.

Similarly, if they want to invite people to join their Chapter it would be helpful to be able to do the same sort of thing from the Chapter's dashboard.

Nice to have:

Logged out users should be sent away to log in and then brought back to the shared link if they consent.

Sboonny:

Ref #1721


client/src/modules/dashboard/Events/components/Actions.tsx
```diff
@@ -72,6 +73,9 @@ const Actions: React.FC<ActionsProps> = ({
       <Button size="sm" colorScheme="blue" onClick={clickEmailAttendees}>
         Email Attendees
       </Button>
+      <ShareEventPop
+        link={`${process.env.CLIENT_LOCATION}/events/${event.id}?emaillink=true`}
```
ojeytonwilliams, line 77:
```suggestion
        link={`${process.env.NEXT_PUBLIC_CLIENT_URL}/events/${event.id}?emaillink=true`}
```
Sorry, the env vars are needlessly complicated. I shouldn't have stored `CLIENT_LOCATION` in two variables.

Anyway, this is the one you need.

client/src/modules/dashboard/Events/components/ShareEventPopOver.tsx
```diff
@@ -0,0 +1,35 @@
...
+          Share
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent>
+        <PopoverArrow />
+        <PopoverCloseButton />
+        <PopoverHeader>
+          You have copied the link below into your clipboard!
+        </PopoverHeader>
+        <PopoverBody>{link}</PopoverBody>
```
ojeytonwilliams, line 31: How about not showing the link?  It's a little prettier without it!

ojeytonwilliams (changes requested):

ojeytonwilliams:
@Sboonny were you planning to handle the 'share chapter' link in this PR?

Sboonny:
I want the discussion about the join and subscribing to chapter to end first, because I don't know what people want to share. 

Are they going to ask people to follow (subscirbe) to their chapter, or to join it?

But I may be lazy and add both of them, but if users grow attach to a feature removing it would be annoying :think:

To be fair, I may be overthinking this, this feels like I am overthinking it.

ojeytonwilliams:
> Are they going to ask people to follow (subscirbe) to their chapter, or to join it?

I'd just mirror what we do for events.  So, a link that triggers the "are you sure" modal for joining a chapter (though we should update the text, since we're triggering it automatically).

As for subscriptions, that's still being discussed https://github.com/freeCodeCamp/chapter/issues/1639 but this PR doesn't have to worry since it's just triggering the current behaviour.  i.e. making a request to the `joinChapter` resolver and it handles subscriptions.

ojeytonwilliams:
> I am not sure if the link works.

It won't, yet, but we can handle that in another PR.

ojeytonwilliams (approved):
LGTM :+1: Thanks, @Sboonny!

PR #1730, opened 2022-10-17
feat: re-implement is_bannable logic

gikf:

---

- Re-implements logic for the `is_bannable` field, determining if logged in user should be able to ban other specific user. That's used in client to hide/show the ban/unban button by specific user.
- Instead of at the chapter user resolver, the field is added in bulk to the dashboard chapter users query.
- Adds optional `is_bannable` field in `ChapterUser` graphql type.
- Moves `DASHBOARD_CHAPTER_USERS` from chapter queries file to dashboard chapter queries.
- ~~I don't remember specifics how the instance role / chapter role relation was supposed to be handled, this is the current table who can ban who in the re-implementation.~~ Updated table.

| can ban?\ban who?| chapter member | chapter admin | instance owner* |
|------------------|----------------|---------------|-----------------|
|  chapter member  |       no       |       no      |      no         |
|  chapter admin   |       yes      |       no      |      no         |
|  instance owner  |       yes      |      yes      |      no         |


ojeytonwilliams:
| can ban?\ban who?| chapter member | chapter admin | instance owner |
|------------------|----------------|---------------|-----------------|
|  chapter member  |       no       |       no      |      no         |
|  chapter admin   |       yes      |       no      |      no      |
|  instance owner  |       yes      |      yes      |    no     |

I think that's probably the way to go.  Allowing instance owners to ban each other from chapters doesn't feel right.  It doesn't have any affect on their permissions, it just boots them out of any events they were attending.  However, they can just rejoin if they wish.

How does that sound?

Sboonny:
Random question, this is re-implemented. Why was it deleted before?

server/src/graphql-types/ChapterUser.ts
```diff
@@ -21,4 +21,7 @@ export class ChapterUser {
 
   @Field(() => User)
   user: User;
+
+  @Field(() => Boolean)
```
ojeytonwilliams, line 25:
```suggestion
  @Field(() => Boolean, { nullable: true })
```
Otherwise we'd be implying this exists every time that `ChapterUser` is returned and that's not always true.

It's a pretty minor deal for booleans, but I think we should for the sake of consistency, if nothing else.

ojeytonwilliams (changes requested):
I had a good go with it locally and it's working flawlessly.  All I could find was one minor GQL type issue:

ojeytonwilliams (approved):
LGTM :+1:

PR #1733, opened 2022-10-18
feat: change Add Sponsor button colorscheme for consistency

Sboonny:

Add Sponsor is an action of labeling data, so I decided to make it use similar color as similar actions, instead of being colorless like cancel actions


gikf (approved):
Looks good :+1:

PR #1739, opened 2022-10-19
fix: only instrument code in test environments

ojeytonwilliams:

We had setup istanbul such that it always instrumented the code, even in production. This makes sure it only happens when `NODE_ENV=test`


Sboonny (approved):
That is an interesting branch name

ojeytonwilliams:
I've no idea why they called it that. It's not great when you want to search for it!

PR #1747, opened 2022-10-20
fix: render only when ready

ojeytonwilliams:
- fix: check isReady before rendering any page
- fix: update pages relying on isReady

There is a trade-off here.  All pages wait for `router.isReady` to be true before rendering, even if they don't need the router.  This makes pages like / and /chapters a little slower because they could have started rendering immediately, but now cannot.  However, when testing locally, I could see no difference between the two approaches.

### DIFF
--- a/client/src/modules/chapters/pages/chapterPage.tsx
+++ b/client/src/modules/chapters/pages/chapterPage.tsx
@@ -13,7 +13,7 @@ import {
 import { CheckIcon } from '@chakra-ui/icons';
 import { NextPage } from 'next';
 import NextError from 'next/error';
-import React, { useEffect } from 'react';
+import React from 'react';
 
 import { useConfirm } from 'chakra-confirm';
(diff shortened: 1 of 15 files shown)


client/src/modules/chapters/pages/chapterPage.tsx
```diff
@@ -136,7 +127,9 @@ export const ChapterPage: NextPage = () => {
     }
   };
 
-  const isLoading = loading || !isReady || !data;
+  const isLoading = loading || !data;
+
+  console.log(data, isLoading);
```
gikf, line 132:
```suggestion
```

gikf (approved):
Looks good :+1:

ojeytonwilliams:
I'm still not 100% convinced this is the perfect solution, but it does remove one headache.

Anyway, thanks for the QA!

PR #1750, opened 2022-10-20
fix: stop logging missing sessions

ojeytonwilliams:

Whenever the session is invalid, the only thing we can usefully do is delete the session.  This can happen for a variety of reasons (i.e. the user deletes their cookies or logs out from multiple tabs or browsers), none of which we need to log.

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -129,10 +129,9 @@ export const main = async (app: Express) => {
 
   // no need to check for identity provider's token on logout
   app.delete('/logout', (req, res, next) => {
-    if (!req.session) return next('session not found');
-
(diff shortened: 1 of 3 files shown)


Sboonny (approved):
Nicely done 👍

Sboonny:
btw where do I found more about `P2025`, can't find it in this docs https://www.postgresql.org/docs/current/errcodes-appendix.html

ojeytonwilliams:
@Sboonny that's a good point - the constants were a bit cryptic.  The comment I've added should help future devs who're curious where these came from.

gikf (approved):
Looks good :+1:

Issue #1713, opened 2022-10-14
As a user, I can download my personal information from Chapter


ojeytonwilliams:
Per https://gdpr-info.eu/art-15-gdpr/ a user has the right to access any of their personal information that we store.  Precisely what constitutes personal information is a bit subjective, so we should err on the side of inclusion.  

With that in mind I think we want data that is connected to their entry in the the `users` table.  Aside from the entry itself we'd want: the details of chapters they are in, their role in these chapters, the same for events, their ban status, and finally their instance_roles. I think that covers it, but let me know if I've missed something.

On the UX side, all we need is a button on their profile to download the data.  The data can be a JSON string.

TODO:
- [x] create a resolver that returns the user's data: https://github.com/freeCodeCamp/chapter/pull/1756
- [ ] create a button to fetch and download that data

allella:
Yeah, I guess the tables with "user" in the name and the things they join to.

I could see session potentially having some partial personal info if cookies are stored there. 

Is there anything in the database we'd see as being a security risk?  Being that it's an OSS project, the schema and roles and such are no secret.  I'm just wondering if there are any tokens or session related things which could be potentially harmful and worth excluding.

allella:
I posted an example of a popular US site's download interface at
https://github.com/freeCodeCamp/chapter/issues/263#issuecomment-1270169105

This uses a series of .csv files that are zipped up, which is likely more work than needed for MVP. I think I dumped fCC's data and it was a JSON blob that's sort of hard to read for the lay person, so I guess the format, like CSVs vs JSON, isn't decreeed?

ojeytonwilliams:
> I'm just wondering if there are any tokens or session related things which could be potentially harmful and worth excluding.

There are a few things that we must not leak.  

- Email addresses of anyone other than the user making the request
- The session id (just to be on the safe side)
- Anything in _google_tokens_

> I think I dumped fCC's data and it was a JSON blob that's sort of hard to read for the lay person, so I guess the format, like CSVs vs JSON, isn't decreed?

I don't believe so, but if we find out otherwise we can always come up with a nice approach.

PR #1756, opened 2022-10-21
feat(server): query for personal information

Jonathan1599:

Ref #1713 

Created `userInformation` query that returns user details regarding chapters they're in, chapter roles, events, bans and instance roles


gikf:
Hey @Jonathan1599, there's no need to keep merging every new change from `main` branch into the PR branch, PR should be possible to be merged cleanly without that. Unless there are merge conflicts requiring manual resolving, or PR branch is very outdated, all that does is start again the automated CI checks.

server/src/controllers/User/resolver.ts
```diff
@@ -49,6 +55,58 @@ export class UserWithInstanceRoleResolver {
     });
   }
 
+  @Query(() => UserInformation)
```
ojeytonwilliams, line 58:
```suggestion
  @Query(() => UserInformation, { nullable: true })
```
Unfortunately it's not enough to set the return type, we have to include `nullable: true` so that the schema reflects this.

server/src/controllers/User/resolver.ts
```diff
@@ -49,6 +55,58 @@ export class UserWithInstanceRoleResolver {
...
+  async userInformation(
+    @Ctx() ctx: ResolverCtx,
+  ): Promise<UserInformation | null> {
+    if (!ctx.user) {
+      return null;
+    }
+    return await prisma.users.findFirst({
```
ojeytonwilliams, line 65:
```suggestion
    return await prisma.users.findUnique({
```
Nitpick: this is more declarative since there can only be a single user record with this id.

ojeytonwilliams (changes requested):
Nice work @Jonathan1599!

I found a couple of very minor issues, but I think this is pretty much what we want.

ojeytonwilliams (approved):
This looks good to me. Thanks for taking this on, @Jonathan1599

ojeytonwilliams:
@allcontributors add @Jonathan1599 for code

PR #1766, opened 2022-10-23
feat: dashboards with admined content only

gikf:

To do
- [x] Ensure dashboard and non-dashboard queries are updated when needed.
- [x] Reduce repeating.
- [x] Check cache issue when relogging between accounts.
- [x] Fix conflicts.

---

- Chapters displayed in chapters dashboard are only admined chapters.
- Events displayed in events dashboard are only events belonging to admined chapters.
- Venues displayed in venues dashboard are only venues belonging to admined chapters.

Especially important places to look at:
- Cache when logging out and relogging to different account. Different query results returned depending on user, brings up potential issue that after logout, or reloggin to different account, cache shouldn't keep query results from different person. `client.resetStore()` is supposed to do that, but making sure that's happening definitely won't hurt.
- Logging out when on one of these dashboard pages. There was point when doing that would happen with some unexpected error. I think that's not going to happen again, but is another good area to take a closer look at.


cypress/support/util.ts
```diff
@@ -27,3 +27,19 @@ export function gqlOptions(body: any, additionalOptions?: any) {
...
 }
+
+export function expectHrefIdToBeInArray(link, ids: number[]) {
+  cy.wrap(link)
+    .invoke('attr', 'href')
+    .then((href) => {
+      const id = parseInt(href.match(/\d+/)[0]);
+      expect(ids).includes(id);
+    });
+}
```
ojeytonwilliams, line 38: With something like this, we can keep things nice and readable by defining helper functions, but leaving the final `expect` inside the test itself.

So, for this we could do:
```suggestion
export function getFirstPathParam(link: JQuery<HTMLElement>) {
  return parseInt(link.attr('href').match(/\d+/)[0]);
}
```

and I'll add a comment below with an example usage.

cypress/support/util.ts
```diff
@@ -27,3 +27,19 @@ export function gqlOptions(body: any, additionalOptions?: any) {
...
+
+export function expectOnlyObjectIdsInLinks(objects, linksDataCy) {
+  const ids = objects.map(({ id }) => id);
+  cy.get(`[data-cy=${linksDataCy}]`).each((link) =>
+    expectHrefIdToBeInArray(link, ids),
```
ojeytonwilliams, line 43:
```suggestion
    expect(ids).to.include(getFirstPathParam(link)),
```
Since chai lets you write pretty readably stuff, we don't really need to name the function that does this.

cypress/support/util.ts
```diff
@@ -27,3 +27,19 @@ export function gqlOptions(body: any, additionalOptions?: any) {
...
+  cy.wrap(link)
+    .invoke('attr', 'href')
+    .then((href) => {
+      const id = parseInt(href.match(/\d+/)[0]);
+      expect(ids).includes(id);
+    });
+}
+
+export function expectOnlyObjectIdsInLinks(objects, linksDataCy) {
+  const ids = objects.map(({ id }) => id);
```
ojeytonwilliams, line 41: Again, it might be better to not bother with a named function and, instead, just write the object map and expect code where they're used (i.e. events-index.cy.ts and venues.cy.ts)

Since the code is pretty short it doesn't feel terrible to repeat it. Also, you can use `const eventIds` or `const venueIds` to make it even more explicit what's going on.

server/src/util/adminedChapters.ts
```diff
@@ -0,0 +1,32 @@
...
+  };
+}
+
+export const isAdminFromInstanceRole = ({ instance_role }: User) =>
+  instance_role.instance_role_permissions.some(
+    ({ instance_permission }) =>
+      instance_permission.name === Permission.ChapterEdit,
+  );
+
+export const explicitlyAdminedWhere = (user_id: number) => ({
```
ojeytonwilliams, line 15:
```suggestion
export const isChapterAdminWhere = (user_id: number) => ({
```
Maaaybe... what do you think?

server/src/controllers/Events/resolver.ts
```diff
@@ -358,6 +363,25 @@ export class EventResolver {
     });
   }
 
+  @Query(() => [EventWithVenue])
+  async dashboardEvents(
+    @Ctx() ctx: Required<ResolverCtx>,
+    @Arg('limit', () => Int, { nullable: true }) limit?: number,
+    @Arg('showAll', { nullable: true }) showAll?: boolean,
```
ojeytonwilliams, line 370:
```suggestion
    @Arg('showAll', { nullable: true }) showAll?: boolean,
```
Do you have plans for this?

ojeytonwilliams (changes requested):
This is working perfectly - I couldn't find any issues with cached results sticking around and the various views are all good.

Minor style nitpicks only:

cypress/support/util.ts
gikf: Thanks for all these, those are great points. Before putting it in a bit convoluted, general function I had in fact ids named explicitly like this.

server/src/util/adminedChapters.ts
gikf: I like it more :rocket:

server/src/controllers/Events/resolver.ts
gikf: Right, not really.

gikf:
I remembered about logout when account is deleted, so I've added resetting resetting Apollo store there too.

ojeytonwilliams (approved):

ojeytonwilliams:
> I remembered about logout when account is deleted, so I've added resetting resetting Apollo store there too.

Nice catch!

PR #1769, opened 2022-10-24
feat: fix flashing img while loading image

### LINKED ISSUE #1755: Brief flash of fallback image
### Describe the bug:

The fallback image (with the dog) shows while the main image is loading:

SurpriseDog

Ideally the fallback should only show if the image fails to load.  Also, cute though the dog is, it's probably better to fallback to something boring like an empty space or something abstract.

Sboonny:

Closes #1755 

<s>we are having an issue, `fallbackstratagy` in chakra doesn't do the intended function.

image
</s>

Sorted Kudos to oliver

### DIFF
--- a/client/src/modules/chapters/pages/chapterPage.tsx
+++ b/client/src/modules/chapters/pages/chapterPage.tsx
@@ -159,7 +159,11 @@ export const ChapterPage: NextPage = () => {
           <Image
             boxSize="100%"
             maxH="300px"
-            src={data.chapter.banner_url}
+            src={
+              data.chapter.banner_url === ''
+                ? undefined
+                : data.chapter.banner_url
+            }
             alt=""
             borderRadius="md"
             objectFit="cover"
--- a/client/src/modules/chapters/pages/chapterPage.tsx
+++ b/client/src/modules/chapters/pages/chapterPage.tsx
@@ -155,22 +155,20 @@ export const ChapterPage: NextPage = () => {
   return (
     <VStack>
       <Stack w={['90%', '90%', '60%']} maxW="600px" spacing={6} mt={10} mb={5}>
-        <Box height={'300px'}>
-          <Image
-            boxSize="100%"
-            maxH="300px"
-            src={
-              data.chapter.banner_url === ''
-                ? undefined
-                : data.chapter.banner_url
-            }
-            alt=""
-            borderRadius="md"
-            objectFit="cover"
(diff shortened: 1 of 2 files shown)


ojeytonwilliams:
Sorry, @Sboonny, I realise I wasn't terribly clear about this.  If there's no image at all, we don't need to fallback to an empty box. I think, in that case, it would be better to show nothing at all:

image

What do you reckon?

Sboonny:
Sure don't see why not 👍

client/src/modules/chapters/pages/chapterPage.tsx
```diff
@@ -154,15 +155,21 @@ export const ChapterPage: NextPage = () => {
...
-          boxSize="100%"
-          maxH="300px"
-          src={data.chapter.banner_url ? data.chapter.banner_url : ''}
-          alt=""
-          borderRadius="md"
-          objectFit="cover"
-          fallbackSrc="https://cdn.freecodecamp.org/chapter/puppy-small.jpg"
-        />
+        {data.chapter.banner_url === '' ||
+        data.chapter.banner_url === null ? null : (
```
ojeytonwilliams, line 159:
```suggestion
        {data.chapter.banner_url && (
```
The truthiness of `banner_url` is the only thing that matters here, so we can simplify a bit.

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -201,16 +202,21 @@ export const EventPage: NextPage = () => {
...
+            maxH="300px"
+            src={data.event.image_url === '' ? undefined : data.event.image_url}
+            alt=""
+            borderRadius="md"
+            objectFit="cover"
+            fallbackSrc="https://cdn.freecodecamp.org/chapter/brown-curtain-small.jpg"
+            fallbackStrategy="onError"
+          />
+        </Box>
+      ) : null}
```
ojeytonwilliams, line 219:
```suggestion
      {data.event.image_url && (
        <Box height={'300px'}>
          <Image
            data-cy="event-image"
            boxSize="100%"
            maxH="300px"
            src={data.event.image_url}
            alt=""
            borderRadius="md"
            objectFit="cover"
            fallbackSrc="https://cdn.freecodecamp.org/chapter/brown-curtain-small.jpg"
            fallbackStrategy="onError"
          />
        </Box>
      )}
```
Same principle as above.

ojeytonwilliams (changes requested):

ojeytonwilliams (approved):
LGTM 👍

PR #1784, opened 2022-10-25
refactor: single user middleware

gikf:

---

- Pulls _user_ middlewares together into a single one.

### DIFF
--- a/server/src/controllers/Auth/middleware.ts
+++ b/server/src/controllers/Auth/middleware.ts
@@ -91,7 +91,7 @@ export const user = (req: Request, _res: Response, next: NextFunction) => {
         }),
         prisma.events.findMany({
           select: { id: true, chapter_id: true },
-          where: { event_users: { some: { user_id: id } } },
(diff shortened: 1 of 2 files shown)


server/src/controllers/Auth/middleware.ts
```diff
@@ -57,52 +57,6 @@ export type Venues = Merge<
   Prisma.venuesGetPayload<{ select: { id: true; chapter_id: true } }>[]
 >;
 
-// TODO: get events in user middleware and only get those that the user is an
-// event_user for
```
gikf, line 61: @ojeytonwilliams shouldn't this be events belonging to chapters, for which user is in chapter_users? :thinking:

ojeytonwilliams: Ooops, yes it should.  Technically we do check event roles, but since they don't have any permissions attached to them, they never authorize users!

ojeytonwilliams (approved):
LGTM :+1: 

This should definitely help if/when we get around to using passport.  Even if we don't, it's still cleaner.  Thanks!

PR #1805, opened 2022-10-29
fix: check instance role when chapter admin leaves chapter

gikf:

---

- Chapter admin might need to have instance role changed, when leaving admined chapter.
- I'm using `ChapterRoles.member` as an imitation of not being chapter admin anymore, for that chapter, in the check if instance role change is needed.


server/src/controllers/ChapterUser/resolver.ts
```diff
@@ -140,6 +173,19 @@ export class ChapterUserResolver {
...
+    const userChapters = ctx.user.user_chapters;
+    const chapterRole = userChapters.find(
+      ({ chapter_id }) => chapter_id === chapterId,
+    )?.chapter_role.name;
+    if (chapterRole === ChapterRoles.administrator) {
+      await updateInstanceRoleForChapterRoleChange({
+        changedChapterId: chapterId,
+        newChapterRole: ChapterRoles.member,
+        user: ctx.user,
+      });
```
ojeytonwilliams, line 186:
```suggestion
    await updateInstanceRoleForChapterRoleChange({
      changedChapterId: chapterId,
      newChapterRole: ChapterRoles.member,
      user: ctx.user,
    });
```
Do we need the conditional?  How about leaving that logic to `getInstanceRoleName`?

ojeytonwilliams (changes requested):
Just some nitpicks, really:

server/src/controllers/ChapterUser/resolver.ts
gikf: I'm seeing it as a kind of explanation why instance role might need changing. Otherwise I think some comment would be required to make that clear. What do you think?

ojeytonwilliams: Good question. A comment is probably the way to go, since it's not completely obvious why we're making this check.

Something like

> Certain chapter roles have instance roles associated with them, so we have to check and update accordingly

maybe?

ojeytonwilliams (approved):
LGTM :+1:

PR #1807, opened 2022-10-31
feat: validate event dates

### LINKED ISSUE #1781: Parsing of date entered directly to Start at or End at inputs
### Describe the bug:
`client/src/modules/dashboard/Events/components/EventForm.tsx`

When date is entered directly into the input field (ie. like it's done with cypress), the `DatePicker` tries to parse entered date from the first character. As `ends_at` and `start_at` fields are sometimes updated simultaneously - to not allow `start_at` to be after `ends_at`

To reproduce:
- Go to `/dashboard/events/new`
- Change _Start at_ date via input or `DatePicker` ie. to 15th of May 2022.
- Focus on the _End at_ input and start typing `2022-05-16 10:11`.
- After the first entered character (2), `DatePicker` will jump to February, and also _Start at_ date will be adjusted to not start after that date.

One option is using `strictParsing` in `DatePicker`. That prop will make parsing adhere strictly to the `dateFormat` prop, currently that's `MMMM d, yyyy h:mm aa`. Only after entering the full `May 16, 2022 10:11 AM` date would be parsed by `DatePicker`, so the _Start at_ date wouldn't be changed before that to some bogus date.

gikf:

Closes #1781

---

- Adds basic validation of the event dates relation - `start_at` < `ends_at`, in place of auto changing them.

### DIFF
--- a/client/src/modules/dashboard/Events/components/EventDatesForm.tsx
+++ b/client/src/modules/dashboard/Events/components/EventDatesForm.tsx
@@ -1,5 +1,5 @@
 import { FormControl } from '@chakra-ui/form-control';
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import DatePicker from 'react-datepicker';
 import { useFormContext } from 'react-hook-form';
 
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (approved):
LGTM :+1: Thanks!

PR #1808, opened 2022-10-31
refactor: separate init and seed

ojeytonwilliams:
- refactor: rename generator to seed
- refactor: separate init and seed

In development the database is created in three phases:

1. Create schema (db:sync)
2. Initialize the database by adding necessary records (roles and rsvps currently)
3. Seed the db with random data (chapters, events, user and so on)

When setting up a new production db, only steps 1. and 2. are necessary.

This PR is just to make it more obvious that init is distinct from seed.  This should make it clearer what can (exports from init) and cannot (exports from seed) be used in the rest of the server.


Sboonny (approved):

PR #1845, opened 2022-11-05
feat: automatic unsubscribe at cancel RSVP

### LINKED ISSUE #1844: Cancelling your RSVP to an event should remove your subscription automatically.
Luckily, gikf pointed out that my logic is flawed in https://github.com/freeCodeCamp/chapter/pull/1845#discussion_r1014630410, and I don't want to add Subscribing to RSVP. 

Because someone may want to RSVP but not subscribe, or in post MVP, subscribe to a repeating Event but not RSVP for the current Event, maybe wanting to attend the future Event in the next week.

So instead of RSVP removing subscription, should we create `cancel and unsubscribe` button, so people can do both and we don't have to worry about the logic

Sboonny:

Closes #1844


server/src/controllers/Events/resolver.ts
```diff
@@ -511,7 +511,7 @@ export class EventResolver {
     await updateWaitlistForUserRemoval({ event, userId: ctx.user.id });
 
     const updatedEventUser = await prisma.event_users.update({
-      data: { rsvp: { connect: { name: 'no' } } },
+      data: { rsvp: { connect: { name: 'no' } }, subscribed: false },
```
gikf, line 514: There might be `event_reminder` that needs to be deleted. `eventUser` from line `487` needs to include `event_reminder`, and then:

```suggestion
      data: {
        rsvp: { connect: { name: 'no' } },
        subscribed: false,
        ...(eventUser.event_reminder && { event_reminder: { delete: true } }),
      },
```

Sboonny: Ya, I thought It will magically stop the reminder, Thank you 👍

server/src/controllers/Events/resolver.ts
```diff
@@ -513,8 +513,10 @@ export class EventResolver {
     const updatedEventUser = await prisma.event_users.update({
       data: {
         rsvp: { connect: { name: 'no' } },
-        subscribed: false,
-        ...(eventUser.event_reminder && { event_reminder: { delete: true } }),
+        ...(eventUser.subscribed && {
+          event_reminder: { delete: true },
+          subscribed: false,
+        }),
```
gikf, line 519: This will error if there isn't reminder. Ie. when reminder was already send and deleted.

Sboonny: Ya, I thought the subscription is tied to eventReminder, but seeing the code I am wrong

```ts
  @Mutation(() => User)
  async toggleAutoSubscribe(@Ctx() ctx: Required<ResolverCtx>): Promise<User> {
    return await prisma.users.update({
      data: { auto_subscribe: !ctx.user.auto_subscribe },
      where: { id: ctx.user.id },
    });
  }
```

Sboonny: I am moving the discussion to the issue, because I am wondering what is the best course going forward

gikf: There are other emails dependent on the event user `subscribed` status. `subscribed` cannot be used in place of event reminder, as there wouldn't be way to ensure just single reminder is send, flipping `subscribed` after sending reminder would stop also other emails.

allella:
There are multiple _subscribed_ fields and it would seem toggling the _[event_user.subscribed](https://opensource.freecodecamp.org/chapter/tables/event_users.html)_ would be fine for a cancelled RSVP.

There was some debate about toggling _[chapter_users.subscribed](https://opensource.freecodecamp.org/chapter/tables/chapter_users.html)_ if someone turned off event notifications, but wanted to still be part of the _chapter_.

Does this help with the comments above?

ojeytonwilliams:
@Sboonny could you remove the extra commits? It looks like a merge went a bit wrong.

server/src/controllers/Events/resolver.ts
```diff
@@ -511,7 +511,13 @@ export class EventResolver {
     await updateWaitlistForUserRemoval({ event, userId: ctx.user.id });
 
     const updatedEventUser = await prisma.event_users.update({
-      data: { rsvp: { connect: { name: 'no' } } },
+      data: {
+        rsvp: { connect: { name: 'no' } },
+        ...(eventUser.subscribed && {
+          subscribed: false,
+        }),
```
ojeytonwilliams, line 518:
```suggestion
        subscribed: false,
```
This is basically the same since if `subscribed` is `false`, then setting to `false` again does nothing.

ojeytonwilliams (changes requested):

ojeytonwilliams (approved):
LGTM 👍

gikf:
I think there's still something not right with the changes to irrelevant files for the PR. As these changes are displayed in the _Files changed_ tab, merging the PR will apply them to main branch.

It appears https://github.com/freeCodeCamp/chapter/pull/1845/commits/c1500a7052dd37a275fe50c06a28499b2be080a1 has them all now. Removing that commit from PR or reverting it should solve that.

Sboonny:
I will create a new branch to be safe, and new PR

PR #1851, opened 2022-11-06
fix(ui): Inconsistent Header Height

### LINKED ISSUE #1653: Header inconsistent height for logged/not logged users
There's slight difference in the height of header, depending if user is logged in or not. It'd be good to have it the same for both states.

imageimage

jaitjacob:

Closes #1653

### DIFF
--- a/client/src/components/PageLayout/Header.tsx
+++ b/client/src/components/PageLayout/Header.tsx
@@ -30,6 +30,7 @@ const HeaderItem = forwardRef<HTMLDivElement, Props>((props, ref) => {
       {...props}
       w="full"
       as="header"
+      height={'90px'}
       px={[2, 4, 8]}
       py={[2, 4]}
       background={'gray.85'}
--- a/client/src/components/PageLayout/Header.tsx
+++ b/client/src/components/PageLayout/Header.tsx
@@ -30,7 +30,7 @@ const HeaderItem = forwardRef<HTMLDivElement, Props>((props, ref) => {
       {...props}
       w="full"
       as="header"
-      height={'90px'}
+      height={'4.5em'}
       px={[2, 4, 8]}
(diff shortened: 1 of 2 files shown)


Sboonny (changes requested):
The issue still presist

imageimage

Sboonny:
I am wondering is it better to change the avatar component width and height, instead of the Header in `/client/src/components/PageLayout/Header.tsx`?

jaitjacob:
> 

@Sboonny I did change the Avatar(Avatar.tsx) width and height **NOT** Header(Header.tsx). Not sure where did I go wrong.

Sboonny:
True you have, but the header's height still changes because `50px`, are bigger than the header height. You can use `36px` in the avatar, but it won't be presentable.

jaitjacob:
@Sboonny thanks for the clarification. I think this should pass.

client/src/components/PageLayout/Header.tsx
```diff
@@ -30,6 +30,7 @@ const HeaderItem = forwardRef<HTMLDivElement, Props>((props, ref) => {
       {...props}
       w="full"
       as="header"
+      height={'90px'}
```
Sboonny, line 33: The increased spaces seems off-putting to me in small screens. what about `4.5em`, so it looks like this 

image


```suggestion
      height={'4.5em'}
```

Sboonny (approved):
#### :sparkles: :tada: **AWESOME!** :tada: :sparkles:

 Hi @jaitjacob, thanks for this pull request and for contributing to the code-base for the first time. We are looking forward to more contributions from you in the future.

 Cheers & happy contributing!

Sboonny:
@allcontributors add @jaitjacob for code

PR #1855, opened 2022-11-06
feat: create  test and banned member commands

Sboonny:

While trying to test member access, I couldn't change to member with `npm run change-user`, because it create `undefined`.

so I created a command for member. To be honest, I have no idea if it's needed, maybe there is a command that I didn't notice


gikf:
The exact command that's used in the script can be used in the terminal: `npm run change-user member@chapter.com` or more general `npm run change-user <email>`. If there isn't user with that email, it will be created.

I'd say some other seeded users (`banned@chapter.admin`, `test@user.org`) could be added for convenience.

PR #1861, opened 2022-11-08
feat: add border to the menu

Sboonny:

Closes #XXXXX


Sboonny:
Which do you prefer?

---
border-bottom
image

---
border-top

image

client/src/components/PageLayout/Header.tsx
```diff
@@ -1,78 +1,43 @@
...
-    >
-      Log In
-    </MenuItem>
-  );
-};
 
 export const Header: React.FC = () => {
   const router = useRouter();
-  const { user, loadingUser } = useAuth();
+  const { user } = useAuth();
```
Sboonny, line 29: Why were we using `loadingUser`? I remember approving its PR, but I can't remember/find the PR

client/src/styles/Header.module.css
```diff
@@ -1,3 +0,0 @@
-.header {
-    gap: max(1em, 5%);
```
Sboonny, line 2: This was added for the searchBar, but because we aren't using the searchBar anymore, I deleted it

ojeytonwilliams:
@Sboonny those both look a little unbalanced.  How about equal spacing above Profile and below Dashboard?

client/src/components/PageLayout/Header.tsx
ojeytonwilliams: It's just there so we can show a spinner while we're fetching user data.

client/src/components/PageLayout/Header.tsx
```diff
@@ -1,78 +1,42 @@
...
-  const canAuthenticateWithGoogle = checkPermission(
-    user,
-    Permission.GoogleAuthenticate,
-  );
-
+  const login = useLogin();
   const goHome = () => router.push('/');
+  const CanEditDashboard = user?.instance_role.instance_role_permissions.filter(
+    ({ instance_permission }) =>
+      instance_permission.name === 'owner' || 'admin',
```
Sboonny, line 34:
```suggestion
      instance_permission.name === 'owner' || 'chapter_administrator',

File: client/src/components/PageLayout/Header.tsx (hunk 1 above)
Comment (Sboonny) [reply client/src/components/PageLayout/Header.tsx:28]: Adding it back 👍, should I add comment about it?

Comment (ojeytonwilliams) [reply client/src/components/PageLayout/Header.tsx:28]: Sure, if it's not clear from the context, then a comment's fine.

File: client/src/components/PageLayout/Header.tsx [hunk 4]
```diff
@@ -1,78 +1,42 @@
...
-    user,
-    Permission.GoogleAuthenticate,
-  );
-
+  const login = useLogin();
   const goHome = () => router.push('/');
+  const CanEditDashboard = user?.instance_role.instance_role_permissions.filter(
+    ({ instance_permission }) =>
+      instance_permission.name === 'owner' || 'chapter_administrator',
+  );
```
ojeytonwilliams, line 35: For this the logic is

```
if (user has permission to see any of the dashboards) {
show the dashboard link
} else {
don't
}
```

So, once we know which permissions matter we can just check if the user has at least one of those. Currently only `Permission.SponsorView` and `Permission.UsersView` are used and are checked in the Layout component.

ojeytonwilliams: Can we keep this PR focused on style changes and handle hiding the dashboard separately?

client/src/components/PageLayout/Header.tsx
```diff
@@ -84,27 +48,75 @@ export const Header: React.FC = () => {
...
+                    </Box>
+                  )}
+                  {user ? (
+                    <MenuItem
+                      data-cy="logout-button"
+                      onClick={() => logout().then(goHome)}
+                      fontWeight="600"
+                      height={'100%'}
+                    >
+                      logout
```
ojeytonwilliams, line 97:
```suggestion
                      Logout
```
To be consistent with the rest.

client/src/components/PageLayout/Header.tsx
```diff
@@ -84,27 +48,75 @@ export const Header: React.FC = () => {
...
+                    </MenuItem>
+                  ) : (
+                    <MenuItem
+                      data-cy="login-button"
+                      onClick={login}
+                      fontWeight="600"
+                      height={'100%'}
+                    >
+                      login
```
ojeytonwilliams, line 106:
```suggestion
                      Login
```

ojeytonwilliams (changes requested):

ojeytonwilliams (approved):
LGTM! Thanks, @Sboonny

PR #1867, opened 2022-11-09
fix: center chapterCards in chaptersPage section

Sboonny:

This is part of polish chapter PR, that should be in its own PR for easier review

image


ojeytonwilliams (approved):
Do think it might look better if the cards took up more of the screen width?  It already looks better, though.

PR #1872, opened 2022-11-10
feat: warn users that it's a testing site

ojeytonwilliams:

For now it's hard-coded, but when we launch we should only show it on staging.

### DIFF
--- a/client/src/modules/home/index.tsx
+++ b/client/src/modules/home/index.tsx
@@ -7,10 +7,6 @@ import {
   useToast,
   Flex,
   Text,
-  Alert,
-  AlertIcon,
-  AlertDescription,
-  AlertTitle,
 } from '@chakra-ui/react';
 import React, { useState } from 'react';
 import { Link } from 'chakra-next-link';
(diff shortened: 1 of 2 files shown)


Sboonny (commented):
What about moving the alert to the policy, or are you putting it in homepage to stop people from logging in unnecessary?

Because Policy is the only used for staging anyway

ojeytonwilliams:
The idea was just that if someone finds chapter, they'll likely find the homepage.

However, you got me thinking.  Why not put this on every page?  What do you reckon?  It's ugly, but it's only going to be visible while we're still testing.

client/src/modules/home/index.tsx
```diff
@@ -22,21 +26,32 @@ type User = NonNullable<AuthContextType['user']>;
...
+    <>
+      <Alert status="error">
+        <AlertIcon />
+        <AlertTitle> This is a testing site. </AlertTitle>
+        <AlertDescription>
+          Unless you are a freeCodeCamp staff member or a Chapter maintainer,
+          please do not use this site. Any data you enter will be deleted
+          periodically.
+        </AlertDescription>
+      </Alert>
```
Sboonny, line 38: Sure, I like the idea, but let's tune the tone down, because we don't want the maintainers to be on high alert because of the red color as Quincy said. Although I want to look up if that's true 🤔 

here is how it looks like in the suggestion

image


```suggestion
      <Alert status="info">
        <AlertIcon />
        <AlertDescription>
          This is a testing site for freeCodeCamp staff member or a Chapter
          maintainer, be mindful that your data will be deleted periodically.
        </AlertDescription>
      </Alert>
 ```

ojeytonwilliams: We do use red occasionally, though.  'danger zone' on /settings, for example.

I do prefer your description, with a minor tweak:

> This is a testing site for freeCodeCamp staff members and Chapter maintainers. Be mindful that your data will be deleted periodically.

Does that work for you?

Sboonny: yeab 👍

allella:
I believe we also added a noindex tag to the HTML head to block search engine from indexing the site. So, that should keep random people from stumbling upon a dev or test site.

Sboonny (approved):

ojeytonwilliams:
> I believe we also added a noindex tag to the HTML head to block search engine from indexing the site.

We did and that's still the plan for the staging version of the site.  However, chapter.freecodecamp.dev is down (DNS issues), so I'm going to have to move to .org.  Since we do want the .org version to be indexed by Google in the long run, the plan is to rely on this warning to dissuade random visitors.

Also, we won't be advertising it until it's a little more stable, so there probably won't be too many random visitors.

ojeytonwilliams:
Actually, it *was* down and has just come back up.  I still plan to move it to .org, though. In part because it's more stable, but mostly because we'll want to use the staging site purely as a last check before deploying properly.

PR #1873, opened 2022-11-10
refactor: improve consistency in variable naming

ojeytonwilliams:
The idea is that if 'id' is clear from context (i.e. the event query's
id is obviously event.id), then we should use that, but otherwise we
should be more explicit.

Closes #XXXXX


gikf (approved):
LGTM :+1:

Issue #1881, opened 2022-11-11
Event statuses in `/dashboard/events/` are not consistent with statuses on `EventCard`


gikf:
[EventCard](https://github.com/freeCodeCamp/chapter/blob/main/client/src/components/EventCard.tsx) is defining status of the event in slightly more precise manner than the [EventsPage](
https://github.com/freeCodeCamp/chapter/blob/main/client/src/modules/dashboard/Events/pages/EventsPage.tsx) in dashboard. Event dashboard should display the same statuses as the event card.

Sboonny:
I took care of this issue coincidently, because I wanted to test how easy is it to DRY the code with function in it. The answer was it wasn't easy to DRY it and the code was getting too smart, so I hard coded it 😅

PR #1882, opened 2022-11-11
feat: polish dashboard eventsPage's event status

Sboonny:

Update dashboard event status

Ref #1881


ojeytonwilliams (changes requested):
By writing `Ended at: Sat, Nov 12 @ 12:30`, say, we're implying that's when the event ended. However, we only display the start date.

To make it a little more obvious, we could do something like the following. 

- Display both start and end dates. i.e `Sat, Nov 12 @ 12:30 to Sat, Nov 12 @ 12:45` (nice to have: `Sat, Nov 12 @ 12:30 to 12:45` for single day events).
- Use `Ended`, not `Ended at`
- Put the status after the date `Sat, Nov 12 @ 12:30 to 12:45, cancelled`...
- ...or separately `Status: ended`.

What do you reckon?

ojeytonwilliams:
> Someone can always click on the event to see when it started and ended

Amazingly, they can't. The event page hasn't got this info!  I'll create an issue in a sec.

Back on topic: 

> but it's probably more important to show the end time than the start time.

Sure. The main thing is that it's obvious that it's over and when it ended.

allella:
The home page has the same "Ending at" and you can click on the past event.  I didn't log in to check the dashboard and assumed they were on the same template.

ojeytonwilliams:
@Sboonny don't worry about the pending checks. The next time you push a commit, they'll run.

ojeytonwilliams (approved):
We should find a nice way to display the end date, as Jim pointed out, but this is a step in the right direction. 

LGTM :+1:

ojeytonwilliams:
@Sboonny the event card and dashboard's don't match (dashboard was correct, imo, so I merged this), but I left https://github.com/freeCodeCamp/chapter/issues/1881 open as a result.

Sboonny:
have opened #1921 to make them match, you are talking about the texts `Ending at` and `Ending`, right?

ojeytonwilliams:
Yep, I think that's what #1881 was about.

PR #1889, opened 2022-11-13
fix: netcat on gitpod

### LINKED ISSUE #1880: nc command is missing on gitpod
### Describe the bug:
After #1865, installation of netcat can be seen in the terminal when gitpod is started. However after that it still appears to not be installed. With the `nc command is missing!` further down the start and in new, separate terminal:
```bash
gitpod /workspace/chapter (main) $ nc
bash: nc: command not found
```

gikf:

Closes #1880

---

- It's still not clear for me why the `init` method seemingly installed it, but then it wasn't available.

### DIFF
--- a/.gitpod.dockerfile
+++ b/.gitpod.dockerfile
@@ -1,0 +1,5 @@
+FROM gitpod/workspace-postgres
+
+RUN sudo apt-get update \
+ && sudo apt-get install netcat -y \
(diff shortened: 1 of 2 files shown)


ojeytonwilliams (approved):
I'm not sure, either, but this approach LGTM.

PR #1899, opened 2022-11-14
fix: redact serialized objects

ojeytonwilliams:

It turns out we can get these in our errors, too!


gikf:
A bit side question. Is it not needed to redact cookie (session, session.sig) in the errors?

gikf (approved):
Looks good :+1:

PR #1909, opened 2022-11-16
fix: migrate permission names

### LINKED ISSUE #1898: It should be possible to 'initialize' live databases
Whenever we update the [db init script ](https://github.com/freeCodeCamp/chapter/blob/2a9876947cac2de46d84488c7e869201475b7187/server/prisma/init/init.ts), we can create issues for existing databases.  For example, if we add a new permission, no existing database will have it and will behave incorrectly once the application code is updated.

The init script needs to be changed so that it can replace these tables (and only these tables).

ojeytonwilliams:
#1620 added and modified permissions. Since this has to be reflected in
live databases, it should be a migration.

It's a weird migration because it does not modify the schema, it
modifies data. Something we may want to avoid in future (if possible)

Closes #1898


ojeytonwilliams:
Completed in https://github.com/freeCodeCamp/chapter/pull/1918

PR #1915, opened 2022-11-17
fix: consistent layout for dashboard

### LINKED ISSUE #1886: Inconsistent positioning of Events dashboards
Main content of the events dashboard is positioned a bit differently than in the other dashboards.

This is especially visible when there's no entries:
positioning

Keeping those more consistent would be desired.

Sboonny:

Closes #1886

because the numbers are lying, and to make reviewing easier. 

there were instances which `datatable` had two div wrapped around it, I removed the 2nd div and add width 100%


ojeytonwilliams (approved):
LGTM :+1:

PR #1890, opened 2022-11-13
fix: display email address in user profile

SureshKuchana:
- [X] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md).
- [X] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [X] My pull request targets the `main` branch of Chapter.

Closes #XXXXX

PR #1916, opened 2022-11-17
feat: mention email in userProfile

### LINKED ISSUE #1853: Members should be able to see their own email address on their profile
Own email address should be displayed in the user profile.

Sboonny:

follow up to #1890

Closes #1853

### DIFF
--- a/client/src/modules/profiles/graphql/queries.ts
+++ b/client/src/modules/profiles/graphql/queries.ts
@@ -1,10 +1,13 @@
 import { gql } from '@apollo/client';
 
-export const meQuery = gql`
-  query me {
-    me {
+export const userProfileQuery = gql`
+  query userProfile {
+    userInformation {
       id
       name
+      email
+      auto_subscribe
+      image_url
       instance_role {
         instance_role_permissions {
           instance_permission {
@@ -12,12 +15,6 @@ export const meQuery = gql`
           }
         }
       }
-      admined_chapters {
-        id
-        name
-      }
-      auto_subscribe
-      image_url
     }
   }
 `;
--- a/client/src/modules/profiles/graphql/queries.ts
+++ b/client/src/modules/profiles/graphql/queries.ts
@@ -11,6 +11,10 @@ export const userProfileQuery = gql`
       instance_role {
         name
       }
+      admined_chapters {
+        id
+        name
+      }
     }
   }
 `;
--- a/client/src/modules/profiles/graphql/queries.ts
+++ b/client/src/modules/profiles/graphql/queries.ts
@@ -9,11 +9,7 @@ export const userProfileQuery = gql`
       auto_subscribe
       image_url
       instance_role {
-        instance_role_permissions {
-          instance_permission {
-            name
-          }
-        }
+        name
       }
     }
   }
--- a/client/src/modules/profiles/pages/userProfile.tsx
+++ b/client/src/modules/profiles/pages/userProfile.tsx
@@ -62,7 +62,7 @@ export const UserProfilePage = () => {
             Profile
           </Heading>
           <Heading as="h2" size={'lg'}>
-            Welcome {getNameText(data.userInformation?.name)}
+            Welcome, {getNameText(data.userInformation?.name)}
           </Heading>
           {data.userInformation?.admined_chapters.length > 0 && (
             <>
@@ -79,7 +79,7 @@ export const UserProfilePage = () => {
             </>
           )}
           <Heading as="h2" marginTop={'2em'} size="lg" fontWeight={500}>
-            This will change the setting in{' '}
+            Email address:{' '}
             <Text as="span" fontWeight={700}>
               {data.userInformation?.email}
             </Text>
--- a/client/src/modules/profiles/pages/userProfile.tsx
+++ b/client/src/modules/profiles/pages/userProfile.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
-import { Flex, Heading, Text, useToast } from '@chakra-ui/react';
+import { Heading, Text, useToast } from '@chakra-ui/react';
 import { useConfirmDelete } from 'chakra-confirm';
-import { Link } from 'chakra-next-link';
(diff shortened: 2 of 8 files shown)


Sboonny:
Creating the button will refactor the fields too much for small PR, this is enough to have the same functionality as previous code for now.

After this is merged, I will work on adding Download data button, that should be easier with different graphql fields

client/src/modules/profiles/graphql/queries.ts
```diff
@@ -0,0 +1,20 @@
...
+      email
+      auto_subscribe
+      image_url
+      instance_role {
+        name
+      }
+      admined_chapters {
+        id
+        name
+      }
```
Sboonny, line 17: for future, user_chapters have chapters in it, but the field doesn't allow for the client to request them, so the fields will be updated

server/src/graphql-types/User.ts
```diff
@@ -21,25 +21,7 @@ export class UserWithInstanceRole extends User {
...
-
-  @Field(() => [ChapterUser])
-  chapter_users: ChapterUser[];
-
-  @Field(() => [InstanceRole])
-  instance_role: InstanceRole[];
-
-  @Field(() => [EventUser])
-  event_users: EventUser[];
-}
```
Sboonny, line 39: wasn't used so deleted

server/src/graphql-types/User.ts
```diff
@@ -21,25 +21,7 @@ export class UserWithInstanceRole extends User {
...
-
-@ObjectType()
-export class UserInformation extends User {
+export class UserInformation extends UserWithInstanceRole {
```
Sboonny, line 24: this is needed, so userinformation query have admined_chapters

But it won't be needed after refactoring so it should be reverted in the future PR

Sboonny (commented):
Notes for the PR in the future

client/src/modules/profiles/pages/userProfile.tsx
```diff
@@ -56,32 +56,37 @@ export const UserProfilePage = () => {
...
                   <Link key={id} href={`/chapters/${id}`}>
                     {name}
                   </Link>
                 ))}
               </Flex>
             </>
           )}
-
+          <Heading as="h2" marginTop={'2em'} size="lg" fontWeight={500}>
+            This will change the setting in{' '}
```
ojeytonwilliams, line 82:
```suggestion
            Email address:{' '}
```
This seems clearer.  Ideally the styling should be uniform across Email address, Name and Profile picture, but that doesn't have to be this PR.

ojeytonwilliams (changes requested):

server/src/controllers/User/resolver.ts
```diff
@@ -55,16 +55,11 @@ export class UserWithInstanceRoleResolver {
...
-    if (!ctx.user) {
-      return null;
-    }
-    return await prisma.users.findUnique({
+  @Query(() => UserInformation)
+  async userInformation(@Ctx() ctx: ResolverCtx): Promise<UserInformation> {
+    return await prisma.users.findFirstOrThrow({
       where: {
-        id: ctx.user.id,
+        id: ctx.user?.id,
```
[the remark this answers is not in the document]
Sboonny: I think to please typescript, meQuery which create userInformation can be null, so userInformation can be undefined.

Will test it out and give updated feedback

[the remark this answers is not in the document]
Sboonny: Yeab to please TypeScript. The server run without issue, and  `npm run gen` is running without issues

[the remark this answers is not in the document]
ojeytonwilliams: Right, but you can return early if `ctx.user` is `null` and avoid querying unnecessarily.

server/src/controllers/User/resolver.ts
```diff
@@ -55,14 +55,9 @@ export class UserWithInstanceRoleResolver {
...
-  ): Promise<UserInformation | null> {
-    if (!ctx.user) {
-      return null;
-    }
-    return await prisma.users.findUnique({
+  @Query(() => UserInformation)
+  async userInformation(@Ctx() ctx: ResolverCtx): Promise<UserInformation> {
+    return await prisma.users.findFirstOrThrow({
       where: {
         id: ctx.user.id,
```
Sboonny, line 62:
```suggestion
        id: ctx.user?.id,

File: server/src/controllers/User/resolver.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (Sboonny) [reply server/src/controllers/User/resolver.ts:62]: But typescript will complain about the possibility of the object being null/undefined

and I would have to update the promise type `Promise<UserInformation | undefined>`, `Promise<UserInformation | null>`, which isn't true type of what userInformation should be.

I am adding `if (!ctx.user) return undefined` before 


```ts
    if (!ctx.user) return undefined
    return await prisma.users.findFirstOrThrow({
      where: {
        id: ctx.user?.id,
        ...
```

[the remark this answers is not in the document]
ojeytonwilliams: That's true.  We can go one of two ways: throw an error if `ctx.user` is missing (and handle the error response on the client) or return `null` if `ctx.user` is missing (...and handle _that_ on the client).  Either seems fine for now.

We shouldn't call the db when we know for sure that it hasn't got the data.

server/src/controllers/User/resolver.ts
```diff
@@ -57,49 +57,56 @@ export class UserWithInstanceRoleResolver {
...
+              user: true,
             },
-            user: true,
           },
         },
-      },
-    });
+      });
+    } catch (e) {
+      console.log('Unable to download Data');
```
Sboonny, line 107:
```suggestion
      console.log('Unable to fetch attendee data');
```

ojeytonwilliams (approved):
I reverted the query to return early, so that we don't make a query when we already know it will return null.

This doesn't change anything on the client side since any request can fail and the client is able to handle that.

Issue #1910, opened 2022-11-16
Define database through schema rather than data


ojeytonwilliams:
### Discuss your ideas or share your views: 
Since roles, permissions and rsvps are all represented by rows in the database, this creates some difficulties.  Whenever we modify the [init script](https://github.com/freeCodeCamp/chapter/blob/02a86eadb5dadca7c672e64fe2fb47ec63160779/server/prisma/init/init.ts) this, effectively, requires a migration.The reason for this is that any live databases will need these changes to function correctly. These changes also should only be applied once and a migration is the safest way to ensure that happens.

However, these are not normal migrations - they have no effect on the db schema.  This means, for one thing, that rolling back is not trivial since prisma cannot help you. You have to write additional queries to undo the migration. Also, each migration has to be created from scratch, rather than generated by prisma.  All of which can be done, but are more time consuming and error prone than schema changes.

### Why we're doing this

The plan is to allow _owner_ users to create custom roles specific to their instances. These roles need to be persisted and the database was the obvious place to do that.

### What we might do differently

One thought I had was to bake the permissions into the schema.  i.e. rather than each permission being a row, it could be a column.  Then a given role would still be still be data, but the permissions would not be. For example, instance roles could look like this

| name  |  chapter-join  | sponsor-manage  |   
|---|---|---|
|  chapter_administrator |  true  | false   |   
|  owner | true  | true    

This seems like it would at least reduce the complexity of the migration.

allella:
Would the custom queries / migration be limited to defining just the default / first owner? If so, is that causing complexity, or is the thought that other roles need to be hard coded?  I've been imagining the owner is like that in the Drupal CMS. "User 1" is the "owner" and all other users are added to the database with only user id 1 having a special meaning.

ojeytonwilliams:
> Would the custom queries / migration be limited to defining just the default / first owner? 

No, but that is something we need to address before too long. For now I've just been updating the db by hand after the owner signs up. Eventually we'll want some sort of initial signup process to create the first owner.

It's possible I'm overthinking this, but the problem is that the definition of a given instance role (owner, say) is stored as rows in three different tables (_instance_roles_, _instance_role_permissions_ and _instance_permissions_).  So, we end up having to write migrations like https://github.com/freeCodeCamp/chapter/pull/1909, by hand, to modify the permissions data.

[...]

PR #1918, opened 2022-11-17
chore: remove init script

### LINKED ISSUE #1898: It should be possible to 'initialize' live databases

ojeytonwilliams:
- feat: create migration to replace init script
- chore: delete init script

Closes https://github.com/freeCodeCamp/chapter/issues/1898

This is a partial solution to the issues raised in https://github.com/freeCodeCamp/chapter/issues/1910. It doesn't touch the schema, but it does make the migrations entirely responsible for setting up and migrating the database.  Prior to this PR this responsibility was shared with the init script.


ojeytonwilliams:
Re: deployment, this migration has already been applied to the live db, so all that needs to happen is for it to be marked as such.

gikf:
With these changes how can for example, another permission be added, to include in new migration?

ojeytonwilliams:
We'd need to write the SQL queries ourselves, unfortunately.  For the specific example of a new permission that was owner only, it would look like this:

```sql
INSERT INTO instance_permissions (name, created_at, updated_at)
VALUES ('new-permission', NOW(), NOW());

INSERT INTO instance_role_permissions (instance_role_id, instance_permission_id, created_at, updated_at)
VALUES 
    ((SELECT id FROM instance_roles WHERE name = 'owner'),
    (SELECT id FROM instance_permissions WHERE name = 'new-permission'),
    NOW(),
    NOW());
```

It's awkward and I'd love to know of a better way.  However, it feels like this is necessary since, effectively, permission data is part of the schema.

What do you think? Does this all make sense?

gikf (approved):
I really don't mind, just making sure.

ojeytonwilliams:
Okay, cool. Thanks for the review!

Let's see how this works in practice.

PR #1920, opened 2022-11-17
feat: log in when confirming RSVP

### LINKED ISSUE #1924: As a logged out user, you briefly see the 'join this event' modal before being redirected to Auth0

Sboonny:

Closes https://github.com/freeCodeCamp/chapter/issues/1924

The sequence of the logic was making people see a message for a second before login, here is video staging tester provided

https://user-images.githubusercontent.com/88248797/202520802-70b6b2b0-9b05-4afa-bf5c-f17f23a94766.mp4

### DIFF
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -107,6 +107,7 @@ export const EventPage: NextPage = () => {
     const ok = await confirm(confirmOptions);
 
     if (ok) {
+      if (!user) await login();
       try {
         await joinChapter({ variables: { chapterId } });
         await rsvpToEvent({
@@ -145,13 +146,11 @@ export const EventPage: NextPage = () => {
 
   // TODO: reimplment this the login modal with Auth0
   async function checkOnRsvp(options?: { invited?: boolean }) {
-    if (!user) await login();
     await onRsvp(options);
   }
 
   // TODO: reimplment this the login modal with Auth0
   async function checkOnCancelRsvp() {
-    if (!user) await login();
     await onCancelRsvp();
   }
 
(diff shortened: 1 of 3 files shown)


ojeytonwilliams:
@Sboonny unfortunately this doesn't work with Auth0, since it still tries to join the chapter while redirecting to the login page.  I think I know how to fix this, though (use a popup instead, basically).  I'll see what I can do.

ojeytonwilliams:
@Sboonny this has one unintended side-effect, it no-longer redirects new signups to the policy page.  However, what I think we should do (after we've spoken to Quincy about wording) is include any Chapter specific details in http://www.freecodecamp.org/news/privacy-policy so that new signups will get to see it without it interrupting RSVP flows and so on.

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -130,6 +131,7 @@ export const EventPage: NextPage = () => {
     });
 
     if (ok) {
+      if (!isLoggedIn) await login();
```
Sboonny, line 134: A nitpick, you can't see the cancel rsvp button and interact with it without logging in, so the check will always be `isLoggedIn`

Sboonny (commented):
Quincy mentioned to me, that the policy page can be simple about:

- we aren't selling your data
- you can delete your data

----

I say ignore the side effects, no one of testers mentioned policy page, and we can change how to display policy page instead of login to see it

ojeytonwilliams:
@gikf if you've got a sec, this could use a review and both Muhammed and I have worked on it!

gikf (commented):
Generally I think it's okay.

There's problem when while logged out, you try to rsvp to event to which you are already rsvped, but that isn't caused by this PR (it's present on `main` too). See #1948.

ojeytonwilliams:
Okay, cool. Since you've already raised that issue, I think this PR is done.

Thanks for the review!

PR #1956, opened 2022-11-24
fix: administrators abusing their permissions

ojeytonwilliams:
- refactor: rename tests
- fix: stop admin using permission in other chapters

The authorization checker was too willing to believe the incoming request.  Users could abuse the fact that they could pass in spurious `variableValues` with a request to fool the checker into using the wrong credentials.

For example, an administrator of chapter 1 sends an `updateEvent` request targeting event 2 in chapter 2. If they sent `{ eventId: 2, chapterId: 1 }` the checker would accept that this was a request to chapter 1, confirm that the administrator is allowed to update events in chapter 1 and authorize the request.

This PR prevents this by making the checker find all possible `chapterId`s.  If they are identical (and the user has the permission) the request is authorized, otherwise it is rejected.

The `{ eventId: 2, chapterId: 1 }` request has two `chapterId`s, 1 and 2, so it will be rejected.

### DIFF
--- a/common/authorization.ts
+++ b/common/authorization.ts
@@ -75,19 +75,27 @@ function getRelatedChapterId(
 ): number | null {
   const { chapterId, eventId, venueId } = variableValues;
 
-  if (chapterId) return chapterId;
-
-  if (eventId) {
-    const event = events.find(({ id }) => id === eventId);
-    if (event) return event.chapter_id;
+  function inferId(
+    id: number,
+    data: { id: number; chapter_id: number }[],
+  ): number | null {
+    const target = data.find((x) => x.id === id);
+    return target ? target.chapter_id : null;
   }
 
-  if (venueId) {
-    const venue = venues.find(({ id }) => id === venueId);
-    if (venue) return venue.chapter_id;
-  }
+  // We're using 'undefined' when the id should not be inferred, leaving 'null'
+  // to indicate that we've failed to infer the id (and so the request is
+  // invalid)
+  const inferredFromEvent = eventId ? inferId(eventId, events) : undefined;
+  const inferredFromVenue = venueId ? inferId(venueId, venues) : undefined;
 
-  return null;
+  const chapterIds = [inferredFromEvent, inferredFromVenue, chapterId].filter(
+    (x) => typeof x !== 'undefined',
+  );
+
+  return chapterIds.length && chapterIds.every((x) => x === chapterIds[0])
+    ? chapterIds[0]
+    : null;
 }
 
 function isAllowedByChapterRole(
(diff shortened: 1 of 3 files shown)


gikf (commented):
Those administrators! Interesting finding. Sounds like rejecting requests with mismatched `chapterId` is the least complicated way to deal with it.

This made me think, if it's possible to use similar approach to authorize instance permission, when having just chapter permission. In my try that's unfortunately possible. The test below is failing, as the response is `true`. Could you take a look if I'm not missing something? From the way it looks, it's kind like a standard chapter role permission check.

```typescript
    it('should reject request when instance permission is missing, but chapter permission is present', () => {
      expect.assertions(1);
      const user = userWithRoleForChaptersOneAndTwo;

      const resolverData = merge(resolverDataWithEventsAndVenues, {
        context: { user },
        info: { variableValues: { chapterId: 1 } },
      });

      expect(authorizationChecker(resolverData, ['some-permission'])).toBe(
        false,
      );
    });
```

ojeytonwilliams:
So, I could have sworn I saw the 'fake' variables getting passed yesterday, but not today. The GraphQL server filters out any `variableValues` that don't appear in the query/mutation arguments.  

Say you have
```gql
  mutation confirmRsvp($eventId: Int!, $userId: Int!) {
    confirmRsvp(eventId: $eventId, userId: $userId) {
```
and provide

```json
"variables": { "eventId": 1, "userId": 2, "chapterId": 3 }
```

`chapterId` is simply discarded by the server.  The `variableValues` passed to `authorizationChecker` are just ` { "eventId": 1, "userId": 2}`

What does this mean for us?  The attacks I was thinking of and the one you found (nice catch, by the way!) should fail because the request is not valid.

- If the attacker sends a variable that's not used, it's ignored.
- If the attacker uses a variable that's not in the schema (e.g. `mutation confirmRsvp($eventId: Int!, $userId: Int!, $chapterId: Int!)` that fails GQL validation.
- If the attacker omits a variable, that fails due to a `BAD_USER_INPUT` error.

So, every request must contain all the variables the query/mutation expects.  It can call them anything it likes, but they must correspond to arguments.

i.e. if you send the mutation

```gql
  mutation confirmRsvp($eventId2: Int!, $userId: Int!) {
    confirmRsvp(eventId: $eventId2, userId: $userId) {
```

`$eventId2` is perfectly valid,  It'll get rejected by `authorizationChecker`, because it's looking for `eventId`, but that's intended.

That means you can 

---

So, I think this PR is okay, because it requests with mismatching ids are invalid, so rejecting them here seems reasonable.  However, the vulnerabilities I thought I'd found don't seem to exist. Oops.

gikf (approved):
Thanks for the explanation and further exploring. I also think this is fine to add anyway. :+1:

ojeytonwilliams:
No problem, thanks for the review. Sorry for the false alarm.

Issue #1941, opened 2022-11-22
Chapter dashboard - links to events


gikf:
When in specific chapter dashboard (`/dashboard/chapter/{id}`), there's list with events from this chapter. Links to the events are leading to public event page (`/events/{id}`).

There's no link to the event dashboard page (`/dashboard/events/{id}`). This makes navigating to certain event, found on the chapter's event list, slightly inconvenient, as it's needed to go to events dashboard (`/dashboard/events`) and find wanted event on another list.

Simplest would be replacing link to lead to the event dashboard. Better might be including both links, with visual, or otherwise, indication which link is to event dashboard and which to event page.

dineshigdd:
Do you mean there should be a link from /events/{id} to /dashboard/event/{ id } because at the moment you need to go to the /dashboard/events to access /dashboard/events/{id}?

gikf:
Sorry, I mean the links to the related events at the `/dashboard/chapter/{id}`. That is the `EventList` component -  https://github.com/freeCodeCamp/chapter/blob/main/client/src/modules/dashboard/shared/components/EventList.tsx

dianachu18:
I'm working on this issue.

PR #1960, opened 2022-11-24
feat: filter event in chapterCard

### LINKED ISSUE #1719: Event status does not appear on Chapter cards

Sboonny:

Closes #1719

### DIFF
--- a/server/src/controllers/Chapter/resolver.ts
+++ b/server/src/controllers/Chapter/resolver.ts
@@ -36,6 +36,9 @@ export class ChapterResolver {
     return await prisma.chapters.findMany({
       include: {
         events: {
+          where: {
+            AND: [{ canceled: false }, { ends_at: { gt: new Date() } }],
+          },
           include: {
             venue: true,
           },
--- a/server/src/controllers/Chapter/resolver.ts
+++ b/server/src/controllers/Chapter/resolver.ts
@@ -39,6 +39,7 @@ export class ChapterResolver {
           where: {
             AND: [{ canceled: false }, { ends_at: { gt: new Date() } }],
           },
+          orderBy: { start_at: 'asc' },
           include: {
             venue: true,
           },


Sboonny:
This proves Oliver point, which he mentioned to me before. It's hard to associate chapter's event with chapter, so to review this you have to select 4 events for 4 different chapters and check if updating them changes in `/chapter` page, and `/home` page.

ojeytonwilliams:
>  It's hard to associate chapter's event with chapter, so to review this you have to select events for chapters and check if updating them changes in /chapter page, and /home page.

Is there a UX issue that needs resolving?

server/src/controllers/Chapter/resolver.ts
```diff
@@ -36,6 +36,9 @@ export class ChapterResolver {
     return await prisma.chapters.findMany({
       include: {
         events: {
+          where: {
+            AND: [{ canceled: false }, { ends_at: { gt: new Date() } }],
+          },
           include: {
```
ojeytonwilliams, line 42:
```suggestion
          orderBy: { start_at: 'asc' },
          include: {
```
Since we're here, might as well order them.

Sboonny:
There is a good first timer issue for it https://github.com/freeCodeCamp/chapter/issues/1941, and we have administrator roles for instance-owner, if they are overwhelmed and need help.

So I would say it should be resolved, but post MVP

ojeytonwilliams:
Sure, that makes sense.

ojeytonwilliams (approved):

PR #1966, opened 2022-11-26
feat: send email when event data changes

### LINKED ISSUE #1944: Updates to event times should email users

Sboonny:

Closes #1944
This isn't finished, I need to format the code to account for user time zone
image

<s>And I haven't tested "if the user is unsubscribed, will they receive an email"</s> Fixed


gikf:
Won't it be a bit weird when two emails are send when venue and date is changed at the same time?

Sboonny:
I am surprised I thought I would get more conflicts :thinking:

ojeytonwilliams (changes requested):
I've made a few specific suggestions below, but the fundamental issue here is that we don't need three `buildEmailFor` functions. One function that adds info about the new venue and/or info about the new dates would do the job better.

server/src/controllers/Events/resolver.ts
```diff
@@ -134,12 +156,67 @@ const updateReminders = (event: EventWithUsers, startAt: Date) => {
...
-  (isOnline(event.venue_type) && data.streaming_url !== event.streaming_url) ||
-  (isPhysical(event.venue_type) && data.venue_id !== event.venue_id);
+const hasVenueChanged = (data: EventInputs, event: EventWithUsers) => {
+  return (
+    data.venue_type !== event.venue_type ||
+    (isOnline(event.venue_type) &&
+      data.streaming_url !== event.streaming_url) ||
+    (isPhysical(event.venue_type) && data.venue_id !== event.venue_id)
+  );
+};
```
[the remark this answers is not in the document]
Sboonny: while working on the file, it was easier to read for me this way. I will revert it 👍

server/src/controllers/Events/resolver.ts
```diff
@@ -740,21 +785,12 @@ ${unsubscribeOptions}`,
...
-          yield { email, subject, text };
-        }
-      });
-    }
+    hasVenueChanged(data, event) ||
+      (hasDateChanged(data, event) &&
+        createEmailForSubscribers(
+          buildEmailForUpdatedEvent(data, event),
+          event,
+        ));
```
ojeytonwilliams, line 793: This is quite a weird pattern. Normally if you want to do something conditionally

```
if(cond) {
  // do stuff
}
// or
if (cond) // od stuff
```
is typical, but

```
cond && // do stuff
```

is not.

server/src/controllers/Events/resolver.ts
```diff
@@ -138,34 +160,57 @@ const hasVenueChanged = (data: EventInputs, event: EventWithUsers) =>
...
+  End: ${formatDate(data.ends_at)}<br />
+  ----------------------------<br />
+  <br />
+  `);
+  const body = `Updated venue details<br/>
+${venueContent?.join('')}<br />
+----------------------------<br />
+<br />
+${dateChangeContent}
+`;
```
ojeytonwilliams, line 213: This is quite a confusing sequence.  All that really needs to happen is something like this

```ts
  const streamingUrlUpdates = hasStreamingUrlChanged(data, event)
    ? createStreamingUrlUpdates()
    : '';
  const dateUpdates = hasDateChanged(data, event)
    ? createDateUpdates()
    : '';
// ...
  return `a template string with the various updates inserted`
 ```

For example: 

https://github.com/freeCodeCamp/freeCodeCamp/blob/main/tools/challenge-helper-scripts/helpers/get-step-template.ts#L62-L89

**Review by ojeytonwilliams — CHANGES_REQUESTED** [review]

File: server/src/controllers/Events/resolver.ts [hunk 4]
```diff
@@ -134,38 +156,65 @@ const updateReminders = (event: EventWithUsers, startAt: Date) => {
...
   data: EventInputs,
   event: EventWithUsers,
 ) => {
-  const subject = `Venue changed for event ${event.name}`;
-  let venueDetails = '';
+  const subject = `Details changed for event ${event.name}`;
 
-  if (isPhysical(event.venue_type)) {
+  // this returns [object Promise] and I don't know why
+  const createVenueLocationContent = async () => {
```
Sboonny, line 175: When testing this always result in [object Promise], I don't know why

gikf: Notice the function is async, so depending on the usage, it might need using `await` when it's called, to get the actual value.

server/src/controllers/Events/resolver.ts
```diff
@@ -134,38 +156,66 @@ const updateReminders = (event: EventWithUsers, startAt: Date) => {
...
-${venue.name} <br>
-${venue.street_address ? venue.street_address + '<br>' : ''}
-${venue.city} <br>
-${venue.region} <br>
-${venue.postal_code} <br>
+    return `The event is now being held at <br />
+    <br />
+- ${venue.name} <br />
+- ${venue.street_address && venue.street_address + '<br />'}
+- ${venue.city} <br />
```
ojeytonwilliams, line 185:
```suggestion
- ${venue.street_address ? venue.street_address + '<br />- ' : ''}
${venue.city} <br />
```
Otherwise you can end up with messages like

> \- Venue
> \- null - 123
> \- 123

server/src/controllers/Events/resolver.ts
```diff
@@ -134,38 +156,66 @@ const updateReminders = (event: EventWithUsers, startAt: Date) => {
...
 const buildEmailForUpdatedEvent = async (
   data: EventInputs,
   event: EventWithUsers,
 ) => {
-  const subject = `Venue changed for event ${event.name}`;
-  let venueDetails = '';
+  const subject = `Details changed for event ${event.name}`;
 
-  if (isPhysical(event.venue_type)) {
+  // this returns [object Promise] and I don't know why
```
ojeytonwilliams, line 174:
```suggestion
```

ojeytonwilliams (changes requested):
This is looking good.  I just spotted a minor issue with street_address while testing:

ojeytonwilliams (approved):
LGTM :+1:

Issue #1885, opened 2022-11-11
Google Calendar refresh token expiration


ojeytonwilliams:
# Issue to discuss

Refresh tokens can expire for [a variety of reasons](https://developers.google.com/identity/protocols/oauth2#expiration) and when they do the calendar integration will break (or will shortly afterwards when the access token also expires).  As such, we need a way to handle this.

### Inform the owner

Since their token has expired, they can re-authenticate with Google and it will give Chapter new tokens to work with. If we email the owner in the event we get a Invalid Credentials error, they can fix it.

To make sure we don't email them all the time, we should look for errors where the 'reason' is `authError` and the `message` is `Invalid Credentials`.  Also, we can set a flag in the db when that error occurs and remove the flag when they re-authenticate. That flag could ensure the owner only gets emailed when it breaks, not on every failed request.

### Have a backup

We currently have a single point of failure: the owner's account is the only one with access to the calendars and events.  If we allow more than one account access, then as long as at least one token is valid, the integration can continue.

My thinking is we could have use two accounts and rotate which credentials we use. That way neither account's refresh token should expire due to inactivity.

gikf:
There could be also displayed alert on page for instance owners, to not rely only on the email notification.

Do you know if there's any specific way to test if the integration (`refresh_token`) is still working/valid, which could be used periodically? Rather than handling it only after some action ends up with error. My searches for that are not yielding anything concrete.

ojeytonwilliams:
Sorry, @gikf I was sure I replied to this.

> Do you know if there's any specific way to test if the integration (refresh_token) is still working/valid, which could be used periodically?

We could do something like request a [calendar list](https://developers.google.com/calendar/api/v3/reference/calendarList/list).  It feels slightly sketchy as a health-check, but maybe it's okay. Any request that requires authentication should work, unless the user has revoked some scopes (if that's even possible).

> More drastic condition for displaying alert could be the lack of entries in the google_tokens table, including time before first authorization. If refresh_token is invalid we can safely delete the entry, it can't be used anymore anyway.
> 
> That might be though annoying, if somebody is not planning to use integration with calendar.

How about something like this. `if (tokensStored && tokensAreInvalid) showAlert();`. That way we avoid annoying people who never integrated in the first place.

ojeytonwilliams:
Thanks for the detailed breakdown, @gikf, it's much clearer now.

[...]

PR #1977, opened 2022-11-27
feat: calendar integration status

gikf:

Ref https://github.com/freeCodeCamp/chapter/issues/1885#issuecomment-1321724859

---

- Part of the ref.
- Adds util with calendar integration status.
- Adds `is_valid` field to `google_tokens` and adds migration for it.
- Basic reflection of the integration status in the `/dashboard/calendar`.
- Displays _token_ statuses on the `/dashboard/calendar`, using as a reference email (partially redacted) that was used to authenticate.
- Adds alert when calendar integration is broken.
- Stops trying to create calendar, when creating chapter, if there's no valid token.
- Token invalidation is not yet there. I'd rather do that in another PR, but that might be inconvenient.


Sboonny (commented):
PolicyPage has Auth with Google button, too. Although I don't think there is a need to add the logic, because people shouldn't be checking that page for auth button. So I am wondering should we:

- delete it from there? 
- add the `data?.calendarIntegrationStatus === null` to it, and DRY after this PR is merged?

gikf:
> PolicyPage has Auth with Google button, too. Although I don't think there is a need to add the logic, because people shouldn't be checking that page for auth button. So I am wondering should we:
> 
> * delete it from there?
> * add the `data?.calendarIntegrationStatus === null` to it, and DRY after this PR is merged?

I'd remove the button from policy, maybe leaving just link to the _Calendar dashboard_ in its place.

Sboonny (approved):

client/src/components/PageLayout/index.tsx
```diff
@@ -5,13 +5,31 @@ import {
   AlertDescription,
   AlertTitle,
 } from '@chakra-ui/react';
-import React from 'react';
+import React, { useEffect } from 'react';
```
ojeytonwilliams, line 8:
```suggestion
import React from 'react';
```
There's a way to do what we want without `useEffect`, using `skip` (I'll demonstrate below)

client/src/components/PageLayout/index.tsx
```diff
@@ -5,13 +5,31 @@ import {
...
 import { SkipNavContent } from '@chakra-ui/skip-nav';
 
+import Link from 'next/link';
+import { useCalendarIntegrationStatusLazyQuery } from '../../generated/graphql';
```
ojeytonwilliams, line 12:
```suggestion
import { useCalendarIntegrationStatusQuery } from '../../generated/graphql';
```

client/src/components/PageLayout/index.tsx
```diff
@@ -5,13 +5,31 @@ import {
...
+    user,
+    Permission.GoogleAuthenticate,
+  );
+  const [loadStatus, { data }] = useCalendarIntegrationStatusLazyQuery();
+
+  useEffect(() => {
+    if (canAuthenticateWithGoogle) {
+      loadStatus();
+    }
+  }, [canAuthenticateWithGoogle]);
```
ojeytonwilliams, line 31:
```suggestion
  const { data } = useCalendarIntegrationStatusQuery({
    skip: !canAuthenticateWithGoogle,
  });
```
This seems like the expected way to conditionally load.  Same basic idea as `useEffect`, but it's more concise and declarative.

client/src/modules/dashboard/Calendar/pages/Calendar.tsx
```diff
@@ -1,64 +1,95 @@
...
 import { useAuth } from '../../../auth/store';
 import { Layout } from '../../shared/components/Layout';
+import { DashboardLoading } from '../../../../modules/dashboard/shared/components/DashboardLoading';
 import { NextPageWithLayout } from '../../../../pages/_app';
 
 const serverUrl = process.env.NEXT_PUBLIC_SERVER_URL || 'http://localhost:5000';
 
 export const Calendar: NextPageWithLayout = () => {
-  const { user } = useAuth();
+  const { user, isLoggedIn } = useAuth();
```
ojeytonwilliams, line 20:
```suggestion
  const { user } = useAuth();
```
It doesn't look like we need this

client/src/modules/dashboard/Calendar/pages/Calendar.tsx
```diff
@@ -1,64 +1,95 @@
...
     user,
     Permission.GoogleAuthenticate,
   );
+  const { loading, error, data } = useCalendarIntegrationStatusQuery();
+  const {
+    loading: loadingStatuses,
+    error: errorStatuses,
+    data: dataStatuses,
+  } = useTokenStatusesQuery();
+  const isLoading = loading || loadingStatuses || !data || !isLoggedIn;
```
ojeytonwilliams, line 31:
```suggestion
  const isLoading = loading || loadingStatuses;
```
If they're not logged in, they just see the 401 page because of the dashboard `Layout` and we can handle the absence of data below.

client/src/modules/dashboard/Calendar/pages/Calendar.tsx
```diff
@@ -1,64 +1,95 @@
...
+  if (isLoading || error || errorStatuses)
+    return <DashboardLoading error={error || errorStatuses} />;
+
+  if (!canAuthenticateWithGoogle)
+    return <NextError statusCode={403} title="Access denied" />;
+
+  const isAuthenticated = data.calendarIntegrationStatus;
```
ojeytonwilliams, line 38:
```suggestion
  const isAuthenticated = data?.calendarIntegrationStatus;
```

server/src/graphql-types/TokenStatus.ts
```diff
@@ -0,0 +1,10 @@
+import { ObjectType, Field } from 'type-graphql';
+
+@ObjectType()
+export class TokenStatus {
+  @Field(() => String)
+  email: string;
```
ojeytonwilliams, line 6:
```suggestion
  redacted_email: string;
```
This might be overkill, but it makes it abundantly clear what this is.

server/src/controllers/Calendar/resolver.ts
```diff
@@ -0,0 +1,34 @@
...
+    const replacer = (
+      _match: string,
+      firstLetter: string,
+      middle: string,
+      lastLetterWithDomain: string,
+    ) => `${firstLetter}${middle.replace(/./g, '*')}${lastLetterWithDomain}`;
+    const statusesWithRedactedEmail = statuses.map((status) => {
+      const { email } = status;
+      const redactedEmail = email.replace(/^(.)(.*)(.@.*)$/, replacer);
+      return { ...status, email: redactedEmail };
```
ojeytonwilliams, line 30:
```suggestion
      const { email, ...rest } = status;
      const redactedEmail = email.replace(/^(.)(.*)(.@.*)$/, replacer);
      return { ...rest, redacted_email: redactedEmail };
```
If we make it super explicit, this logic would need to change.  Also a few other places would need email -> redacted_email, but TS will tell us where.

ojeytonwilliams (changes requested):
Nice work, @gikf. The UX is clear and I appreciate the thought that's gone into this.

I didn't find anything substantial when reviewing, just a few minor things.  I *think* `email` -> `redacted_email` is a bit better, but it's not a big deal if you'd rather keep it as is.

client/src/components/PageLayout/index.tsx
gikf: That's pretty handy!

server/src/graphql-types/TokenStatus.ts
gikf: I like explicit :rocket:

ojeytonwilliams (approved):
Great stuff, this all works beautifully :rocket:

PR #1979, opened 2022-11-28
fix: screen reader reading extra content in menuItem

Sboonny:

When I have read Bruce review to PR in freeCodeCamp, I thought of our menu component and how it's similar.

Here is the review https://github.com/freeCodeCamp/freeCodeCamp/pull/48211#issuecomment-1320484412

### DIFF
--- a/client/src/components/PageLayout/Header.tsx
+++ b/client/src/components/PageLayout/Header.tsx
@@ -74,7 +74,6 @@ export const Header: React.FC = () => {
                 fontWeight="600"
                 borderRadius="5px"
               >
-                {/* We are using Nextlink because of confustion in NextJs which harm the loading functionality and force the page to wait for the JS loading to display it */}
                 <NextLink passHref href="/chapters">
                   <MenuItem as="a">Chapters</MenuItem>
                 </NextLink>
--- a/client/src/components/PageLayout/Header.tsx
+++ b/client/src/components/PageLayout/Header.tsx
@@ -4,7 +4,6 @@ import {
   Box,
   Image,
   Button,
-  Flex,
   Menu,
   MenuList,
   MenuItem,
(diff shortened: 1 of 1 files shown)


client/src/components/PageLayout/Header.tsx
```diff
@@ -69,32 +68,42 @@ export const Header: React.FC = () => {
...
                     borderRadius="5px"
                   >
-                    <NextLink passHref href="/chapters">
-                      <MenuItem as="a">Chapters</MenuItem>
-                    </NextLink>
+                    <MenuItem as="li" cursor="pointer">
+                      <Link width="100%" href="/chapters">
+                        Chapters
+                      </Link>
+                    </MenuItem>
```
Sboonny, line 81: I have wrapped the link around the text, because I think this way the screen readers read it as `link to chapter`.

While previously, they would have read `link link to chapter`.

client/src/components/PageLayout/Header.tsx
```diff
@@ -69,32 +68,42 @@ export const Header: React.FC = () => {
...
+                          <MenuItem
+                            data-cy="menu-dashboard-link"
+                            as="li"
+                            cursor="pointer"
+                          >
+                            <Link width="100%" href="/dashboard/chapters">
                               Dashboard
-                            </MenuItem>
-                          </NextLink>
+                            </Link>
```
Sboonny, line 105: I have no idea why we used `NextLink`, I think it was because of error message in the concole but I can't find the PR, or replicate the issue. I should learn more about git

client/src/components/PageLayout/Header.tsx
```diff
@@ -123,9 +130,9 @@ export const Header: React.FC = () => {
             )}
           </Box>
           {user && (
-            <NextLink passHref href="/profile">
+            <Link width="100%" href="/profile">
               <Avatar user={user} cursor="pointer" />
-            </NextLink>
+            </Link>
```
Sboonny, line 135: I am worried how screen reader read this, but I don't have one and reading about it in text doesn't give a good concept of how they actually read it. 😞

Sboonny (commented):
Here is notes of why I did the changes

client/src/components/PageLayout/Header.tsx
Sboonny: with windows narrator, it says "link the owner". but this isn't link to the owner it's link to profile, we need to change the context of the link to mention so, something like "link to your profile"

Sboonny:
fixed the profile screen reader text in https://github.com/freeCodeCamp/chapter/pull/1972, and reverted my old changes because I should use screen reader `narrator` instead of hunch, and the screen reader experience was fine 👍

client/src/components/PageLayout/Header.tsx
Sboonny: fixed

client/src/components/PageLayout/Header.tsx
```diff
@@ -53,75 +52,74 @@ export const Header: React.FC = () => {
...
+                Menu
+              </MenuButton>
+              <MenuList
+                paddingBlock={0}
+                display="flex"
+                flexDirection="column"
+                fontWeight="600"
+                borderRadius="5px"
+              >
+                {/* We are using Nextlink because of confustion in NextJs which harm the loading functionality and force the page to wait for the JS loading to display it */}
```
ojeytonwilliams, line 77: I don't really understand this, could you explain?  The purpose of next/link is to handle within app page transitions, which it can only do once the js has been evaluated (but it's still just an anchor element until then).

ojeytonwilliams (commented):
From a a11y perspective, should we be avoiding menu entirely?  Is something like https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-navigation.html more appropriate?

Sboonny:
> From a a11y perspective, should we be avoiding menu entirely? Is something like https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-navigation.html more appropriate?

Looks like it

client/src/components/PageLayout/Header.tsx
ojeytonwilliams: Since it's talking about behaviour we're not running into now, yeah, let's get rid of it.

ojeytonwilliams (approved):
LGTM :+1:

PR #1990, opened 2022-11-29
fix: rsvp to rsvped event

### LINKED ISSUE #1948: Rsvp to rsvped event when logged out
Error occurs when user, while logged out, tries to rsvp to event to which is already rsvped.

To reproduce:
- Login
- Rsvp to event
- Logout
- Go to the same event page, and press _RSVP_
- Confirm wanting to Rsvp
- (logging in)
- Error

gikf:

Closes #1948 

---

- Would you believe me, if I told you how much troubles I had with this?
- Fixes error when logged out user tries to rsvp to already rsvped event, and gets error when logging in.
- Adds modal _Waiting for login_. I found a bit confusing when _Confirm_ modal disappeared, while waiting for login.
- Tested both with dev login and test auth0 login.


Sboonny (approved):
LGTM 🎉, thank you for your contribution to the page! 👍. Awesome work!

ojeytonwilliams:
> Would you believe me, if I told you how much troubles I had with this?

Probably!  What issues did you run into?  If there's something systematically wrong with what we're doing, maybe I can help a bit.

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -147,9 +134,43 @@ export const EventPage: NextPage = () => {
...
+    }
   }
 
+  useEffect(() => {
+    if (rsvpAfterLogin && isLoggedIn) {
+      setRsvpAfterLogin(false);
+      modalProps.onClose();
+      onRsvp();
+    }
+  }, [rsvpAfterLogin, isLoggedIn]);
```
ojeytonwilliams, line 172: I'm guessing this is where the pain came from (or rather that this was how you solved it).

It got me thinking, we could have `login` return the updated meQuery result.  That's probably a good idea in general and would mean you could use the `user` data to get `rsvpStatus` rather than having to wait for the page reload.  How does that sound?

gikf:
> Probably! What issues did you run into? If there's something systematically wrong with what we're doing, maybe I can help a bit.

The main issue I had was making `onRsvp` to use the `rsvpStatus` updated after the login.

Seemingly every way `checkOnRsvp` could call `onRsvp`, the `onRsvp` would have `rsvpStatus` before login - undefined. This includes shenanigans with `useRef`, `useState`, trying to pass updated `rsvpStatus` to `checkOnRsvp` first, or using `setInterval` in various places to literally check every second, if `rsvpStatus` has changed. :eyes:

client/src/modules/events/pages/eventPage.tsx
gikf: It might be useful in general to have that option. Not sure though how convenient it would be to extract such details from it. In this case `rsvp` would need to be included in the meQuery as well.

ojeytonwilliams: I don't believe so.  We're after `rsvpStatus` which is obtained from `user` (which we don't have yet, but will be able to soon https://github.com/freeCodeCamp/chapter/pull/2009) and from `data` out of `useEventQuery` which is not refetched.

gikf: Ohh, I totally didn't think about it from this side.

ojeytonwilliams (approved):
Nice work, @gikf this LGTM :+1:

Issue #1975, opened 2022-11-27
Accessibility concern in dashboard component


Sboonny:
### Discuss your ideas or share your views: 
In dashboard component we have `actions` that act as link (interactive component), here is an example

https://github.com/freeCodeCamp/chapter/blob/7a21f751446a1a4c715edb4c7f9824a9744e3bd3/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx#L14-L36?plain=1

```ts
const actionLinks = [
  {
    colorScheme: 'blue',
    size: 'xs',
    href: (id: number) => `/dashboard/chapters/${id}/new-venue`,
    text: 'Add venue',
    requiredPermission: Permission.VenueCreate,
  },
  {
    colorScheme: 'blue',
    size: 'xs',
    href: (id: number) => `/dashboard/chapters/${id}/new-event`,
    text: 'Add Event',
    requiredPermission: Permission.EventCreate,
  },
  {
    colorScheme: 'blue',
    size: 'xs',
    href: (id: number) => `/dashboard/chapters/${id}/edit`,
    text: 'Edit',
    requiredPermission: Permission.EventEdit,
  },
];
```

Although this is fine, I am worried about the repetition of the actions, which lead to this accessibility concern.

So here is what I am worried about, we have multiple `edit` button or alike, that do the same thing `edit chapter` but for different chapters, so people may get confused of the lack of info in these buttons and have a worse experience.

I am wondering if we should add `aira-label` to these buttons https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label, so people with accessibility know that they are editing a specific chapter, and to give access to voice users playbook https://blog.google/technology/next-billion-users/voice-users-playbook/, because they need to ask the device to interact with a component, and giving a unique name should help them.

ojeytonwilliams:
It seems to me that repeating which chapter you're in three times would be more frustrating than helpful.

Instead, why don't we make it clearer that you're in the action row for a given chapter? i.e. when a user tabs to that row it could announce "Actions for ${chapter.name}" and then should be clear that Add venue, Add event and Edit all refer to that.

Sboonny:
Ya, I like it. adding `tabindex=0` and `aria-label` should do the trick. 👍 Good call.

Sboonny:
wait what about people who use voice playback, using this won't account for them

ojeytonwilliams:
[...]

PR #1991, opened 2022-11-29
feat: add screen reader text in chapterDashboardPage

Sboonny:

Ref #1975

Adding `aria-label` doesn't add text in `narrator`, so user who use their voice to navigate may have an issue, but I hope `sr-only` be good enough.

### DIFF
--- a/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
@@ -95,7 +95,11 @@ export const ChaptersPage: NextPageWithLayout = () => {
                       href={href(chapter.id)}
                     >
                       {text}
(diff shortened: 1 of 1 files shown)


ojeytonwilliams:
> Adding aria-label doesn't add text in narrator, 

Is the narrator something different from the screen reader?  Orca reads `aria-label`s just fine, so I'm a bit confused.

Sboonny:
Ya it doesn't pick up `aria-label`, I need to check the component `dataTable` itself and see if there are special cases. this is a small update for now, because I want to finish email content and user download button, then update the dataTable with the requested menu from the testers

ojeytonwilliams (approved):
Thanks for walking me through this. I tried it out with Orca and it all seems to be working as intended.

LGTM :+1:

PR #2000, opened 2022-11-30
fix: the footer's syntax in page layout

Sboonny:

I have no idea, where to put it. I have thought of placing it in the menu, but I come to hate because policy isn't something that will be viewed daily by users (I do see the irony, by me putting it in footer 🤣). I guess I can create Card in HomePage and place the context there. What do you think of this?


Sboonny:
I have moved it to the alert, it feels random tho

gikf:
Could you write why is it being removed?

client/src/components/PageLayout/component/Footer.tsx
```diff
@@ -1,69 +1,33 @@
...
-      </UnorderedList>
-      {/* The layout need to have div wrapped around it to give desired looks
-          This isn't normal, ToDo: find out why the need for div,
-          and remove it for cleaner code
-      */}
-      <Box
-        display="inline-grid"
-        placeItems={{ base: 'flex-start', lg: 'center' }}
-        width="100%"
+      <Link href="/policy" fontWeight="600">
```
Sboonny, line 15:
```suggestion
      <Link href="/chapters" fontWeight="600">
```

client/src/components/PageLayout/component/Footer.tsx
```diff
@@ -1,69 +1,33 @@
...
+        Chapters
+      </Link>
+      <Link href="/policy" fontWeight="600">
```
Sboonny, line 18:
```suggestion
      <Link href="/events" fontWeight="600">
```

ojeytonwilliams (commented):
Is it possible to keep the footer at the bottom of the page when there's not much content?

It looks a little odd when the page is basically empty:
Floating footer

If this is tricky to implement, don't worry, it's a fairly minor issue!

Sboonny:
Not really hard to implement, I need to add height property to the Box, and if it doesn't work it should be easy to figure out a hack

PR #2015, opened 2022-12-02
fix: stop new rsvps cancelling attendance

ojeytonwilliams:

The issue was that we sent requests to the Google api with `attendees: [ { email: 'an@add.ress'} ]` rather than `attendees: [ { email: 'an@add.ress'}, responseStatus: ... ]`, that meant Google would remove their previous `responseStatus`.

The fix was to first get the event data, modify it and then send the updated data.

Tests to follow in another PR.


server/src/services/Google.ts
```diff
@@ -189,98 +185,166 @@ function getStandardPatchBody({
...
 
-  await calendarApi.events.patch({
-    calendarId,
-    eventId: calendarEventId,
-    sendUpdates: 'all',
-    requestBody: getStandardPatchBody(eventUpdateData),
-  });
+function filterFromAttendees(email: string) {
+  return (attendees: calendar_v3.Schema$EventAttendee[]) =>
+    attendees.filter((attendee) => attendee.email === email);
```
gikf, line 285: Are we leaving here other attendees, or just the one having `email`?
```suggestion
    attendees.filter((attendee) => attendee.email !== email);
```

gikf (commented):
I like it very much in general.

server/src/services/Google.ts
ojeytonwilliams: Oops. It's what you suggested.

Looks like I'm writing those tests now!

ojeytonwilliams:
Phew! All done.

A few of the tests are a little trivial (does `deleteCalendarEvent` call `events.delete`? Yes. Yes it does), but overall I think they're reasonable.

I've not tested them locally, yet, I ran out of time. That's next on my list, but I figured it was worth sharing the code anyway.

server/tests/google.test.ts
```diff
@@ -0,0 +1,286 @@
...
+        { attendeeEmail: 'b@person' },
+      );
+
+      expect(updatedEvent.data.attendees?.length).toBe(2);
+      expect(updatedEvent.data.attendees).toEqual(
+        arrayContaining(remainingAttendees),
+      );
+    });
+
+    it("should do nothing if the email one of the attendee's", async () => {
```
gikf, line 116:
```suggestion
    it("should do nothing if the email is not one of the attendee's", async () => {
```

server/src/util/calendar.ts
```diff
@@ -2,64 +2,34 @@ import { inspect } from 'util';
...
-
 interface CreateCalendarEventData {
   attendeeEmails: string[];
   calendarId: string;
   event: Pick<Event, 'id' | 'ends_at' | 'start_at' | 'name'>;
 }
 
-export const createCalendarEvent = async ({
+// TODO: consider pulling this back into the resolver (it's only used twice,
+// it's not that complex so I don't think we need to abstract it out)q
```
Sboonny, line 15:
```suggestion
// it's not that complex so I don't think we need to abstract it out)
```

Sboonny (approved):
I think I couldn't test it, because `responseStatus` did fail for me, but I have gone to the calendar and activated google auth. 🤔

Well, doesn't matter anymore.  👍 Thank you @gikf, for reviewing this. 💚

ojeytonwilliams:
@Sboonny so we don't accidentally spam emails during Cypress testing, the attendees are set to `[]` when you're running the site in development mode.  Could it be that?

gikf (approved):
@ojeytonwilliams is there anything else you want to do here?

Sboonny:
Oh I think, I am the one causing the delay, because the talk about not being able to test it

ojeytonwilliams:
> is there anything else you want to do here?

Nope, I think it's fine.  I'll deploy it now and see if anything breaks.

PR #2023, opened 2022-12-03
feat: turn the user image in nav into a menu

### LINKED ISSUE #2042: Menu button text does not match accessible name
### Describe the bug:

The menu button in the upper right corner has the visible text "Menu" but the accessible name on the button is set to "Options" using `aria-label`. This is going to make it hard for people using voice control software to use the menu since they won't be able to see the actual accessible name. It will also be confusing for screen reader users who can see the text on the page (not all screen reader users are blind).

### Solution

Decide whether this button should be called "Menu" or "Options" and then use that as the button text and get rid of the `aria-label` completely as it is not needed.

Sboonny:

This is the last feedback from the tester, that is possible for me to do. I am going back to doing what I was previous doing, which I forgot what it was 🤣.

Btw did someone faced a situation in which graphql just said "nope"?

Closes #2042

image


client/src/components/PageLayout/Header.tsx
```diff
@@ -42,18 +51,9 @@ export const Header: React.FC = () => {
...
-          href="/"
-          _focus={{
-            outlineColor: 'blue.600',
-            outlineOffset: '5px',
-          }}
-          _focusVisible={{
-            boxShadow: 'none',
-          }}
-        >
+        <Link href="/">
```
Sboonny, line 56:
```suggestion
        <Link
          href="/"
          _focus={{
            outlineColor: 'blue.600',
            outlineOffset: '5px',
          }}
          _focusVisible={{
            boxShadow: 'none',
          }}
        >

**Comment by Sboonny** [comment]
Btw, this https://github.com/freeCodeCamp/chapter/pull/2023/commits/7b1f7350d0c9f3d05f0ef5a66695245373e1d32c didn't work because although chakra allow you to add `aira-control` it doesn't respect it and overwrite it, which lead to the component doesn't have the aria-control that I wanted

**Review by gikf — COMMENTED** [review]
I don't have much opinion on moving menu. What stands out for me, is having two Login/Logout buttons so close to each other.

**Review by ojeytonwilliams — CHANGES_REQUESTED** [review]
> This is the last feedback from the tester, that is possible for me to do

Which feedback was this?  I remember Estefania (I think) mentioning something about this, but I can't find anything in my notes :/

Anyways, to this PR. How about having logout in the menu, but login separate (as you have it right now)?  I think that's a fairly standard pattern, partly to encourage people to login and to make it less likely they accidentally sign out.

Also the ArrowUpDownIcon seems a little strange because there's no up. It's really a dropdown, so I think the icon should indicate that.

**Comment by Sboonny** [comment]
> Which feedback was this? I remember Estefania (I think) mentioning something about this, but I can't find anything in my notes :/

I forgot who requested it 🤣, thank you for the reminder. It's in my DMs about.

She sent me, these 

image

image

File: client/src/components/PageLayout/Header.tsx [hunk 2]
```diff
@@ -16,16 +15,26 @@ import NextLink from 'next/link';
 import { useRouter } from 'next/router';
 import React from 'react';
 
+import { ArrowDownIcon } from '@chakra-ui/icons';
```
ojeytonwilliams, line 18:
```suggestion
import { ChevronDownIcon } from '@chakra-ui/icons';
```
The chevron is a standard way of representing a collapsible menu, let's stick to UI elements people are familiar with.

client/src/components/PageLayout/Header.tsx
```diff
@@ -64,85 +73,86 @@ export const Header: React.FC = () => {
...
+                  as={Button}
+                  data-cy="menu-button"
+                  padding="0"
+                  width="4.5em"
+                  {...(user ? menuButtonStyles.login : menuButtonStyles.logout)}
+                >
+                  {user ? (
+                    <HStack spacing="0">
+                      <Avatar user={user} cursor="pointer" aria-label="menu" />
+                      <ArrowDownIcon
```
ojeytonwilliams, line 99:
```suggestion
                      <ChevronDownIcon
```
Ideally this should flip from down to up as you go from collapsed to expanded, but it's fine to be static for this PR.

ojeytonwilliams (changes requested):

ojeytonwilliams (approved):
Cheers, @Sboonny, thanks for the a11y and UX improvements.

PR #2035, opened 2022-12-05
feat: make stream_url input lowerCase

Sboonny:

While testing #1966, it was possible to send email if someone change the capitalization on stream url. There were no emails sent to MailHog on my end, so hopefully this fixes it


ojeytonwilliams (changes requested):
URLs are, in general, case sensitive so we shouldn't mess with them.

Sboonny:
There are,  I thought there shouldn't be emails

ojeytonwilliams:
Cool, thanks for confirming.

Issue #2044, opened 2022-12-05
Color contrast on focus indicator is not high enough


bbsmooth:
### Describe the bug:
The light blue focus indicator does not have enough color contrast with the white background to be accessible. Using the color picker tool in gimp, the light blue color used for the focus indicator is approximately `#8ec2ed`. This only has a contrast ratio of `1.89:1` with the white background. It needs to be at least `3:1`.

Additionally, when it is around a blue button the contrast between the dark blue button and the focus indicator is only `2.12:1`. And with the red buttons it is only slightly better at `2.17:1`. 

### Solution

The focus ring definitely needs to have a contrast of at least `3:1` with the white background. As for contrast with the darker buttons, one option would be to place a little distance between the button and the focus indicator so that the white background separates them (this can be as little as 1px). As long as you have that separation then you don't need to worry about the contrast between the focus indicator and the button color since they aren't touching. For example, if you were using `outline` for the focus indicator then you could use `outline-offset: 1px` to create that separation.

Sboonny:
Went through the pages, didn't find a button that need outline so I am closing this

PR #2050, opened 2022-12-06
fix: outline for blue buttons

Sboonny:

Ref #2044

Changed the button color, to our primary one and add spacing for outline, as recommended. 

This doesn't add outline to all primary color buttons, the blue ones only.

there are two blue buttons left, they require refactoring which can be left to last.

### DIFF
--- a/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
@@ -13,21 +13,42 @@ import { useAuth } from 'modules/auth/store';
 
 const actionLinks = [
   {
-    colorScheme: 'blue',
+    background: 'gray.85',
+    color: 'gray.10',
+    _hover: { color: 'gray.85', backgroundColor: 'gray.10' },
+    _focusVisible: {
+      outlineColor: 'blue.600',
+      outlineOffset: '1px',
+      boxShadow: 'none',
+    },
     size: 'xs',
     href: (id: number) => `/dashboard/chapters/${id}/new-venue`,
     text: 'Add venue',
     requiredPermission: Permission.VenueCreate,
   },
   {
-    colorScheme: 'blue',
+    background: 'gray.85',
+    color: 'gray.10',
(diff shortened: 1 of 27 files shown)


Sboonny:
Every change, I have made the past two days will create conflict with this, so we can add conflict tag until everything is merged

ojeytonwilliams:
Changing the outline really should not mean we have to change 25 files.

Chakra lets you define [colour schemes and variants](https://chakra-ui.com/docs/components/button/theming) of every component. So, rather than finding every blue button and altering its outline, why not create a default variant that has the outline you want?  Then you can just use those components and only need to configure them if there's a special case.

ojeytonwilliams:
Sure, I take the point about the number of changes.  It's just that once these components are set up it's really easy to keep things consistent and future PRs really will just change a few lines of code.

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -90,7 +90,7 @@ export const ChaptersPage: NextPageWithLayout = () => {
                         chapterId: chapter.id,
                       }),
                   )
-                  .map(({ colorScheme, size, text, href }) => (
+                  .map(({ size, text, href, colorScheme }) => (
```
ojeytonwilliams, line 93:
```suggestion
                  .map(({ colorScheme, size, text, href }) => (
```

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -156,7 +156,7 @@ export const ChaptersPage: NextPageWithLayout = () => {
                               chapterId: chapter.id,
                             }),
                         )
-                        .map(({ colorScheme, size, href, text }) => (
+                        .map(({ size, text, href, colorScheme }) => (
```
ojeytonwilliams, line 159:
```suggestion
                        .map(({ colorScheme, size, href, text }) => (
```

ojeytonwilliams (changes requested):
Just to cut down on the diff noise:

ojeytonwilliams (approved):
LGTM :+1:

PR #2057, opened 2022-12-07
fix: heading structure on events page

### LINKED ISSUE #2046: Improper heading structure on Events page
### Describe the bug:
The Events page has `h3` headings nested within `h4` headings which is a WCAG failure of [SC 1.3.1](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html).

### Solution
I'm guessing that the `h4`s are supposed to be `div`s and this is just a mistake. 

Also, in a perfect world the `h3`s would be `h2`s so that heading levels aren't skipped.

Sboonny:

Closes #2046

I can't use h2, because the component is used in homepage, maybe I should use params but, I wonder should we count cards as their own section or part of the Events section and keep them without heading like chapterCard

### DIFF
--- a/client/src/components/EventCard.tsx
+++ b/client/src/components/EventCard.tsx
@@ -1,13 +1,4 @@
-import {
-  Heading,
-  Tag,
-  Box,
-  Flex,
-  Image,
-  Grid,
-  GridItem,
-  Text,
-} from '@chakra-ui/react';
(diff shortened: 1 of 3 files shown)


ojeytonwilliams:
I'm not sure, honestly, but this seems fine. There doesn't seem to be entirely appropriate semantic element for cards.  `li`, maybe, since they are elements in a list :man_shrugging:

ojeytonwilliams (approved):
LGTM :+1:

PR #2068, opened 2022-12-08
feat: email user for change in chapter role

Sboonny:

Ref #2037

### DIFF
--- a/server/src/controllers/ChapterUser/resolver.ts
+++ b/server/src/controllers/ChapterUser/resolver.ts
@@ -122,26 +122,17 @@ interface Args {
   }>;
 }
 
-type EmailProps =
-  | {
-      email: string[];
-      emailSubject: string;
-      emailBody: string;
-    }
-  | {
-      email?: never;
-      emailSubject?: never;
-      emailBody?: never;
-    };
+type EmailProps = {
+  email: string[];
+  emailSubject: string;
+  emailBody: string;
+};
 
 async function updateInstanceRoleForChapterRoleChange({
   changedChapterId,
   newChapterRole,
   user,
-  email,
-  emailSubject,
-  emailBody,
-}: EmailProps & Args) {
+}: Args) {
   const oldInstanceRole = user.instance_role.name;
   const userChapters = user.user_chapters;
   const newInstanceRole = getInstanceRoleName({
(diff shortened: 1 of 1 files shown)


server/src/controllers/ChapterUser/resolver.ts
```diff
@@ -140,6 +156,13 @@ async function updateInstanceRoleForChapterRoleChange({
       where: { id: user.id },
     });
   }
+  if (email && emailSubject && emailBody) {
+    await new MailerService({
+      emailList: email,
+      subject: emailSubject,
+      htmlEmail: emailBody,
+    }).sendEmail();
+  }
```
ojeytonwilliams, line 165: This mixes concerns and would be better as two functions.  Previously `updateInstanceRoleForChapterRoleChange` did exactly that, but these additions make that name a lie.

Instead it's perfectly fine to create a helper function `emailUserAboutRoleChange`, say, and call that with whatever it needs (user, role, etc)

server/src/controllers/ChapterUser/resolver.ts
```diff
@@ -277,10 +300,18 @@ export class ChapterUserResolver {
       where: { user_id_chapter_id: { chapter_id: chapterId, user_id: userId } },
     });
 
+    const subject = `Your role in chapter has been changed to ${newChapterRole}`;
+    const body = `Hi ${chapterUser.user.name}.<br />
+    Your role in chapter has been changed to ${newChapterRole}.<br />
+    `;
```
ojeytonwilliams, line 306: These are a touch repetitive. How about `Role changed in ${chapter.name}` as the title and `Your role in chapter ${chapter.name} has been changed from ${oldChapterRole} to ${newChapterRole}`? Something like that.

ojeytonwilliams (changes requested):

server/src/controllers/ChapterUser/resolver.ts
```diff
@@ -264,6 +282,7 @@ export class ChapterUserResolver {
             user_chapters: { include: { chapter_role: true } },
           },
         },
+        chapter: true,
```
ojeytonwilliams, line 285:
```suggestion
        chapter: { select: { name: true } },
```
We only need the `name` after all.

ojeytonwilliams (changes requested):
Thanks, @Sboonny, this is looking good.  I gave it one final check and found one minor issue:

ojeytonwilliams (approved):
Thanks!

LGTM :+1:

PR #2077, opened 2022-12-09
fix: remove link when authenticated

gikf:

---

- `isDisabled` prop on the `Button` gives the visual cue, but it was still possible to click and follow the link.


Sboonny (approved):
Interesting, just wondering are we removing it from the DOM, because to stop people from copying it from the devtools and casing trouble?

> isDisabled prop on the Button gives the visual cue, but it was still possible to click and follow the link

well, that answer my question 🤣

ojeytonwilliams (changes requested):
I took a brief look into this and it seems we're kind of abusing semantic html a little here.

Since it's either

- A link to /authenticate-with-google
- An noninteractive element that tells you everything is fine, nothing to see here

it shouldn't ever be a button. Link in the first case and I'm not sure about the second (just a div?)

Also, until I start the the validation process with Google (next week, probably), I really need to be able to reauthenticate (since the refresh tokens only last 7 days).  So, until we have a way to check that the token is invalid and enable this link, I'm relying on the bug!

gikf:
> Also, until I start the the validation process with Google (next week, probably), I really need to be able to reauthenticate (since the refresh tokens only last 7 days). So, until we have a way to check that the token is invalid and enable this link, I'm relying on the bug!

That's indeed good reason to keep it for a bit. Token invalidation and proper way to reauthorize (making sure there's new refresh token to save) is in my near plan.

ojeytonwilliams:
Great stuff.  I can work with the bug for now, so it's all good.

gikf:
I've replaced `Button` with `LinkButton`, which seems to behave more as desired, with some changes. `LinkButton` appear to do things a bit differently when href is external, so I had to use the `as="div"` to force div element for it.

I think making it appear more distinctly than a _simple text_ link is beneficial, for the authentication action to stand out.

Sboonny (commented):
Instead of using `LinkButton` that has 2 interactive elements, and turning one of them into a `div`, we can stop showing the Button if they are authenticated.

Something like:

```ts
      {isAuthenticated ?
          <Button
            background={'gray.85'}
            color={'gray.10'}
            marginBlock={'1em'}
            cursor="not-allowed"
            isDisabled={true}
          >
            Authenticated
          </Button>
          :
          <Button
            as='a'
            href={new URL('/authenticate-with-google', serverUrl).href}
            fontWeight="600"
            background={'gray.85'}
            color={'gray.10'}
            height={'100%'}
            marginBlock={'1em'}
            borderRadius={'5px'}
            paddingBlock={'.65em'}
            _hover={{ color: 'gray.85', backgroundColor: 'gray.10' }}
          >
            {isBroken
              ? 'Reauthenticate with Google'
              : 'Authenticate with Google'}
          </Button>
        }
```

I am saying this because turning `LinkButton` into a `div`, won't make sense for new developers, and they will find this odd without looking at the source code, and may try to change it.

gikf:
I'm not sure if I understand this correctly. That example still contains button for case when authenticated. Could you elaborate?

And to clarify, while the last commit uses `as="div"`, the result still has the button-like appearance. It's kind of the other way around, only way to notice something odd about this, is looking at code. Otherwise it seems to be a button.
image image

Sboonny:
I am not referring to how the button looks, don't worry.

---

In the source code of `LinkButton`, if we don't provide `justLink` it will add an interactive element inside an interactive element, which is violation for accessibility, you have worked around this by setting it as `div`.

https://github.com/Zeko369/chakra-next-link/blob/1123d9e614055d86f517902dff141ba76343437d/src/index.tsx#L56-L86

My point is, we don't need to work around it, because in the future, if we don't use 2 different buttons for different state (`isAuthenticated`, `!isAuthenticated`), we will have to add styles based on the state to the same component and make it too smart, for example add `cursor` if the `isAuthenticated` is true, or remove the `_hover` if `isAuthenticated` is true.

This smart pattern is worrying me, because it's harder to understand than this. 

```ts
if (isAuthenticated) return "display button indication that user is authenticated with Google Calendar"
else return "display button Please authenticate with Google Calendar"
```

gikf:
I'm not sure if complexity issue is here a concern at this point. The differing props can be always extracted elsewhere to reduce noise. Example below additionally create just single element.
```typescript
const enabledProps: LinkButtonProps = {
  as: 'div',
  isDisabled: true,
  _hover: { color: 'gray.10', background: 'gray.85' },
};
const disabledProps: LinkButtonProps = {
  nextAs: 'a',
  href: new URL('/authenticate-with-google', serverUrl).href,
  _hover: { color: 'gray.85', backgroundColor: 'gray.10' },
};

(...)

        <LinkButton
          {...(isAuthenticated ? enabledProps : disabledProps)}
          fontWeight="600"
          background={'gray.85'}
          color={'gray.10'}
          height={'100%'}
          marginBlock={'1em'}
          borderRadius={'5px'}
          paddingBlock={'.65em'}
        >
          {isAuthenticated
            ? 'Authenticated'
            : isBroken
            ? 'Reauthenticate with Google'
            : 'Authenticate with Google'}
        </LinkButton>
```

ojeytonwilliams (approved):
LGTM :+1:

PR #2081, opened 2022-12-10
feat: add focus outline to Button

Sboonny:

Ref #2044

This includes all the easy parts, there is linkButton and some which use maps or objects to set the styles, which will need refactoring.


ojeytonwilliams (changes requested):
This is both semantically a bit weird and quite repetitive. What I mean is that 'outline' isn't a special variant of the Button, it's just the default (by default, buttons should get an outline that's clearly visible). That leads to unnecessary repetition.

If we use a default style all of that goes away.

Sboonny:
Maybe, i am not typing the syntax right. But this is the second time, I use global styles in chakra and it doesn't work, I have tried to add dark more before and I need to use global for this, but failed

ojeytonwilliams:
It depends what the goal is.

If you want a global style (i.e. add css for every page) then https://chakra-ui.com/docs/styled-system/global-styles is what you need. I guess that makes sense if you're changing the outline for everything. No idea if that's helpful.

If you want to style just Button, then https://chakra-ui.com/docs/components/button/theming is what you need.

Sboonny:
global-styles is what I want, because I won't need to stand behind someone shoulder telling them "btw you need to add variant, because the original outline isn't accessible", but adding it doesn't work, I mean literally doesn't work.

If you go to localhost/3000, it won't change the button outlines, although they are in the DOM a button component, they won't get affected

ojeytonwilliams:
One thing I would keep in mind is that global styles should be used sparingly and typically for styles that need to be applied to multiple components.

If the plan really is to only modify a single component (Button), then there are tools for that. The variants can all extend the `baseStyle`, so you won't need to tell anyone which variant they have to use.

Sboonny:
it actually worked, lets gooooo 😭 🥳

ojeytonwilliams (approved):
That's what I was hoping we could do :rocket: 

LGTM :+1:

PR #2082, opened 2022-12-10
feat: ensure refresh_token is obtained when authenticating

gikf:

---

- If during authentication with calendar api there isn't `refresh_token` obtained, and there isn't already entry, redirects to `prompt: 'consent'` to force receiving it after authentication.
- Second case - no `refresh_token`, but existing entry, might need some adjusting. After playing around with invalidating tokens (another PR), I'm a bit worried updating refresh token might not happen when there's existing entry. Maybe just always use `prompt: 'consent'` when authenticating or also throw missing refresh token error (what will trigger using `prompt: 'consent'`) during authentication?
- I'm both surprised, by what had to do to get exception thrown in `onTokens`, to eventually end up in the route callback, and amazed that it worked. I'm assuming though there's must be some way simpler solution for that. Any alternative ideas would be appreciated.


server/src/services/InitGoogle.ts
```diff
@@ -92,36 +92,46 @@ async function onTokens(tokens: Credentials) {
...
+        onTokens(tokens).then(resolve).catch(reject);
+      });
+    }),
+    new Promise((resolve, reject) => {
+      oauth2Client
+        .getToken(code)
+        .then(resolve)
+        .catch(() => reject(new Error('Failed to get tokens')));
+    }),
+  ]);
```
ojeytonwilliams, line 124: I see what you mean.  This is tricky.

I'll sleep on it, but what I think is that I'd do this:
```suggestion

  try {
    const tokens = (await oauth2Client.getToken(code)).tokens;
    await onTokens(tokens);
  } catch (e) {
    console.error('Failed to get tokens');
    throw redactSecrets(e);
  }
```

I'm not sure what we should be doing to _other_ `on('tokens')` errors. Ideally we should communicate the issue to the client, somehow, but I'm not sure we want to be messing around with websockets just yet!

server/src/services/InitGoogle.ts
```diff
@@ -92,36 +92,46 @@ async function onTokens(tokens: Credentials) {
...
-    // a new refresh token.
     await prisma.google_tokens.update({
       where: { id: TOKENS_ID },
       data: { ...update },
     });
+  } else {
+    // TODO: Handle the case where the refresh token is not sent *and* the
+    // record doesn't exist. If this happens, we need to redirect them to a
+    // new auth url, but with prompt: 'consent', so that Google will provide
+    // a new refresh token.
```
ojeytonwilliams, line 105:
```suggestion
    // TODO: this should not happen, but if it did, presumably something went
    // wrong during the Google authentication flow. All we can do is ask the
    // user to try again.
```

server/src/services/InitGoogle.ts
gikf: Thank you, it's much better now.

Because there are situations when lack of refresh_token is okay, I added optional parameter for it.

server/src/services/InitGoogle.ts
```diff
@@ -92,41 +100,46 @@ async function onTokens(tokens: Credentials) {
...
     state,
+    ...(prompt && { prompt: 'consent' }),
   });
 }
 
 async function createCredentialedClient() {
-  const oauth2Client = createOAuth2Client().on('tokens', onTokens);
+  const oauth2Client = createOAuth2Client().on('tokens', (tokens) =>
+    onTokens(tokens),
+  );
```
ojeytonwilliams, line 142:
```suggestion
  const oauth2Client = createOAuth2Client().on('tokens', onTokens);
```
TS is clever enough to recognise that the second argument is optional, so there's no need to wrap the function call.

ojeytonwilliams (changes requested):
Nice work, @gikf. Other than one small refactor this looks ready to me.

server/src/services/InitGoogle.ts
gikf: Thanks, sometimes I'm getting in a create-functions-everywhere mode.

ojeytonwilliams (approved):

PR #2083, opened 2022-12-11
feat: token invalidation

gikf:

---

- I was playing a bit with invalid (invalid, wrong or empty) tokens as saved ones, errors I've encountered are in the `tokenErrors`.
- Sends email informing about invalid token.

### DIFF
--- a/server/src/services/Google.ts
+++ b/server/src/services/Google.ts
@@ -1,23 +1,41 @@
 import type { calendar_v3 } from '@googleapis/calendar';
 
 import { isProd, isTest } from '../config';
-import { createCalendarApi } from './InitGoogle';
+import { createCalendarApi, invalidateToken } from './InitGoogle';
 interface CalendarData {
   summary: string;
   description: string;
 }
 
+const tokenErrors = [
+  'Invalid Credentials',
+  'invalid_grant',
+  'No access, refresh token, API key or refresh handler callback is set.',
+];
+
+async function errorHandler(err: unknown) {
+  if (err instanceof Error && tokenErrors.indexOf(err.message) !== -1) {
+    console.error('Marking token as invalid');
+    await invalidateToken();
+    // TODO send email to users with GoogleAuthenticate permission
+  }
+}
+
 export async function createCalendar({ summary, description }: CalendarData) {
   const calendarApi = await createCalendarApi();
 
-  const { data } = await calendarApi.calendars.insert({
-    requestBody: {
-      summary,
-      description,
-    },
-  });
-
-  return data;
+  try {
+    const { data } = await calendarApi.calendars.insert({
+      requestBody: {
+        summary,
+        description,
+      },
+    });
+    return data;
+  } catch (err) {
+    await errorHandler(err);
+    throw err;
+  }
 }
 
 interface EventData {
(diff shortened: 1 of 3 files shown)


server/src/services/Google.ts
```diff
@@ -1,21 +1,47 @@
...
+}
+
+async function wrapWithHandler<T>(func: GaxiosPromise<T>): GaxiosPromise<T> {
+  try {
+    return await func;
+  } catch (err) {
+    await errorHandler(err);
+    throw err;
+  }
+}
```
ojeytonwilliams, line 32: This is a tiny bit sketchy, because we're creating the promise and then wrapping it in a try catch block. It's not wrong, by any means, it's just that we're relying on the fact that it's only synchronous code between the creation and the awaiting of the promise.

Obviously that's the case here, but we can completely eliminate potential issues by passing a function to `wrapWithHandler`.  Also, how about `callWithHandler`? Sort of the same, but slightly more explicit.

server/src/services/Google.ts
```diff
@@ -1,21 +1,47 @@
...
 export async function createCalendar({ summary, description }: CalendarData) {
   const calendarApi = await createCalendarApi();
 
-  const { data } = await calendarApi.calendars.insert({
-    requestBody: {
-      summary,
-      description,
-    },
-  });
+  const { data } = await wrapWithHandler<calendar_v3.Schema$Calendar>(
```
ojeytonwilliams, line 37: No need to supply the generic param here, TS can infer it from the argument.

ojeytonwilliams (approved):
I like this approach. It looks like the way to go.

server/src/services/Google.ts
ojeytonwilliams: I should have written some code...

```ts
async function callWithHandler<T>(func: () => GaxiosPromise<T>): GaxiosPromise<T> {
  try {
    return await func();
```

Should do the trick.

gikf: Do you mean to also change the way it's called?

Ie.
```typescript
  const { data } = await callWithHandler(() =>
    calendarApi.calendars.insert({
      requestBody: {
        summary,
        description,
      },
    }),
  );
```

I initially tried something similar in the wrapper, but without wrapping `calendarApi` in anonymous function like above. Of course typescript complained about that, hence the weird try with _await func_ construct.

ojeytonwilliams: Yeah, that's what I had in mind. It's maybe a little defensive, but what I'm thinking of are situations like this

```js
var error = Promise.reject('error');
var returnsError = function () {
  return Promise.reject('error from function');
};
var success = new Promise(function (resolve) {
  setTimeout(function () {
    resolve('success');
  }, 1000);
});

async function bad() {
  await success;
  try {
    await error;
  } catch (e) {
    console.log('bad', e);
    console.log('error');
  }
}

bad();

async function good() {
  await success;
  try {
    await returnsError();
  } catch (e) {
    console.log('good', e);
    console.log('error');
  }
}

good();

```

If you comment out the `bad` call, it works just fine, but with it in you end up with an UnhandledErrorException.

I just want to make clear, I'm not implying you're doing anything like this. It's just that I saw the Promise being created outside of a try catch block and I got to thinking about what could go wrong!

ojeytonwilliams (approved):
Nice work, @gikf. I'll deploy this right away :rocket:

PR #2084, opened 2022-12-12
feat: added RSVP functionality for logged out users when using the ask_to_confirm query param

### LINKED ISSUE #2006: Ask to login when using the ask_to_confirm query param
Due to @gikf's hard work, https://github.com/freeCodeCamp/chapter/pull/1990, we'll soon be able to handle repeated RSVP attempts.

Once that's in it would be nice if the ask_to_confirm links worked for logged out users.  At the moment they function as a toggle - they ask users to confirm cancellation if attending and to confirm attendance if not. However, we only use them as invitations, so we can simplify the logic so it just tries to RSVP.

Tasks:

- [ ] Convert ask_to_confirm links from toggle to just attempt to RSVP
- [ ] Login when the user confirms
- [ ] Update the modal to say "Would you like to login and attend?" or words to that effect

spham67:

Closes #2006

To address this functionality, there were 4 subtasks done: 

1) The ask_to_confirm query param was changed to confirm_rsvp.  This is to avoid confusion with what the query param is actually doing.  Intuitively, the logic is that for any user (logged in or logged out), if they hit events/{id}/confirm_rsvp=true then they are trying to rsvp for the event

2) The toggle functionality that cases on rsvpStatus was removed.  Instead any instance where confirm_rsvp=true will result in a an attempt to rsvp via the tryToRsvp function.  

3) The user is then prompted with a confirmation modal that cases on whether or not they are logged in or not.  Logged out users will be asked to log in first whereas logged in users will be directly asked for RSVP confirmation.

4) Updating the event cypress tests to account for this new RSVP workflow

### DIFF
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -113,9 +113,10 @@ export const EventPage: NextPage = () => {
   }
 
   async function onCancelRsvp() {
-    const ok = await confirm({
-      title: 'Are you sure you want to cancel your RSVP?',
-    });
+    // const ok = await confirm({
+    //   title: 'Are you sure you want to cancel your RSVP?',
+    // });
+    const ok = true;
 
     if (ok) {
       try {
@@ -146,28 +147,29 @@ export const EventPage: NextPage = () => {
           ),
         }
       : {
-          title: 'Join this event?',
+          title: 'Would you like to login and join this event?',
           body: `Note: joining this event will make you a member of the event's chapter.`,
         };
-    const ok = await confirm(confirmOptions);
+    const ok = true; //await confirm(confirmOptions);
 
     if (ok) {
       if (user) {
         await onRsvp(rsvpStatus);
         return;
+      } else if (!user) {
+        await confirm(confirmOptions);
+        modalProps.onOpen();
+        const {
+          data: { me },
+        } = await login();
+        modalProps.onClose();
+        const eventUser = data?.event?.event_users.find(
+          ({ user: event_user }) => event_user.id === me?.id,
+        );
+        await onRsvp(eventUser?.rsvp.name);
       }
-      modalProps.onOpen();
-      const {
-        data: { me },
-      } = await login();
-      modalProps.onClose();
-      const eventUser = data?.event?.event_users.find(
-        ({ user: event_user }) => event_user.id === me?.id,
-      );
-      await onRsvp(eventUser?.rsvp.name);
     }
   }
-
   // TODO: reimplment this the login modal with Auth0
   async function checkOnCancelRsvp() {
     await onCancelRsvp();
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -44,7 +44,7 @@ import { useLogin } from '../../../hooks/useAuth';
 export const EventPage: NextPage = () => {
   const { param: eventId } = useParam('eventId');
   const router = useRouter();
-  const { user, loadingUser} = useAuth();
+  const { user, loadingUser } = useAuth();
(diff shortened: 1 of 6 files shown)


cypress/e2e/events/[eventId].cy.ts
```diff
@@ -87,7 +87,7 @@ describe('event page', () => {
     cy.login(users.testUser.email);
     cy.joinChapter(chapterId).then(() => {
       cy.rsvpToEvent({ eventId, chapterId }).then(() => {
-        cy.visit(`/events/${eventId}?ask_to_confirm=true`);
+        cy.visit(`/events/${eventId}?confirm_rsvp=true`);
 
         cy.contains('Are you sure you want to cancel your RSVP?');
         cy.findByRole('button', { name: 'Confirm' }).click();
```
Comment (gikf) [inline cypress/e2e/events/[eventId].cy.ts:93]: I cannot select in github whole test, but test checking if it's possible to cancel with `?confirm_rsvp=true` won't be needed with the changes.

client/src/modules/events/pages/eventPage.tsx
```diff
@@ -149,37 +146,37 @@ export const EventPage: NextPage = () => {
...
-  }
-
   useEffect(() => {
     if (canShowConfirmationModal && !hasShownModal) {
-      if (!rsvpStatus || rsvpStatus === 'no') {
-        checkOnRsvp({ invited: true });
-      } else {
-        checkOnCancelRsvp();
-      }
+      tryToRsvp();
```
gikf, line 179:
```suggestion
      tryToRsvp({ invited: true });
```
This is reason why one of the tests fails

gikf (commented):
It's really nice. I've added specific suggestions regarding failing tests. I'd be good to add now another test for not logged user, but that doesn't need to be necessarily in this PR.

I've also tried to restructure `tryToRsvp` a bit, to try to limit going deep into multiple brackets. It's not possible to suggest it in github. See what do you think about it.
```typescript
  async function tryToRsvp(options?: { invited?: boolean }) {
    const confirmOptions = options?.invited
      ? {
          title: 'You have been invited to this event',
          body: (
            <>
              {user
                ? 'Would you like to attend?'
                : 'Would you like to log in and join this event?'}
              <br />
              Note: joining this event will make you a member of the
              event&apos;s chapter.
            </>
          ),
        }
      : {
          title: 'Join this event?',
          body: `Note: joining this event will make you a member of the event's chapter.`,
        };

    const ok = await confirm(confirmOptions);
    if (!ok) return;

    if (user) {
      await onRsvp(rsvpStatus);
      return;
    }
    modalProps.onOpen();
    const {
      data: { me },
    } = await login();
    modalProps.onClose();
    const eventUser = data?.event?.event_users.find(
      ({ user: event_user }) => event_user.id === me?.id,
    );
    await onRsvp(eventUser?.rsvp.name);
  }
```

cypress/e2e/events/[eventId].cy.ts
Comment (spham67) [reply cypress/e2e/events/[eventId].cy.ts:93]: I see now.  Thank you!

client/src/modules/events/pages/eventPage.tsx
spham67: Got it thanks!

gikf (approved):
Looks good :+1:

spham67:
> Looks good 👍

Hi @gikf, do you mind opening another issue for adding new event cypress tests for not logged users?

Sboonny:
#### :sparkles: :tada: **AWESOME!** :tada: :sparkles:

 Hi @spham67, thanks for this pull request and for contributing to the code-base for the first time. We are looking forward to more contributions from you in the future.

 Cheers & happy contributing!

gikf:
@all-contributors would you be so kind and add @spham67 for code ?

PR #2091, opened 2022-12-13
chore: refactor MailerService

### LINKED ISSUE #2070: Refactor email service
To send an email we first instantiate a MailerService with the `emailList`, `subject` and so on.  Then, to send the email, we call `sendEmail` with no arguments. i.e. we initialize the service every time we want to use it.

A more intuitive pattern would be to initialize the service once in the MailerService module and then use that object whenever we need to send an email.  This wouldn't be a huge change. For example, sending an email would go from

```ts
new MailerService({
        emailList: [email],
        subject,
        htmlEmail,
        ...options,
      }).sendEmail()
```

to

```ts
mailerService.sendEmail({
        emailList: [email],
        subject,
        htmlEmail,
        ...options,
      })
```

Jonathan1599:

Closes #2070

Refactored MailerService to make sure only a single instance can be initialized. 
- Removed setting values from the constructor to `sendEmail`
- Created another function `loadValues` to get back values used in `Mailer.test.ts`


server/src/services/MailerService.ts
```diff
@@ -86,7 +63,51 @@ export default class MailerService {
...
+    this.emailPassword = process.env.EMAIL_PASSWORD || 'secret.1';
+    this.emailService = process.env.EMAIL_SERVICE;
+    this.emailHost = process.env.EMAIL_HOST || 'localhost';
+
+    if (!process.env.SENDGRID_KEY || !process.env.SENDGRID_EMAIL) {
+      this.createTransporter();
+    }
+
+    return this;
+  }
```
ojeytonwilliams, line 85: I don't believe we need this function.  Reason being, we can always instantiate `MailerService` if we want to test it.

Also, since we're passing `MailerData` to `sendEmail`, it's no longer necessary to keep the details of a specific email (`emailList`, `subject` and so on) as properties of the `MailerService`.

server/src/services/MailerService.ts
```diff
@@ -86,7 +63,51 @@ export default class MailerService {
...
+
+  public async sendEmail(data: MailerData): Promise<SentMessageInfo> {
+    this.emailList = data.emailList;
+    this.subject = data.subject;
+    this.htmlEmail = data.htmlEmail;
+    this.backupText = data.backupText || '';
+    this.iCalEvent = data.iCalEvent;
+
+    // to be replaced with env vars
```
ojeytonwilliams, line 94:
```suggestion
```
`data` can be passed on to `this._sendEmail`

server/src/services/MailerService.ts
```diff
@@ -86,7 +63,51 @@ export default class MailerService {
...
+
+    if (!process.env.SENDGRID_KEY || !process.env.SENDGRID_EMAIL) {
+      this.createTransporter();
+      this._sendEmail = this.sendViaNodemailer;
+    } else {
+      this.sendgridKey = process.env.SENDGRID_KEY;
+      this.sendgridEmail = process.env.SENDGRID_EMAIL;
+      sendgrid.setApiKey(this.sendgridKey);
+      this._sendEmail = this.sendViaSendgrid;
+    }
```
ojeytonwilliams, line 109: These are all properties of the `MailerService` and so should be added to it once during instantiation

server/src/services/MailerService.ts
```diff
@@ -86,7 +63,51 @@ export default class MailerService {
...
+
     try {
       return await this._sendEmail();
```
ojeytonwilliams, line 112:
```suggestion
      return await this._sendEmail(data);
```

ojeytonwilliams (changes requested):
Hi @Jonathan1599 thanks a ton for working on this.

My high level observation is that we should separate `MailerService` specific data (ourEmail, emailUsername and so on) from email specific data (`MailerData`).  Specific details below.

Also, server/src/controllers/Messages/resolver.ts and server/tests/Resolver.test.ts can simply be deleted. We don't use the mutation on client, so the server does not need to support it.  It makes sense to clear it out in this PR, if that's okay with you, because otherwise the tests will fail.

Jonathan1599:
There are a few tests in Mailer.test.ts that I thought weren't useful after the refactor so I commented them out for now. Let me know if they have to be modified or deleted.

server/tests/Mailer.test.ts
```diff
@@ -76,19 +76,19 @@ describe('MailerService Class', () => {
...
   });
 
-  it('Should provide a blank string if backup text is undefined', () => {
-    const mailer = new MailerService(data);
-    expect(mailer.backupText).toEqual('');
-  });
+  // it('Should provide a blank string if backup text is undefined', () => {
+  //   const mailer = new MailerService();
+  //   expect(mailer.backupText).toEqual('');
+  // });
```
ojeytonwilliams, line 82:
```suggestion
```
Yeah, this can go. Thanks!

server/tests/Mailer.test.ts
```diff
@@ -76,19 +76,19 @@ describe('MailerService Class', () => {
...
 
-  it('Should default iCalEvent to undefined', () => {
-    const mailer = new MailerService(data);
-    expect(mailer.iCalEvent).toBeUndefined();
-  });
+  // it('Should default iCalEvent to undefined', () => {
+  //   const mailer = mailerService.loadValues(data);
+  //   expect(mailer.iCalEvent).toBeUndefined();
+  // });
```
ojeytonwilliams, line 87:
```suggestion
```

server/tests/Mailer.test.ts
```diff
@@ -76,19 +76,19 @@ describe('MailerService Class', () => {
...
-  it('Should log a warning if emailUsername, emailPassword, or emailService is not specified', () => {
-    // stub(Utilities, 'allValuesAreDefined').callsFake(() => false);
-    new MailerService(data);
-    expect(console.warn).toHaveBeenCalledTimes(1);
-  });
+  // it('Should log a warning if emailUsername, emailPassword, or emailService is not specified', () => {
+  //   // stub(Utilities, 'allValuesAreDefined').callsFake(() => false);
+  //   const mailer = new MailerService();
+  //   expect(console.warn).toHaveBeenCalledTimes(1);
+  // });
```
ojeytonwilliams, line 93:
```suggestion
  it('Should log a warning if emailUsername, emailPassword, or emailService is not specified', () => {
    new MailerService();
    expect(console.warn).toHaveBeenCalledTimes(1);
  });
```
While the validation can definitely be improved, I think this is better than nothing.

server/tests/Mailer.test.ts
```diff
@@ -44,30 +44,30 @@ describe('MailerService Class', () => {
...
-    expect(backupText).toEqual(mailer.backupText);
-    expect(calendar).toEqual(mailer.iCalEvent);
-    expect(emailAddresses).toEqual(expect.arrayContaining(mailer.emailList));
-  });
+  //   expect(subject).toEqual(mailer.subject);
+  //   expect(htmlEmail).toEqual(mailer.htmlEmail);
+  //   expect(backupText).toEqual(mailer.backupText);
+  //   expect(calendar).toEqual(mailer.iCalEvent);
+  //   expect(emailAddresses).toEqual(expect.arrayContaining(mailer.emailList));
+  // });
```
ojeytonwilliams, line 63:
```suggestion
```

server/tests/Mailer.test.ts
```diff
@@ -44,30 +44,30 @@ describe('MailerService Class', () => {
...
     const { auth, service } = mailer.transporter.options as any;
     expect(service).toEqual(mailer.emailService);
     expect(auth.user).toEqual(mailer.emailUsername);
     expect(auth.pass).toEqual(mailer.emailPassword);
   });
 
-  it('Should correctly instantiate values', () => {
-    const mailer = new MailerService(dataWithOptional);
+  // it('Should correctly instantiate values', () => {
+  //   const mailer = new MailerService();
```
ojeytonwilliams, line 56:
```suggestion
```

server/tests/Mailer.test.ts
```diff
@@ -7,32 +7,32 @@ const emailAddresses = [
...
-const dataWithOptional: MailerData = {
-  ...data,
-  backupText: backupText,
-  iCalEvent: calendar,
-};
+// const dataWithOptional: MailerData = {
+//   ...data,
+//   backupText: backupText,
+//   iCalEvent: calendar,
+// };
```
ojeytonwilliams, line 36:
```suggestion
```

server/tests/Mailer.test.ts
```diff
@@ -7,32 +7,32 @@ const emailAddresses = [
...
+// BEGIN:VEVENT
+// UID:3b4a4873-161f-43e0-873d-10aa90488339
+// SEQUENCE:0
+// DTSTAMP:20220403T072207Z
+// DTSTART:20220325T234500Z
+// DTEND:20220326T014500Z
+// SUMMARY:Rolfson, Emmerich and Davis
+// URL;VALUE=URI:http://localhost:3000/events/15?ask_to_confirm=true
+// END:VEVENT
+// END:VCALENDAR`;
```
ojeytonwilliams, line 23:
```suggestion
```

ojeytonwilliams (changes requested):
Thanks for going through all those tests, @Jonathan1599. There's one that still seems worth having, but the rest can go.

ojeytonwilliams (approved):
Nice work, @Jonathan1599 this looks great!

PR #2092, opened 2022-12-13
fix: move key prop to outer element

gikf:

---

- Fixes console warning _Each child in a list should have a unique "key" prop_ by moving `key` prop to the outer element.
- Short syntax doesn't allow adding props, so changing it to `Fragment` was needed.


Sboonny (approved):

PR #2099, opened 2022-12-14
chore: events resolver clean up

gikf:

---

- Removes unused query.
- Stops including event roles, and `subscribed` field of `event_user` in the event query. `subscribed` for the logged in user is taken now from `meQuery`.
- Stops the need to always include role permissions with the event user role.
  - For that was required to shuffle a bit `EventUser` object types. Naming is very rough, to not say horrible. I haven't thought of anything _better_ yet, that's not even longer.

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -26,7 +26,8 @@ import { ResolverCtx } from '../../common-types/gql';
 import {
   Event,
   EventUserWithRelations,
-  EventWithRelations,
+  EventWithRelationsWithEventUserRelations,
+  EventWithRelationsWithEventUser,
   EventWithChapter,
   EventWithVenue,
   User,
@@ -63,11 +64,7 @@ import { EventInputs } from './inputs';
 const eventUserIncludes = {
   user: true,
   rsvp: true,
-  event_role: {
-    include: {
-      event_role_permissions: { include: { event_permission: true } },
-    },
-  },
+  event_role: true,
 };
 
 type EventWithUsers = Prisma.eventsGetPayload<{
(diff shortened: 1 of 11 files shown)


server/src/controllers/Events/resolver.ts
```diff
@@ -289,38 +286,6 @@ const getNameForNewRsvp = (event: EventRsvpName) => {
 
 @Resolver()
 export class EventResolver {
-  @Query(() => [EventWithRelations])
-  async events(
-    @Arg('limit', () => Int, { nullable: true }) limit?: number,
-    @Arg('showAll', { nullable: true }) showAll?: boolean,
-  ): Promise<EventWithRelations[]> {
```
ojeytonwilliams, line 296: I can't believe this has hung around so long.  Thanks for the clean up!

server/src/controllers/Events/resolver.ts
```diff
@@ -365,27 +330,18 @@ export class EventResolver {
...
   }
 
-  // TODO: Check we need all the returned data
   @Authorized(Permission.EventEdit)
-  @Query(() => EventWithRelations, { nullable: true })
+  @Query(() => EventWithRelationsWithEventUserRelations, { nullable: true })
   async dashboardEvent(
     @Arg('id', () => Int) id: number,
-  ): Promise<EventWithRelations | null> {
+  ): Promise<EventWithRelationsWithEventUserRelations | null> {
```
ojeytonwilliams, line 337: I see what you mean about the name, but it's okay. At least, I can't come up with good names for these, either.

ojeytonwilliams (approved):
Always nice to be able to remove an entire query resolver :cold_sweat: 

LGTM :+1:

server/src/controllers/Events/resolver.ts
gikf: It might have been still used somewhat recently in events dashboard, before `dashboardEvents` query was made.

ojeytonwilliams: Most likely. I probably just forgot that /events doesn't use it 🤷

PR #2100, opened 2022-12-15
fix: disambiguation invitation and confirmation

ojeytonwilliams:

Using "Invitation" for both was quite confusing.

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -106,12 +106,12 @@ const sendRsvpInvitation = async (
 
   await new MailerService({
     emailList: [user.email],
-    subject: `Invitation: ${event.name}`,
-    htmlEmail: `Hi${user.name ? ' ' + user.name : ''},</br>
-To add this event to your calendar(s) you can use these links:
-</br>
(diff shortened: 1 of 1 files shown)


Sboonny (approved):

server/src/controllers/Events/resolver.ts
```diff
@@ -106,12 +106,12 @@ const sendRsvpInvitation = async (
...
-To add this event to your calendar(s) you can use these links:
-</br>
+    subject: `Confirmation of attendance: ${event.name}`,
+    htmlEmail: `Hi${user.name ? ' ' + user.name : ''},<br>
+You should receive a calendar invite shortly. If you do not, you can add the event to your calendars by clicking on the links below:<br>
+<br>
 <a href=${google(linkDetails)}>Google</a>
-</br>
+<br>
 <a href=${outlook(linkDetails)}>Outlook</a>
```
Sboonny, line 115: Do we have outlook way of RSVPing? should we remove this?

ojeytonwilliams: At this point the user has RSVPed, that's what triggers this email.  The `calendar-link` package generates links so that users can add this event to their calendars directly.

PR #2102, opened 2022-12-15
chore: remove unused arg from sendEventInvite

ojeytonwilliams:

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -973,7 +973,7 @@ ${unsubscribeOptions}`,
       event.chapter.user_bans.map(({ user_id }) => user_id),
     );
 
-    const users: User[] =
+    const users =
       event.chapter.chapter_users?.filter(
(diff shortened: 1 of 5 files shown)


server/src/controllers/Events/resolver.ts
```diff
@@ -984,35 +973,10 @@ ${unsubscribeOptions}`,
...
-      );
-      users.push(...confirmedUsers);
-    }
-    if (emailGroups.includes('canceled')) {
-      const canceledUsers = event.event_users.filter(
-        ({ rsvp }) => rsvp.name === 'no',
-      );
-      users.push(...canceledUsers);
-    }
+    const users: User[] =
```
gikf, line 976:
```suggestion
    const users =
```
Explicit type is no longer needed, along the interface definition (not possible to suggest in github).

gikf (commented):
One suggestion below for small thing that also can be removed.

gikf (approved):
Looks good :+1:

PR #2103, opened 2022-12-15
feat: enable auto_subscribe by default

ojeytonwilliams:
Since users can be invited to a chapter by shared links, they may not
realise that they can subscribe to chapters.

Since auto_subscribe can be controlled via the profile, and users can
unsubscribe easily via the links in each email it seems reasonable to
have people subscribed to chapters they join by default.

I know we (and I) went back and forth on this on the original issue, but I'd like to try this approach. I'll check in with the test users tomorrow, but I think they're unaware that subscribing is a thing they might want to do. I'll leave this blocked until then

@gikf @allella let me know if you've had any more thoughts on this.


gikf (approved):
Looks good :+1:

PR #2115, opened 2022-12-18
feat: don't update calendar attendees for events in the past

### LINKED ISSUE #1661: Members of the chapter shouldn't get notified by the creation of an event in the past
### Describe the feature:

As discussed in Discord, owner may want to create events for documentation purposes, but creating this events may result in notification been sent to the chapter members. 

We need some flag which stops the events, that were created before (`new Date()`), from using our notification system

gikf:

Closes #1661

---

- Doesn't include attendees when creating calendar event. (This must be included, as creating calendar event can happen also after creation of the event).
- Ignores all attendee changes for event in the past. Maybe it should be ignored only for adding new attendee?

### DIFF
--- a/server/src/services/Google.ts
+++ b/server/src/services/Google.ts
@@ -9,6 +9,27 @@ interface CalendarData {
   description: string;
 }
 
+interface EventData {
+  start?: Date;
+  end?: Date;
+  summary?: string;
+  attendees?: calendar_v3.Schema$EventAttendee[];
+  status?: 'cancelled';
+}
+
+interface CalendarId {
+  calendarId: string;
+}
+
+interface EventIds {
+  calendarId: string;
+  calendarEventId: string;
+}
+
+interface Attendee {
+  attendeeEmail: string;
+}
+
 const tokenErrors = [
   'Invalid Credentials',
   'invalid_grant',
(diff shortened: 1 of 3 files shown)


server/src/services/Google.ts
```diff
@@ -69,7 +70,9 @@ function parseEventData({
     ...(summary && { summary }),
     // Since Google will send emails to these addresses, we don't want to
     // accidentally send emails in testing.
-    ...(attendees && { attendees: isProd() || isTest() ? attendees : [] }),
+    // Additionally include attendees only for future events.
+    ...((!start || isFuture(start)) &&
+      attendees && { attendees: isProd() || isTest() ? attendees : [] }),
```
ojeytonwilliams, line 75:
```suggestion
    ...(attendees && { attendees: isProd() || isTest() ? attendees : [] }),
```
Unfortunately this doesn't quite work how you'd expect.  If you don't provide a list of attendees the calendar api will remove any existing attendees.

The approach has to be to send whatever you want the event to become after the call.  In this case, if it's an event update we just want to reflect back the existing attendee list.

For event creation, I think we just have to handle that as a special case.

ojeytonwilliams (dismissed):

server/tests/google.test.ts
```diff
@@ -175,6 +208,24 @@ describe('Google Service', () => {
...
+      const updatedEvent = await cancelEventAttendance(
+        { calendarId: 'id', calendarEventId: 'id' },
+        { attendeeEmail: 'b@person' },
+      );
+
+      const expectedAttendees = [
+        ...attendees.filter((attendee) => attendee.email !== 'b@person'),
+        { email: 'b@person', responseStatus: 'declined' },
+      ];
+
```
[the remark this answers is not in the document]
ojeytonwilliams: Err, above, actually.  Sorry, I always forget that the comments appear in the order you add them, not where they are in the code.

server/src/services/Google.ts
gikf: Hmm, I believe that's what's happening already. There's test for updating past event and ignoring changed attendees in the send `eventData` - https://github.com/gikf/chapter/blob/d73cb953ff5ad572bd58e2964b1a1d40e83eaca2/server/tests/google.test.ts#L316
I've locally added another one, updating future event, and it also kept attendees intact.

If commented code won't include `attendees` in the parsed data, which means that result of `createEventRequestBody` call won't have it, the original attendees from `data` will be used.

https://github.com/freeCodeCamp/chapter/blob/8d6d0145cfd24c6b3c5037d87bd36ecd97a92432/server/src/services/Google.ts#L140-L147

Or is there some another edge case?

ojeytonwilliams: Good point. No, I don't believe there is another edge case.

While I was digging into this, I did some pretty heavy refactoring.  Do you mind if I push a commit?

ojeytonwilliams: Cheers, I think this is a little better.

ojeytonwilliams:
If you're happy with those changes and want to merge as is, that's fine with me. That or incorporate the test suggestions.  Up to you.

server/tests/google.test.ts
```diff
@@ -175,6 +208,24 @@ describe('Google Service', () => {
...
+      expect(updatedEvent.data.attendees?.length).toBe(3);
+      expect(updatedEvent.data.attendees).toEqual(
+        expect.not.arrayContaining(expectedAttendees),
+      );
```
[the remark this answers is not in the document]
gikf: I've changed it to `not.toContainEqual`, apparently `toContain` would do identity check while comparing objects and test would always pass.

server/src/services/Google.ts
gikf: I agree, I like it.

gikf:
I double checked if tests for not updating attendees in past events are really failing, if attendees are changed.

I think we are good here.

server/src/services/Google.ts
```diff
@@ -8,6 +9,27 @@ interface CalendarData {
   description: string;
 }
 
+interface EventData {
+  start?: Date;
+  end?: Date;
```
Sboonny, line 14: Random question is it possible to have Event without `Date`? like it will have the wrong type of "Date" (because it's `any` in the client)

```suggestion
  start: Date;
  end: Date;
```

gikf: It's not possible. However when updating attendees date isn't passed as changed, but it's taken from `oldEventData`. https://github.com/freeCodeCamp/chapter/blob/c9100ae387e66622409b1cfa24e7d1246538270c/server/src/services/Google.ts#L118-L123

Sboonny (approved):

server/src/services/Google.ts
Sboonny: Thank you for clearing this

PR #2117, opened 2022-12-19
fix: emails breaking lines elements

### LINKED ISSUE #2089: Emails are formatted incorrectly
**Describe the bug**
Html emails are not displaying as expected.  Specifically `</br>` is being ignored. It could be as simple as we're using `</br>` not `<br>`, but I'm not sure.

**To Reproduce**
1. Sign in to https://chapter.freecodecamp.dev
2. Go to https://chapter.freecodecamp.dev/events/18
3. Click on RSVP
4. Check the email you signed in with
5. See "Hi name, To add this event to your calendar(s) you can use these links:"

**Expected behavior**

You should see 
>Hi name,
>To add this event to your calendar(s) you can use these links:

**Additional context**

I saw this in Gmail. Other email clients may display them differently.

Sboonny:

Closes #2089

### DIFF
--- a/server/reminders/reminders.ts
+++ b/server/reminders/reminders.ts
@@ -47,18 +47,18 @@ const reminderMessage = ({
   chapterUnsubscribeToken,
   eventUnsubscribeToken,
 }: ReminderMessageData) => {
-  return `[${event.name}](Link to the event page, like https://{instance domain name}/chapters/${event.chapter.id}]) organized by ${event.chapter.name} is happening soon.</br>
-</br>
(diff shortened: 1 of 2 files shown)


Sboonny (commented):
Because of accessibility concern for screen reader as they may read the existence of elements, but without content inside, which is annoying.

Should we remove the `<br />` and use `<p></p>` instead?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#accessibility_concerns

Sboonny:
Nvm, previous question, I have read #1263

ojeytonwilliams (approved):
Incidentally, I'm not sure why Mailhog is okay with `</br>`, which makes this irritating to test.  Regardless, I think this is correct.

LGTM :+1:

PR #2120, opened 2022-12-19
fix: prevent error on waiting for user to log in using Auth0

### LINKED ISSUE #1997: Uncaught error when Auth0 login form timeouts
To reproduce, with Auth0 login:
- Click Menu and Login to display login form
- Wait around 60s
- `Uncaught (in promise) Error: Login required`

Jonathan1599:

Closes #1997

Added a timeout of 400 seconds to loginWithPopup along with checking if the user is logged in before creating a session when using Auth0


ojeytonwilliams:
Hi @Jonathan1599, I was looking into the issue and your PR and I got to thinking that there should be a (relatively) simple solution.

We can use `isAuthenticated` from `useSession` (which checks if the dev user or auth0 user is authenticated) to conditionally `createSession` in a `useEffect`. That way if something goes wrong or the login times out, we never to `createSession`.

Also, we can use `const { error } = useAuth0()` if we want to handle error states rather than silently ignoring them.

Jonathan1599:
Yep, sure. Will wait for #2170

ojeytonwilliams:
Awesome, thanks for bearing with me on this one.

ojeytonwilliams:
Hey @Jonathan1599 https://github.com/freeCodeCamp/chapter/pull/2170 is in, so everything's ready if you want to take another look at this. No rush, though!

Jonathan1599:
This should work I think. One thing I'm confused about is if I should increase the timeout on `loginWithPopup`. I think it's currently set at 30s.
It didn't really cause a major issue when I tested it but what happens is that after the timeout, logging in causes the login popup window becomes empty and it stays open, giving me no choice but to close the window and click on our login button again, which logs me in without the pop up window.

gikf:
Closed per #2181

ojeytonwilliams:
Gah, I didn't realise you need multiple closes to close multiple issues.  Thanks!

gikf:
> Gah, I didn't realise you need multiple closes to close multiple issues. Thanks!

I didn't think about it before either.

PR #2121, opened 2022-12-20
feat: hide interaction action elements in dashboard

### LINKED ISSUE #1961: Use client permissions check to modify UI in dashboards
Added client side permission check for chapter and event roles, allows now to better modify UI, based on owned permissions. This will be more relevant when more granular user roles will be added.

Dashboards requiring adding UI modifications based on permission checks:
- [ ] Events
- [ ] Sponsors
- [ ] Venues
- [ ] Users

Pages in chapters dashboard have few examples of the permission checks.

Note: Client side check is not intended to replace server side permission check. It merely should hide from user actions that would be later rejected by server anyway. Server side permission check always should be there.

Sboonny:

Closes #1961 

Sponsor has add new and stuff already built in it 
https://github.com/Sboonny/chapter/blob/71fc140bc1f87f01e9977eb96cda6cdf33191933/client/src/modules/dashboard/Sponsors/pages/SponsorsPage.tsx#L19-L22


client/src/modules/dashboard/shared/components/Layout.tsx
```diff
@@ -51,6 +63,17 @@ export const Layout = ({
...
+    linksWithPermissions.filter(
+      (link) => !link.requiredPermission || link.hasPermission,
+    )
+  )
+    return (
+      <NextError
+        statusCode={500}
+        title={`You don't have access to view this page`}
+      />
+    );
```
Sboonny, line 76: This logic is wrong because it shows error page, when the page is loaded

Sboonny: I will open 2nd Issue or 2nd PR for this later

client/src/modules/dashboard/Events/pages/EventsPage.tsx
```diff
@@ -19,12 +21,18 @@ export const EventsPage: NextPageWithLayout = () => {
 
   const isLoading = loading || !data;
   if (isLoading || error) return <DashboardLoading error={error} />;
+  const hasPermissionToCreateEvent = checkPermission(
+    user,
+    Permission.EventCreate,
+  );
+
+  const hasPermissionToEditEvent = checkPermission(user, Permission.EventEdit);
```
gikf, line 29: Most of these checks need to have either `eventId` or `chapterId` passed. Otherwise only check will be done only on the instance permissions.

gikf: For current roles that would be enough. But to give ability to add custom and more granular role permissions, I think it would be better to pass the event id and/or chapter id to `checkPermission` as well. Even if so far these permissions would be only on some instance roles.

Sboonny: Last question 👍, do I add the ability to filter the data that aren't editable in this PR?

client/src/modules/dashboard/Events/pages/EventsPage.tsx
```diff
@@ -51,6 +53,16 @@ export const EventsPage: NextPageWithLayout = () => {
...
+  const hasPermissionToCreateEvent = checkHasChapterPermission(
+    user,
+    Permission.EventCreate,
+  );
+  const hasPermissiontoEditEvent = checkHasChapterPermission(
+    user,
+    Permission.EventEdit,
+  );
+
+  console.log(hasPermissionToCreateEvent, hasPermissiontoEditEvent);
```
Sboonny, line 65:
```suggestion
  console.log(hasPermissionToCreateEvent, hasPermissiontoEditEvent);
```

server/src/controllers/Events/resolver.ts
```diff
@@ -736,6 +736,7 @@ ${unsubscribeOptions}`,
     return updatedEvent ? updatedEvent : event;
   }
 
+  // How does admin have permission to edit, but when checked in the client it results in false
```
Sboonny, line 739:
```suggestion
```

ojeytonwilliams (changes requested):
Hey @Sboonny, I split `checkPermission` into two functions in https://github.com/freeCodeCamp/chapter/pull/2268, so you can use `checkChapterPermission` from client/src/util/check-permission.ts

The idea is that you might have permission for a specific Chapter, so we have to pass the chapter's `chapterId` to the function, so it can check that.

There are flaws with the functions, tracked in https://github.com/freeCodeCamp/chapter/issues/2269, but for this PR all you need do is pass `chapterId` in when calling `checkChapterPermission`.

ojeytonwilliams (changes requested):

client/src/util/check-permission.ts
```diff
@@ -1,5 +1,5 @@
 import { checker } from '../../../common/authorization';
-import {
+import type {
```
Sboonny, line 2:
```suggestion
import {
```

ojeytonwilliams (approved):

ojeytonwilliams (approved):

PR #2125, opened 2022-12-20
fix: no confirmation if user is already rsvp'd

### LINKED ISSUE #2108: Don't show the confirm modal when a user has already RSVP'd

When a user visits the page /events/[id]?confirm_rsvp=true of an event they have already RSVP'd to, the confirmation modal pops up.  This should not happen, since all that the user can do is Cancel (which does nothing) or Confirm (which just triggers the Already RSVPed toast).

It should be sufficient to skip this and just show the toast.

bryanleemoore:

Closes #2108

Removed `rsvpStatus` from the `onRsvp()` function, and instead created a function `isAlreadyRsvp()` that checks if the user is already RSVPd. If the user is already RSVPd, it toasts, and returns true. Otherwise, it does not toast and returns false.

This check would happen in the useEffect before calling `tryToRsvp()`, this way it will not call `tryToRsvp()` if the user is logged in and already RSVP'd. Instead, it will just do the "Already RSVP'd" toast and not show the confirmation modal. 

This check also would happen towards the end of the `tryToRSVP()` function in the case the user is logged out and the user redirects to `/events/[id]?confirm_rsvp=true`, as a confirmation modal is needed since we don't know if the user is RSVP'd or not as the user is logged out and needs to be prompted to log in.


client/src/modules/events/pages/eventPage.tsx
```diff
@@ -89,16 +89,20 @@ export const EventPage: NextPage = () => {
...
+    const alreadyRsvp = rsvpStatus === 'yes' || rsvpStatus === 'waitlist';
+    if (alreadyRsvp) {
+      toast({ title: 'Already RSVPed', status: 'info' });
+      return true;
+    }
+    return false;
+  }
 
-  async function onRsvp(rsvpStatus: string | undefined) {
+  async function onRsvp() {
```
Sboonny, line 101: your code gave me the hint, to test `onRSVP` function without it being `async`, and didn't find issue.

Do you want to remove it, and test it? or do we open new PR for that?

bryanleemoore: @Sboonny wouldn't the `onRSVP` function need to be `async` due to the `await joinChapter` and `await rsvpToEvent` statements inside that function?

Sboonny (approved):

client/src/modules/events/pages/eventPage.tsx
Sboonny: I meant to remove the promises except for `useConfirm`, because currently it is the only value we don't know.

I will create a new PR and test it there 👍

PR #2129, opened 2022-12-22
feat: clear extra promises of RSVP function in eventPage

Sboonny:

Thanks to @bryanleemoore, previous PR I noticed that we are using promises for function which have values already there, this aims to clear these promises. Didn't find bugs, and they maybe added for a reason, so let's test it to find the reason for their addition

### DIFF
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -98,14 +98,14 @@ export const EventPage: NextPage = () => {
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


gikf:
`joinChapter` and `rsvpToEvent` need to be executed in certain order. Otherwise user might not be part of the chapter when `rsvpToEvent` is executed. It might work when user is already part of chapter, because in such case the `joinChapter` effectively doesn't do anything.

Sboonny:
Yeab, makes sense 👍

PR #2140, opened 2022-12-26
feat: improve events list ux in dashboard

### LINKED ISSUE #2014: Dashboard event order and UX

By default the events should appear in descending order of date, since the upcoming events are likely the ones a user is interested in.

Also, it might be useful to separate out the cancelled events from the active ones.

gikf:

Closes #2014

---

- Orders events in dashboard in descending order of date.
- Adds switch to not include canceled events (displayed by default).
- Added parameter to the query makes it awkward, that refetching will not refetch both combinations without doing it explicitly :eyes:.
- One inconvenience is that during using the switch for the first time, it will disappear while loading new list from server. I'm not too enthusiastic about dealing with that problem in this PR. 
- I guess, as an alternative, we could do the filtering just in the client.


client/src/modules/dashboard/Events/pages/EventsPage.tsx
```diff
@@ -24,6 +55,10 @@ export const EventsPage: NextPageWithLayout = () => {
     <VStack data-cy="events-dashboard">
       <Flex w="full" justify="space-between">
```
Sboonny, line 56:
```suggestion
      <Flex w="full" justify="space-between" alignItems={{ base: "", sm: "center" }} flexDirection={{base:"column", sm:"row"}}>
```

client/src/modules/dashboard/Events/pages/EventsPage.tsx
```diff
@@ -12,8 +21,30 @@ import { useAuth } from '../../../auth/store';
...
+const ShowCanceledSwitch = ({
+  setShowCanceled,
+  defaultChecked,
+}: {
+  setShowCanceled: React.Dispatch<React.SetStateAction<boolean>>;
+  defaultChecked: boolean;
+}) => {
+  return (
+    <Flex>
+      <FormLabel htmlFor="show-canceled-events">Show canceled events</FormLabel>
```
Sboonny, line 33:
```suggestion
      <FormLabel htmlFor="show-canceled-events" marginBottom="0">Show canceled events</FormLabel>
```

Sboonny (commented):
Here are some styles, to change its looks to this, in mobile devices.

image

client/src/modules/dashboard/Events/pages/EventsPage.tsx
gikf: I'm not sure about this change, the `0` makes it a bit squeezed.
Before:
image
After:
image

Sboonny: the squeeze to zero is for the alignment in layout bigger than `base`, we can go with `marginBottom={{base: "", sm: "0"}}`

Sboonny (approved):

PR #2143, opened 2022-12-27
feat: display only events from venue

gikf:
- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `main` branch of Chapter.

---

- Narrows events displayed in the _Venue dashboard_ to events organized at the venue. Instead of displaying all events from the chapter.
- Additionally sorts these events in the descending order by date.

PR #2145, opened 2022-12-28
feat: add message for no events

gikf:

Ref https://github.com/freeCodeCamp/chapter/pull/2143#pullrequestreview-1230797439

---

- Adds message when there's no events.


Sboonny (approved):
I have 
- removed the `GridItem` because it wasn't used, and it was adding `div` without reason.
- removed `marginRight` from the `Tag` elements, and used `gap` in their `Flex` parent
- swapped `Heading` with `Text` element, because `Heading` is used to describe sections, and it shouldn't be a new section but text 🤔.

If you are okay with these changes, feel free to merge it

PR #2152, opened 2022-12-29
feat: test saved tokens

gikf:

Ref #1885

---

- Adds ability to run test on saved (valid so far) tokens, via button in `dashboard/calendar`.
- So far test tries to obtain from api `calendarList`.
- ~~Draft mainly because there's currently awkwardness while ensuring all valid tokens are tested, and the right token would be marked as invalid, if it fails. And as usual I'm a bit confused with the async looping.~~


server/src/services/InitGoogle.ts
```diff
@@ -159,14 +163,35 @@ async function createCredentialedClient() {
...
-  return calendar({ version: 'v3', auth });
+export async function createCalendarApi(
+  createClient: () => Promise<OAuth2Client> = async () =>
+    createCredentialedClient(
+      await prisma.google_tokens.findFirstOrThrow({
+        where: { is_valid: true },
+      }),
+    ),
+) {
+  return calendar({ version: 'v3', auth: await createClient() });
```
ojeytonwilliams, line 174:
```suggestion
export async function createCalendarApi(tokenInfo?: TokenInfo) {
  const tokens =
    tokenInfo ??
    (await prisma.google_tokens.findFirstOrThrow({
      where: { is_valid: true },
    }));
  const auth = await createCredentialedClient(tokens);
  return calendar({ version: 'v3', auth });
```
I found passing around `createClient` a little confusing and this approach seems more direct to me.

server/src/services/InitGoogle.ts
```diff
@@ -159,14 +163,35 @@ async function createCredentialedClient() {
...
+  });
+  const apis = [];
+  for (const tokenInfo of tokensInfo) {
+    apis.push({
+      tokenId: tokenInfo.id,
+      createApi: () =>
+        createCalendarApi(() => createCredentialedClient(tokenInfo)),
+    });
+  }
+  return apis;
```
ojeytonwilliams, line 189:
```suggestion
  return await Promise.all(
    tokensInfo.map(async (tokenInfo) => ({
      tokenId: tokenInfo.id,
      calendarApi: await createCalendarApi(tokenInfo),
    })),
  );
```
I prefer this approach for a few reasons. It's functional, it flattens the Promises (it's a Promise, not a Promise of Promises) and finally you don't have to name as many things :D

server/src/services/Google.ts
```diff
@@ -231,3 +236,21 @@ export async function deleteCalendarEvent({
     }),
   );
 }
+
+export async function testTokens() {
+  const apis = await createCalendarApis();
+
+  await Promise.all(
+    apis.map(async ({ tokenId, createApi }) => {
+      const calendarApi = await createApi();
```
ojeytonwilliams, line 245:
```suggestion
    apis.map(async ({ tokenId, calendarApi }) => {
```
The refactoring in `initGoogle` lets us write a simpler loop here.

ojeytonwilliams (changes requested):

server/src/services/InitGoogle.ts
gikf: :D I need to do this more often, because trying to find various names is not coming that great.

gikf:
@ojeytonwilliams Thanks, that makes things clearer and simpler.

ojeytonwilliams:
In case you're interested, I came across the 'trick' of avoiding naming things from Brian Will's youtube channel: https://www.youtube.com/@briantwill I forget which exactly which video, I'm afraid, but the channel as a whole is great.

Anyways, thanks for making those changes!

ojeytonwilliams (approved):

gikf:
Thanks, I think I found that portion in https://www.youtube.com/watch?v=2-G5WTghjYk _avoid naming as much as possible_. Other interesting insights as well :rocket:

PR #2158, opened 2023-01-01
chore: remove firefox downgrading

gikf:

---

- Removes firefox downgrade in cypress workflow. Cause for adding it is was supposed to be fixed in version `106`, current `108` doesn't have issue either.


Sboonny:
To save time, I don't think the issue of the failing test is from this PR, but in the main repo. 🤔

Sboonny:
wait does it fails, because there is no past date 🤣 ?

Sboonny (approved):

Sboonny:
LGTM, Thank you.

gikf:
> Wait, does it fail because there is no past date rofl, in cypress yet?
> 
> image

It's definitely not related with this PR (I hoped at some point, that maybe old firefox version was causing issue, but nah). I cannot reproduce that fail locally. I'm testing couple ideas for selecting date in another way, to see if that fixes issue.

PR #2173, opened 2023-01-04
fix: overflow dashboard links

gikf:

---

- In dashboard layout, links to specific pages, will now maintain minimum width and give ability to scroll when screen width is too small to fit all of them.
- I've tested this in mobile view.

### DIFF
--- a/client/src/modules/dashboard/shared/components/Layout.tsx
+++ b/client/src/modules/dashboard/shared/components/Layout.tsx
@@ -54,16 +54,23 @@ export const Layout = ({
 
   return (
     <div data-cy={dataCy}>
-      <HStack {...rest} data-cy="dashboard-tabs" as="nav" my="2">
+      <HStack
+        {...rest}
+        as="nav"
+        data-cy="dashboard-tabs"
+        my="2"
+        overflowX="auto"
+      >
         {linksWithPermissions
           .filter((link) => !link.requiredPermission || link.hasPermission)
           .map((item) => (
             <LinkButton
-              key={item.link}
-              href={item.link}
               colorScheme={
                 router.pathname.startsWith(item.link) ? 'blue' : 'gray'
               }
+              href={item.link}
+              key={item.link}
+              minWidth="fit-content"
             >
               {item.text}{' '}
               <Text srOnly as="span">


client/src/modules/dashboard/shared/components/Layout.tsx
```diff
@@ -54,16 +54,23 @@ export const Layout = ({
...
   return (
     <div data-cy={dataCy}>
-      <HStack {...rest} data-cy="dashboard-tabs" as="nav" my="2">
+      <HStack
+        {...rest}
+        as="nav"
+        data-cy="dashboard-tabs"
+        my="2"
+        overflowX="auto"
+      >
```
Sboonny, line 63: here is my issue with introducing overflow, if we add new link, people can miss it because they need to scroll to it first to see it.

I recommend using grid instead, because the all links will be shown and the chance of someone missing it, is lower 💭. What do you think about it.

```ts
 <Grid
        {...rest}
        marginTop='1rem'
        as="nav"
        data-cy="dashboard-tabs"
        gridTemplateColumns='repeat(auto-fit, minmax(6rem, 1fr))'
        alignContent='start'
        gap='1rem'
      >
        {linksWithPermissions
          .filter((link) => !link.requiredPermission || link.hasPermission)
          .map((item) => (
            <LinkButton
              colorScheme={
                router.pathname.startsWith(item.link) ? 'blue' : 'gray'
              }
              href={item.link}
              key={item.link}
            >
              {item.text}{' '}
              <Text srOnly as="span">
                Dashboard
              </Text>
            </LinkButton>
          ))}
      </Grid>
 ```

File: client/src/modules/dashboard/shared/components/Layout.tsx [hunk 2]
```diff
@@ -54,16 +54,23 @@ export const Layout = ({
...
           .map((item) => (
             <LinkButton
-              key={item.link}
-              href={item.link}
               colorScheme={
                 router.pathname.startsWith(item.link) ? 'blue' : 'gray'
               }
+              href={item.link}
+              key={item.link}
+              minWidth="fit-content"
```
Sboonny, line 73: if we add grid `minwidth` can go

```suggestion
```

Sboonny: Btw, I remember `key` attribute needs to have `id` value in it (Oliver linked article about it before), but this one looks different, I wonder how is this different from the usual `key` that needs `id`?

gikf: Key needs to have unique value, I think link matches that need.

Sboonny: Good point, I was tunnel visioned on `id` 😓 👍.

client/src/modules/dashboard/shared/components/Layout.tsx
gikf: I agree it might not be clear. I was going to experiment with adding some indication about scroll-ability later.

But we could change it for now to grid and see how that goes. What do you think?

Sboonny: I do like the idea of indication 🤔, if we go with scroll-ability or grid, it should make it clearer when something new is added which makes the site more friendly.

My only issue is, buttons doesn't indicate the look of scroll-ability well. We can go with the `tabs` style, if we keep the scroll-ability.

https://chakra-ui.com/docs/components/tabs/usage

gikf: So far I was thinking about adding on the sides icon buttons that would also scroll the container. Displayed only if all links wouldn't fit.

Sboonny: ya that would work 👍. 

I am resolving this.

Sboonny (approved):

Issue #1831, opened 2022-11-03
Graceful server restarts


ojeytonwilliams:
As pointed out by @Sboonny in https://github.com/freeCodeCamp/chapter/pull/1829#pullrequestreview-1167185216 the server can be crashed by changes to the client.  It's tricky to reproduce, but it does happen occasionally.

It would be nice if we could handle this in development.

Ideally, we'd want the same watch behaviour we currently have, but with not to get stuck in a loop if the server throws an error on restart.

PR #2181, opened 2023-01-06
fix: only create sessions after authenticating

### LINKED ISSUE #1997: Uncaught error when Auth0 login form timeouts

ojeytonwilliams:
- refactor: combine useLogin and useLogout
- refactor: separate dev and auth0 logic
- refactor: use a dev auth provider to mirror auth0
- fix: only create session when authenticated

This was a little more involved than I'd realised, but this reimplements the dev auth code as a provider.  The goal being to further reduce the differences between the two auth providers. This should help us in two ways: we can have more confidence that dev + auth0 behave the same way and it should make it easier to drop in new providers later down the line.

Closes https://github.com/freeCodeCamp/chapter/issues/1997 https://github.com/freeCodeCamp/chapter/pull/2120

@Jonathan1599 I started working on this because I realised there were some underlying issues that caused the dev-login to fail (your code worked nicely for Auth0, which is the most important thing, but the differences between dev and auth0 meant that dev login broke). Anyways, long story short, I incorporated your fix while I was working on this so I added you as a co-author.  I hope that's okay, since we won't be able to merge your PR if this goes in.


Jonathan1599:
Sure, no problem

ojeytonwilliams:
It's taken a while, but I think the dev/auth0 login logic is now reasonably modular. @gikf this seems like a reasonable way to organise/name things, but, as always, any thoughts you have a greatly appreciated.

gikf:
What do you think about adding to `client/src/modules/auth/providers/auth0.tsx` the same provider as for dev login (which wouldn't be then limited to dev login)? If I'm getting this right it would simplify `useAuth` to always use the `useContext(AuthContext)`

gikf:
What I'm trying to say is possibly going one step further with it. Cutting the need for the second conditional assignment here, as that would be obtained with `useContext` from the `AuthContext.Provider`
https://github.com/freeCodeCamp/chapter/blob/e5af391e34d9ab942e58e0efc25fe13775f0cd1b/client/src/modules/auth/context.tsx#L9-L10

The abstractions will not be changed. Instead of having completely separate providers - `Auth0Provider` and `DevAuthProvider`, both will pass their functions via `AuthContext.Provider`. Note that I might be using wrong terms to describe the behavior.

`auth0.tsx`
```typescript
(...)

export const Wrapper = ({ children }: { children: React.ReactNode }) => {
  const {
    isAuthenticated,
    getAccessTokenSilently: getToken,
    loginWithPopup: login,
    logout,
  } = useAuth0();

  return (
    <AuthContext.Provider value={{ getToken, isAuthenticated, login, logout }}>
      {children}
    </AuthContext.Provider>
  );
};


export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
(...)
  return (
    <Auth0Provider
      domain={domain}
      clientId={clientId}
      redirectUri={clientURL}
      audience={audience}
    >
      <Wrapper>{children}</Wrapper>
    </Auth0Provider>
  );
```

`AuthContext` is basically renamed current `DevAuthContext`
```typescript
export const AuthContext = React.createContext<AuthContextType>({
  isAuthenticated: false,
  // eslint-disable-next-line @typescript-eslint/no-empty-function
  login: () => {},
  // eslint-disable-next-line @typescript-eslint/no-empty-function
  logout: () => {},
  // eslint-disable-next-line @typescript-eslint/no-empty-function
  getToken: () => Promise.resolve(''),
});
```

And then in `useSession.ts`
```typescript
  const { getToken, login, logout, isAuthenticated } = useContext(AuthContext);
```

ojeytonwilliams:
Thanks for the detailed explanation. It still look me a long time to understand, but, now that I do, I like it!

Sboonny:
This fixed the set headers issue, that is related to https://github.com/freeCodeCamp/chapter/issues/1831 issue.

ojeytonwilliams:
> This fixed the set headers issue 

All part of the plan :eyes:

gikf (approved):
Looks great :+1: 

Unless somebody wants to look at it some more, I think it can be merged

ojeytonwilliams:
One review's good enough for me!

Issue #1269, opened 2022-07-19
SMTP discussion


ojeytonwilliams:
My two cents, I know fCC is "dogfooding" this, but we did have a long conversation about not adding on specific packages, and rather using secure SMTP for the eventual product.

It seems like a slippery slope, as I assume others will want all of the other packages and unless they can be dynamically enabled from the server, it may end up being a lot of bloat.

I'm not an expert on the JS stack, so feel free to share if/how supporting all the specific 3rd party service libraries would work vs using the SMTP protocol to relay through any mail service that uses SMTP (which I assume to be most / all).

_Originally posted by @allella in https://github.com/freeCodeCamp/chapter/issues/1262#issuecomment-1189374362_

ojeytonwilliams:
I didn't want this discussion getting buried in a merged PR.

For context on sendgrid https://github.com/freeCodeCamp/chapter/pull/1262 the main thing I was after was the ability to send emails from the staging site: https://chapter.freecodecamp.dev. Once I've deployed @naomi-lgbt's PR that should be possible and we can test in a production-like environment.

Once we've got a viable MVP, the focus needs to go back to supporting self-hosting. The key issues being email, authentication, hosting and deployment. All of these need to be cheap, configurable and easy to set up.

While the MVP will involve a lot of FCC specific implementations, having got everything live and working together once should help when we try to support self-hosting again.

allella:
I understand if it's to get to MVP and testing, but just wanted to check since it differed from the [earlier SMTP-only direction](https://github.com/freeCodeCamp/chapter/issues/283).

[Node Mailer only lists two of the popular services](https://nodemailer.com/plugins/) and we figure picking services isn't the way to go when the SMTP protocol should allow for relaying through 3rd party services or self-hosted email services.

allella:
Also, I should mention that [sendmail](https://nodemailer.com/transports/sendmail/) is a supported transport for relaying emails from a locally hosted mail daemon on the web server. While that's nice as an option, it's also the most likely way someone without a lot of sendmail / Postfix knowledge can get themselves in a jam by sending emails without proper validation, like SPF / DKIM.

My guess is most organizations will already use a 3rd party service, or their org is large enough that they have an IT staff that can provide SMTP settings for relaying and validating outbound emails.

naomi-lgbt:
Yes, the use of Sendgrid is primarily a quick fix so we can get things running on `.dev`.

Issue #283, opened 2019-12-01
Support for multiple email clients


vaibhavsingh97:
As @QuincyLarson [comment](https://github.com/freeCodeCamp/chapter/issues/47#issuecomment-543006930). we have to support popular options like Amazon SES, SendGrid, and Mandrill.

allella:
@jacobbogers mentioned in #146 we should be able to [support other transports via plugins](https://github.com/freeCodeCamp/chapter/pull/146#issuecomment-554771089).

However, SMTP is the default transport and that's likely because almost all popular 3rd party mailers offer SMTP authentication: Sendgrid, Mailgun, Postman, Mandrill.  It seems even SES offers SMTP

The [Nodemailer transport page](https://nodemailer.com/transports/) only really offers a plugin for SES and sendmail.

While sendmail could be potentially easy to setup inside a Docker container, it can easily lead to emails being sent with invalid SPF validation and no DKIM signatures.  Thus, these are far more likely to end up in spam folders.

So, the question might be if we really need to offer anything besides SMTP since the only other plugin listed is the SES plugin and sendmail may be more trouble than it's worth.

allella:
Anybody have a reason to keep this open if for the MVP we assume the `administrator` of the `instance` will use a transaction email services #147 that supports SMTP authentication?

As mentioned in the previous comment, the only [transport plugins listed on the Nodemailer plugin page](https://nodemailer.com/transports/) are sendmail, SES (AWS), a testing stream, and rolling your own transport.

I voiced SPF, DKIM, DMARC related concerns above about offering sendmail, but obviously those are not intractable if we document and ensure the `administrator` properly configures their SPF and/or DKIM in their DNS.  Though, this would also require us to install and configure a MTA, like Postfix, in the Docker `instance` image and possibly complicated DKIM + Postfix configuration. 

SES supports SMTP so the Nodemailer SES plugin seems to be more of a convenience than a requirement for using SES.  One feature you may lose here is the ability to [authenticate SES using AWS IAM roles](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/control-user-access.html#:~:targetText=You%20can%20use%20AWS%20Identity,group%2C%20or%20role%20can%20perform.&targetText=To%20use%20IAM%2C%20you%20define,the%20policy%20to%20a%20user.), but that's an advanced feature if it's even one the transport would enable.

QuincyLarson:
@allella said:

[...]

PR #2186, opened 2023-01-09
feat: switch from sendgrid to ses

naomi-lgbt:

Closes #XXXXX

This moves us over to SES for our production demo.

Some things for @ojeytonwilliams:

- [ ] What address do we want to send emails from?
- [ ] Do we want any kind of try/catch error handling on the send call?


allella:
I know fCC is going to use the mail service that makes the most sense and whatever is easiest makes sense for the dogfooding-phase. The Sendgrid-specific package was included for staging and now an AWS specific package for production.

Though, we did decide to use SMTP as the default since the protocol is likely to be supported by the vast majority of relay servers and wouldn't require including other packages unless someone adds more packages to their instance.

I suspect this doesn't prevent other future instance owners from using SMTP, but I'm more generally curious if it will be reasonably easy to allow other instances to exclude and include packages to fit their needs without bloating the package.json.

Linking this to #1269  and https://github.com/freeCodeCamp/chapter/issues/283 for future reference.

ojeytonwilliams:
>  I'm more generally curious if it will be reasonably easy to allow other instances to exclude and include packages to fit their needs without bloating the package.json.

Not easily, to be honest.  npm doesn't, to my knowledge, allow for conditional installation. That said, for now I don't think we need to worry too much about the size of the images.  They're not big enough to be causing problems.

I chatted with Naomi a little and we should be fine to keep all three providers available.  There's not a huge amount of extra code to maintain, so I think it's worth it to keep the flexibility.

Which is to say, SMTP support isn't going anywhere.

ojeytonwilliams:
> What address do we want to send emails from?

Can we use `CHAPTER_EMAIL` for all the providers?  If they can cope with "Name <email@add.ress>" that would be perfect.  

> Do we want any kind of try/catch error handling on the send call?
 
Yeah, please. If you could use https://github.com/freeCodeCamp/chapter/blob/main/server/src/util/redact-secrets.ts and log the errors, that's perfect.

```
console.error(inspect(redactSecrets(e), { depth: null }));
```

is how we've been doing it.

naomi-lgbt:
> Can we use `CHAPTER_EMAIL` for all the providers? If they can cope with `"Name [email@add.ress](mailto:email@add.ress)"` that would be perfect.

I'm not sure about the mailto formatting - I know that both Sendgrid and SES accept `Naomi <naomi@freecodecamp.org>`, though.

ojeytonwilliams (approved):
LGTM :+1: 

Thanks, @naomi-lgbt!

Issue #1970, opened 2022-11-27
disable ProfileForm's save button if the data doesn't change


Sboonny:
**Describe the bug**
It is possible to activate the save button in `userProfile`, when someone focuses on `Profile Picture` section, and move away from it without changing it.

**To Reproduce**
1. Go to /profile
2. Click on Profile picture input
3. Click away from it, without changing anything
4. See save button activate

**Screenshots**

Here is the Owner profileForm
![image](https://user-images.githubusercontent.com/88248797/204120018-bfb2175d-a094-41f8-a107-2278c4907254.png)

---

This happens because profileForm can accept empty string, this is desired because if someone wants to remove their profile picture, an empty string will remove it. But we should account for when there is no profile picture `if user.profile === "" ` and the input is `""`, that the button should stay disabled

---

To get set up locally, take a look at our [contribution docs](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md). And, if you'd like to discuss more, you can head on over to [Discord](https://discord.gg/QbQd7BpaaH) and say hi, or comment here if you prefer.

gikf:
Little context: This is because the initial default value when there's no `image_url` is `null`, but the value in form is `""`.

Sboonny:
Thanks to @bryanleemoore report, the issue seems to be solved and I couldn't replicate it, I don't know why tho.

Here is the discord conversation for context: https://discord.com/channels/692816967895220344/1049822586344833024/1049822588857221191

gikf:
`react-hook-form` was updated in the meantime, maybe there were some changes to that.

There's still slightly related issue:
- Freshly seeded database
- Go to profile
- Add text to the _Profile Picture_ - save button is enabled
- Remove text from the _Profile Picture - save button stays enabled

Sboonny:
Good point, I will check React form if there is an issue for this in there, if yes I will try to fix it.

If not on the other hand, I will check if the issue should be opened for it there, and see if they have an intent on okay with it being fixed.

if not again, I will open an issue with a link to your comment here, and we can fix it here.

PR #2190, opened 2023-01-10
feat: refactor userProfile form submit

Sboonny:

in this PR https://github.com/freeCodeCamp/chapter/pull/2112, react form added `value` that reset the states when run `useForm` is activated.

I have delayed looking into refactoring the form, because of the holiday, but looking at it today there wasn't much to do. I may be missing component and I may be misunderstanding the `value` property 🤔.
 

Ref: #1970

### DIFF
--- a/client/src/modules/chapters/chapterContext.ts
+++ b/client/src/modules/chapters/chapterContext.ts
@@ -1,0 +1,6 @@
+import { createContext } from 'react';
+import { useChapterQuery } from '../../generated/graphql';
+
+const { data } = useChapterQuery({});
+
+export const ChapterCardData = createContext(data);
--- a/client/src/modules/chapters/chapterContext.ts
+++ b/client/src/modules/chapters/chapterContext.ts
@@ -1,6 +1,0 @@
-import { createContext } from 'react';
-import { useChapterQuery } from '../../generated/graphql';
-
-const { data } = useChapterQuery({});
-
-export const ChapterCardData = createContext(data);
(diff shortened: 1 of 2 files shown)


gikf:
There's one outstanding issue with behavior of that form - when profile picture wasn't set before, field is changed and then reverted to the _blank_, save button will stay enabled (see https://github.com/freeCodeCamp/chapter/issues/1970#issuecomment-1340430458). What do you think about dealing with that here?

It can be fixed with changing
https://github.com/freeCodeCamp/chapter/blob/84d1537872da633b6ea54dcc35125c90c673b9d1/client/src/modules/profiles/component/ProfileForm.tsx#L52-L54
to 
```typescript
  const { image_url, ...rest } = data;
  const defaultValues: UpdateUserInputs = {
    image_url: image_url ?? '',
    ...rest,
  };
```

Sboonny:
This can be in the scope of this PR, good call. 👍

client/src/modules/chapters/chapterContext.ts
```diff
@@ -0,0 +1,6 @@
+import { createContext } from 'react';
+import { useChapterQuery } from '../../generated/graphql';
+
+const { data } = useChapterQuery({});
+
+export const ChapterCardData = createContext(data);
```
gikf, line 6:
```suggestion
```
Something sneaked in

gikf (approved):
LGTM :+1:

Issue #1937, opened 2022-11-21
Calendar integration and auto subscribe to event on rsvp


gikf:
When rsvping to an event, by default user is subscribed to it. When instance is authenticated to the Calendar api, this will mean user email is automatically passed to the calendar event as an attendee.

It might be worth to consider changing the default behavior and/or include explicit information that email will be used in such way.

Sboonny:
The amount of functionality linked to RSVP is confusing 😵, should we put this information on policy and the ability to opt out of this in profile? 🤔

ojeytonwilliams:
This sounds like it would be worth adding to the policy.  Adding the ability to opt out sounds reasonable, but that should be post MVP.

Issue #1670, opened 2022-10-06
Prevent duplicate notifications


ojeytonwilliams:
As mentioned in https://github.com/freeCodeCamp/chapter/issues/1639#issuecomment-1266731466 the Google calendar integration means that when a member is invited to an event by Chapter (and emailed by Chapter) they will also get invited to a Google calendar event. This could easily become irritating.

Minimum requirements

- Subscribed event members should receive at most one notification when they choose to attend an event
- Changes to the details of an event (time, venue, if it's canceled and so on) should trigger a single email

Nice to haves

- If only Chapter sends the email It should be as easy to respond 'yes' to the Google calendar as if Google sent it
- If Google sends the email it should have all the information we would put in the Chapter email

Probably impossible to have

- A member that attends a Chapter event should automatically mark that they're attending Google calendar event

I don't believe this is possible, but it would save people the hassle of clicking to indicate they want to go twice.

PR #2199, opened 2023-01-11
feat: create Google meet for new calendar events

ojeytonwilliams:

To keep things simple, this adds meets to all events.  We can improve things by only creating them for online events

### DIFF
--- a/server/src/services/Google.ts
+++ b/server/src/services/Google.ts
@@ -15,6 +15,7 @@ interface EventData {
   summary?: string;
   attendees?: calendar_v3.Schema$EventAttendee[];
   status?: 'cancelled';
+  createMeet?: boolean;
 }
 
 interface CalendarId {
@@ -30,15 +31,17 @@ interface Attendee {
   attendeeEmail: string;
 }
 
-const tokenErrors = [
+const TOKEN_ERRORS = [
   'Invalid Credentials',
   'invalid_grant',
   'No access, refresh token, API key or refresh handler callback is set.',
   'No refresh token is set.',
 ];
 
+const MEET_ID = '1';
+
 function errorHandler(err: unknown) {
-  if (err instanceof Error && tokenErrors.includes(err.message)) {
+  if (err instanceof Error && TOKEN_ERRORS.includes(err.message)) {
     console.error('Marking token as invalid');
     invalidateToken();
     sendInvalidTokenNotification();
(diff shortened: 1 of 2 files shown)


gikf:
Couple more general questions - how does the permissions for accessing the meet later work? Does everybody added to the event in the calendar automatically are allowed to enter that _meet_? What with those not subscribed to event, would they get allowed to enter when having link to the meet, or would it be needed to manually add them permission?

ojeytonwilliams:
> Does everybody added to the event in the calendar automatically are allowed to enter that meet?

Yes, exactly that.  If you're on the Calendar Event's attendee list, you can go into the meet without the meet creator needing to let you in.  That includes if you've cancelled your attendance, but not if you've been removed from the Event.

> What with those not subscribed to event, would they get allowed to enter when having link to the meet, or would it be needed to manually add them permission?

If you don't RSVP the only way in is if you're manually allowed in.  My understanding is that the creator or anyone in that creator's workspace can do that, but if they're not in the call there's nothing to be done. I'll test this properly once it's live.

gikf:
> Yes, exactly that. If you're on the Calendar Event's attendee list, you can go into the meet without the meet creator needing to let you in. That includes if you've cancelled your attendance, but not if you've been removed from the Event.

That's very nice.


> If you don't RSVP the only way in is if you're manually allowed in.

I mean RSVPed, but not subscribed to event. Or am I confused and subscription status doesn't affect being added to attendees in calendar event :thinking:

ojeytonwilliams:
> subscription status doesn't affect being added to attendees in calendar event

The fact that even you aren't sure about this means it's way too confusing! I'll see what I can do to clarify on the UI, at least.

So, subscription only affects email notifications _from Chapter_:

- If I unsubscribe from a chapter then I won't get notified about new events in that chapter.
- If I unsubscribe from an event then I won't get notified by Chapter about changes to that event 

Google operates separately. So, imagine you've RSVPed then unsubscribed from a Chapter event, but you've also accepted the invitation that Google sends. Then you'll be notified about changes to the Google Calendar Event itself, depending on how you've configured your notifications for that Google account.

There's a bit more context here: https://github.com/freeCodeCamp/chapter/issues/1670

gikf:
That what I found when looking at code. I could swear calendar event attendees list was changed when user subscribes/unsubscribes. I even made issue related to it! #1937

ojeytonwilliams:
>  I could swear calendar event attendees list was changed when user subscribes/unsubscribes

It probably did at one point! Right now it just adds them to the Calendar event if it exists.

gikf (approved):
LGTM :+1: 

I was wondering if api won't complain about always using the same `requestId` (`MEET_ID`), for multiple calendar events, but no troubles when testing it.

ojeytonwilliams:
> I was wondering if api won't complain about always using the same requestId (MEET_ID), for multiple calendar events, but no troubles when testing it.

Yeah, it's a bit odd, but I think it just needs to be unique for that event.  So, if you want to create multiple 'conferences' you need multiple ids, but otherwise it doesn't seem to matter.

PR #2212, opened 2023-01-13
fix: prevent double next calls in user middleware

ojeytonwilliams:

https://github.com/freeCodeCamp/chapter/pull/2181 added a bug in the client which exposed a bug in the server.

The client bug is that the client asks the server to destroy the session on page load.  I'll address this in a followup PR.

The server bug is that it's possible for the session to exist when the middleware is called, but be destroyed during the call. In this case, the middleware would see that no user existed, call next, but continue processing and call next again.

The combination of the two means that it's quite likely that the server simply crashes on page load.


gikf (approved):
LGTM :+1:

PR #2226, opened 2023-01-16
fix: consider streaming_url changed when venue_type changes

gikf:

---
- Considers `streaming_url` as changed when `venue_type` is changed. Otherwise the existing streaming url wouldn't be included in the email informing about new venue details. Ie. when event was online-only and now is changed to physical & online, email about changes would include venue location, but streaming url would get lost, making it appear as physical only event going forward.


Sboonny (approved):

PR #2192, opened 2023-01-10
feat: add delete button for venues

Sboonny:
- [x] I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md).
- [x] My pull request has a descriptive title (not a vague title like `Update README.md`).
- [x] My pull request targets the `main` branch of Chapter.

Ref #1558

I have added a simple button that delete the `id`, and I like the suggestions on the PR, I wish I have seen it sooner, but the app is going to be published and we need delete venue button.

For now, we can have a simple delete `id`, and add more to the function, when we create a specific page for a venue. In it, we can have this function with checkbox that if selected, they will delete `events` or an `event` with the `venues`.

PR #2233, opened 2023-01-17
feat: move events to TBD venue when deleting venue

Sboonny:


gikf:
Streaming url is fine to stay as it is.

I thought of something like:
```typescript
    await prisma.events.updateMany({
      where: { venue_id: id },
      data: { venue: { disconnect: true } },
    });
```
I haven't checked how correct that is. Then I found https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#disconnect-all-related-records which would look like:
```typescript
    await prisma.venues.update({
      where: { venue_id: id },
      data: { events: { set: [] } },
    })
```
That's just my assumptions, not sure if either of these will work.

Sboonny:
If my guess is correct, the both approach will set the `venue_id` to `null` (I may be missing something for sure), although everything may function with venue_id, after setting up the events. I am worried about creating two similar state of data, one with `venue_id` equals to `0`, while the other with `venue_id` is `null`, but doesn't show a `TBD` text.

I guess all of this is a question, are we okay with swapping the content of the deleted venue with the `TBD` venue? or should we have two different states?

gikf:
`venue` with id `0` doesn't exist, it's technically event with not connected any venue, and venue type having physical venue.

gikf:
Sorry, I completely forgot about one more related thing - informing attendees of the events having deleted venue about new (undecided) venue.

Sboonny:
There isn't an existing link for venuesUsers 😞, or is there? 

I need to get them from chapter, but not include the one with canceled RSVP, and past Events. 🤔

Sboonny:
Waiting for #2238 for QoL.

Btw, about email contact, if we want it.

Should we go for something like:

```ts
if(list_of_upcoming_events.length > 0) send email
```

> Subject: Chapter won't host the venue anymore
>
> ------
>
> The events related to venue won't be host locally anymore. You can still attend the event in {Streaming_URL}.
> {list_of_upcoming_events}
> If you would like to cancel your RSVP, you can do so in {unsubscribe_event_link}.
>
> {unsubscribe_chapter_link}

server/src/controllers/Venue/resolver.ts
```diff
@@ -99,9 +101,42 @@ export class VenueResolver {
...
+              event: {
+                canceled: false,
+                ends_at: { gt: new Date() },
+                venue_type: isPhysical,
+              },
+            },
+          },
+        },
+      },
+    });
```
ojeytonwilliams, line 127: Is the object to find the users who are going to events whose venues have changed? If so, `user_chapters` can be ignored.

Sboonny: Good point, we don't have venues linked to multiple chapters, so we don't need to specify the `user_chapters`. 👍

ojeytonwilliams: Right. If they're not going to an event, they don't need to know the venue has changed :man_shrugging:

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,48 @@ export class VenueResolver {
...
+          every: {
+            event: {
+              canceled: false,
+              ends_at: { gt: new Date() },
+              venue_type: events_venue_type_enum.Physical,
+            },
+          },
+        },
+      },
+    });
```
gikf, line 119: This seem to find much more than it's needed. `venue_type` don't really matter when we can just look for the `venue_id`. Also, the `every` would require every `user_events` to match the conditions, to find specific user, while `some` needs just one.
```suggestion
    const users = await prisma.users.findMany({
      where: {
        user_events: {
          some: {
            event: {
              canceled: false,
              ends_at: { gt: new Date() },
              venue_id: id,
            },
            subscribed: true,
          },
        },
      },
    });
```

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,48 @@ export class VenueResolver {
...
+
+    const venue = await prisma.venues.findUniqueOrThrow({
+      where: { id: id },
+    });
+    const chapterId = venue.chapter_id;
+
+    const unsubscribeOptions = (userId: number) =>
+      chapterUnsubscribeOptions({
```
gikf, line 127: Wouldn't the `eventUnsubscribeOptions` be better here?

Sboonny: Let's go, I am not the only one who feels that way 🤣. https://github.com/freeCodeCamp/chapter/issues/2262

gikf: Well yes and no. I initially made suggestion to search for the affected users via the `event_users`, but then I realized single user can be in multiple events having this venue what would send them multiple emails stating _Events hosted at ${venue.name}_.

Now when I think about it, maybe it would be better to send one email per event. Or still send one email per user, but specifically include in the email all events affected, to which user is RSVPed.

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,86 @@ export class VenueResolver {
...
+    </tbody>
+</table>`;
+
+    for (const { email, id: currentUserId } of users) {
+      const events = await prisma.events.findMany({
+        where: {
+          canceled: false,
+          ends_at: { gt: new Date() },
+          venue_id: id,
+          event_users: { some: { user_id: currentUserId, subscribed: true } },
```
Sboonny, line 173:
```suggestion
          event_users: { every: { user_id: currentUserId, subscribed: true } },
```

server/src/controllers/Venue/resolver.ts
ojeytonwilliams: > Or still send one email per user, but specifically include in the email all events affected, to which user is RSVPed.

This, please. We should take any opportunity to keep the number of emails low.

Sboonny: oh, I thought this commit sorted this https://github.com/freeCodeCamp/chapter/pull/2233/commits/f0dbf3b454613c2ee7963c2c2596dd986ce52032

Did I misunderstand something?

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,86 @@ export class VenueResolver {
...
+            event: {
+              canceled: false,
+              ends_at: { gt: new Date() },
+              venue_id: id,
+            },
+            subscribed: true,
+          },
+        },
+      },
+    });
```
gikf, line 120:
```suggestion
    const users = await prisma.users.findMany({
      where: {
        user_events: {
          some: {
            event: {
              canceled: false,
              ends_at: { gt: new Date() },
              venue_id: id,
            },
            subscribed: true,
            rsvp: { name: 'yes' },
          },
        },
      },
      include: {
        user_events: {
          where: {
            event: {
              canceled: false,
              ends_at: { gt: new Date() },
              venue_id: id,
            },
            subscribed: true,
            rsvp: { name: 'yes' },
          },
        },
      },
    });
```
:eyes:
- Filter users having at least one `user_events` with:
  - subscribed: true
  - rsvp: yes
  - Event having:
    - canceled: false
    - ends_at: { gt: new Date() }
    - venue_id: id
- For filtered users, include all `user_events` with:
  - subscribed: true
  - rsvp: yes
  - Event having:
    - canceled: false
    - ends_at: { gt: new Date() }
    - venue_id: id

Sorry! :eyes:

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,100 @@ export class VenueResolver {
...
+    </tbody>
+</table>`;
+
+    for (const { email, id: currentUserId } of users) {
+      const events = await prisma.events.findMany({
+        where: {
+          canceled: false,
+          ends_at: { gt: new Date() },
+          venue_id: id,
+          event_users: { every: { user_id: currentUserId, subscribed: true } },
```
Sboonny, line 187: Random question, do we need `subscribed` and `rsvp` here? We have already found user_id, so it should select the same user, and we won't need to filter the user again

```suggestion
          event_users: { every: { user_id: currentUserId } },
```

gikf: Good news is additional queries are not needed, because `event` has relation with `user_event`, so every `user_event` can already link to the affected event. Although one more line is needed to include the actual event in the other query :eyes:

```typescript
    const users = await prisma.users.findMany({
      where: {
        user_events: {
          some: {
            event: {
              canceled: false,
              ends_at: { gt: new Date() },
              venue_id: id,
            },
            subscribed: true,
            rsvp: { name: 'yes' },
          },
        },
      },
      include: {
        user_events: {
          include: { event: true },
          where: {
            event: {
              canceled: false,
              ends_at: { gt: new Date() },
              venue_id: id,
            },
            subscribed: true,
            rsvp: { name: 'yes' },
          },
        },
      },
    });
```

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,91 @@ export class VenueResolver {
...
+        eventId: currentEventId,
+        userId: currentUserId,
+      });
+
+    const eventList = (event: events[], currentUserId: number) => {
+      event.map(
+        ({ name, id }) => `
+      <tr>
+      <td>${name}</td>
+      <td>${unsubscribeOptions(id, currentUserId)}</td>
```
Sboonny, line 155:
```suggestion
    const eventList = (event: events[], currentUserId: number) => {
      event.map(
        ({ name, id }) => `
      <tr>
      <td>${name}</td>
      <td>${eventUnsubscribeOptions(id, currentUserId)}</td>
```

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,85 @@ export class VenueResolver {
...
+      <tr>
+      <td>${name}</td>
+      <td>${eventUnsubscribeOptions({
+        eventId: id,
+        userId: currentUserId,
+      })}</td>
+     </tr>
+  `,
+      );
+    };
```
gikf, line 153:
```suggestion
    const eventList = (event: events[], currentUserId: number) =>
      event.map(
        ({ name, id }) => `
      <tr>
        <td>${name}</td>
        <td>${eventUnsubscribeOptions({
          eventId: id,
          userId: currentUserId,
        })}</td>
      </tr>
      `,
      );
```
Resulting array wasn't returned.

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,85 @@ export class VenueResolver {
...
+    const emailSubject = `Events hosted at ${venue.name} won't be hosted there anymore`;
+    const emailContent = (
+      event: events[],
+      currentUserId: number,
+    ) => `The events related to ${
+      venue.name
+    } won't be host locally anymore.<br />
+    <table>
```
gikf, line 161:
```suggestion
  <table>
```

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,85 @@ export class VenueResolver {
...
+    <thead>
+        <tr>
+          <th>Event</th>
+          <th>Cancel RSVP</th>
+        </tr>
+    </thead>
+    <tbody>
+       ${eventList(event, currentUserId)}
+    </tbody>
+</table>`;
```
gikf, line 171:
```suggestion
  </table>`;
```

server/src/controllers/Venue/resolver.ts
```diff
@@ -102,6 +104,85 @@ export class VenueResolver {
...
+    } won't be host locally anymore.<br />
+    <table>
+    <thead>
+        <tr>
+          <th>Event</th>
+          <th>Cancel RSVP</th>
+        </tr>
+    </thead>
+    <tbody>
+       ${eventList(event, currentUserId)}
```
gikf, line 169:
```suggestion
      ${eventList(event, currentUserId).join('')}
```
By default array will be in output with commas as separators.

server/src/controllers/Venue/resolver.ts
Sboonny: Do we add the eslint rule, that warns about missing `return` in the functions 🤔?

Btw, how did you test it? I am asking because, I can't think of a way to trigger the resolver to test it, that isn't creating a button and requesting the resolver to trigger. https://github.com/freeCodeCamp/chapter/pull/2192

server/src/controllers/Venue/resolver.ts
Sboonny: Nice catches, thank you 💚

gikf (approved):

PR #2240, opened 2023-01-18
fix: allow NODE_ENV to be set to production

ojeytonwilliams:
NextJS expects this to be the case when building the site, but we need to make sure that the dev dependencies are installed.  The site cannot be built without them. Hence the updates to the Dockerfile.

Closes #XXXXX


gikf (approved):
:+1: Thanks

PR #2247, opened 2023-01-19
refactor: move email providers into own classes

ojeytonwilliams:

The MailerService had become a little complicated in the process of supporting 3 different email providers (for want of a better term). This should be easier to maintain.

The code won't win any prizes for OO brilliance, but I think it's a step in the right direction.


gikf (approved):
Looks good :+1:

PR #2248, opened 2023-01-19
fix: logo image being in the center of chaptersPage

Sboonny:
Co-authored-by: gikf <60067306+gikf@users.noreply.github.com>

Originally reported when added the logos, I thought that the cause was `Center` element, but after testing and more digging, that isn't true, and I should have investigated more instead of searching for second solution.

image


gikf (approved):
LGTM :+1:

PR #2252, opened 2023-01-19
refactor: userProfile loading

gikf:

---
- Refactors a bit to use the `Loading` component.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -5731,9 +5731,13 @@
             "description": null,
             "args": [],
             "type": {
-              "kind": "OBJECT",
-              "name": "UserProfile",
-              "ofType": null
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "OBJECT",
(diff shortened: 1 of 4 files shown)


Sboonny (approved):

Sboonny:
I have tried to mess with it, and see why when user log in, the page doesn't change

image

tested

```ts
  if (!isLoggedIn)
    return <NextError statusCode={401} title={'Log in to see this page'} />;
  if (isLoading || !data?.userProfile) return <Loading loading={isLoading} />;
  ```
  
And got empty screen. 😞
  
  
image

gikf:
Yeah, that didn't change. I was hoping to not touch it yet :stuck_out_tongue:, but since you are bringing it up I tried to find the solution. I'm gonna commit it in a bit.

gikf:
I've figured out two ways, one with manual refetch after the `isLoggedIn` is changed. The second is to not load profile component unless user is logged in, that's similar to the `DashboadLayout`. I've had some fun with the second one.

PR #2254, opened 2023-01-20
fix: simplify data fetching

ojeytonwilliams:
Rather than handling null results when the user is required and not
logged in, it's simpler (and more appropriate) to throw errors. Ideally
before querying the db, since we know that query will fail.

The rationale is that these are bad requests and should be treated as
errors. This benefit is that it simplifies the client logic by removing
some null checking.

~~I'll address me() in a separate commit.~~

The meQuery is a special case. Until the client makes a request to the
api it cannot know if it is logged in or not (since the session cookie
determines this and it is httponly). Treating unauthenticated requests
as errors seems wrong, in that case.

In all other cases the client can first check if the user is logged in
before initiating the request.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -5202,13 +5202,9 @@
               }
             ],
             "type": {
-              "kind": "NON_NULL",
-              "name": null,
-              "ofType": {
-                "kind": "OBJECT",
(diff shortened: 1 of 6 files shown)


Sboonny (approved):

PR #2257, opened 2023-01-21
fix: limit unhelpful errors

### LINKED ISSUE #2215: Unsubscribing via email links can fail with an unhelpful message
When a new event is created, Chapter emails the subscribers to let them know about it.  These emails contain two unsubscribe links; one for the chapter (to stop these kinds of emails) and one for the event (to stop notifications about that event specifically).  However, until the user RSVPs the event unsubscribe link is both unnecessary and buggy. The user is shown "Error: No event_users found" and nothing happens.

We should

- [ ] Only send links for things the user has subscribed to (in this case the Chapter that the event belongs to)
- [ ] Handle the case where a unsubscribes more than once (e.g. by following the link twice or unsubscribing via the chapter/event page and then following the link)

gikf:

Closes #2215

---
- Catches `RECORD_MISSING`. Whatever the reason there's no such subscription, technically that's being unsubscribed. It happens only if:
  - `event_users` entry doesn't exist - user never rsvped to the event.
  - `chapter_users` entry doesn't exist - user is not a part of chapter (anymore or yet).
- Event invite has only chapter unsubscribe option.
- Some shuffling, renaming and moving around the unsubscribe options functions.

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -54,8 +54,9 @@ import { createCalendarEventHelper } from '../../util/calendar';
 import { updateWaitlistForUserRemoval } from '../../util/waitlist';
 import { redactSecrets } from '../../util/redact-secrets';
 import {
-  getChapterUnsubscribeOptions,
-  getEventUnsubscribeOptions,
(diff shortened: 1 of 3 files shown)


Sboonny (approved):

PR #2263, opened 2023-01-23
fix: only require chrome users to logout once

ojeytonwilliams:
- feat: stop trying to regenerate session

It's only useful in the weird edge case where a user has deleted their
session cookie, but not their auth0 cookies. In this case, they can just
login again.
- fix: only require single logout request on Chrome

Some browsers (such as Brave) delete third party cookies after when the
page unloads (i.e. refresh or tab closing). Chrome, however, does not
and this means that isAuthenticated is true when useSession checks to
see if it should create or destroy the session.

By specifying the intent, we can ensure the session is deleted on logout

### DIFF
--- a/client/src/hooks/useSession.ts
+++ b/client/src/hooks/useSession.ts
@@ -60,5 +60,5 @@ export const useSession = () => {
     }
   }, [isAuthenticated, canAlterSession]);
 
-  return { createSession, login, logout, isAuthenticated };
+  return { login, logout, isAuthenticated };
 };
--- a/client/src/hooks/useSession.ts
+++ b/client/src/hooks/useSession.ts
@@ -37,28 +37,33 @@ export const useSession = () => {
   // Until the user initiates a login or logout, we don't want to create or
   // destroy a session
   const [canAlterSession, setCanAlterSession] = useState(false);
+  // isAuthenticated does not change immediately after login/logout, so we
+  // need to track the intent to prevent accidental session creation/deletion
+  const [toLogout, setToLogout] = useState(false);
 
   const login = async () => {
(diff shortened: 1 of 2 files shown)


client/src/hooks/useSession.ts
```diff
@@ -37,28 +37,33 @@ export const useSession = () => {
...
   // destroy a session
   const [canAlterSession, setCanAlterSession] = useState(false);
+  // isAuthenticated does not change immediately after login/logout, so we
+  // need to track the intent to prevent accidental session creation/deletion
+  const [toLogout, setToLogout] = useState(false);
 
   const login = async () => {
     await loginAuth();
     setCanAlterSession(true);
+    setToLogout(false);
```
gikf, line 47: Here's something I don't understand. Is it safe to expect, the `useEffect` callback will not run after the `setCanAlterSession`, but before `setToLogout`?

It doesn't seem to be doing that, but if it would, then the wrong action on session would be run.

ojeytonwilliams: Good question. I think it's fine and I'll try to explain why.

How I understand the interaction between state, effect and renders is that `useEffect` runs after each render and `setState` signals to React that it may need to render.  However, it doesn't immediately trigger a render, it lets the rest of the component logic resolve first.  That means that both `setCanAlterSession` and `setToLogout` will resolve and alter their respective states. At that point the page can render and only then will the `useEffect` fire.

gikf: That makes sense, I wasn't sure if that's the expected behavior or just being lucky when testing it. Thanks :rocket

gikf (approved):

PR #2268, opened 2023-01-24
fix: make users-view just an instance permission again

ojeytonwilliams:
- feat: make users-view an instance permission again
- refactor: split checkPermission into two functions

Chapter administrators cannot currently view instance users - they have the chapter permission, but the query resolver ignores `chapterId`, so chapter permissions cannot be used. Nonetheless, it _looks_ like they can and that's confusing. This is addressed by removing it from the chapter permissions (it's already an instance permission).

The bulk of the code is the refactor that uncovered this issue: by creating `checkInstancePermission` and `checkChapterPermission` it makes it obvious how these should be used.

### DIFF
--- a/server/prisma/migrations/20230124093945_remove_users_view_from_chapter_permissions/migration.sql
+++ b/server/prisma/migrations/20230124093945_remove_users_view_from_chapter_permissions/migration.sql
@@ -1,0 +1,18 @@
+-- Reverts
+-- server/prisma/migrations/20221123213911_add_UsersView_to_chapter_permissions/migration.sql
+-- Rationale: only instance owners should be able to view instance users.
+-- Chapter administrators can already view chapter users via the chapter-edit
+-- permission.
+DELETE FROM
+    chapter_role_permissions
+WHERE
+    chapter_permissions_id = (
+        SELECT
+            id
+        FROM
+            chapter_permissions
+        WHERE
+            name = 'users-view'
+    );
+    
+DELETE FROM chapter_permissions WHERE name = 'users-view';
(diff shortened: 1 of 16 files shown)


server/prisma/migrations/20230124093945_remove_users_view_from_chapter_permissions/migration.sql
```diff
@@ -0,0 +1,18 @@
...
+    chapter_permissions_id = (
+        SELECT
+            id
+        FROM
+            chapter_permissions
+        WHERE
+            name = 'users-view'
+    );
+    
+DELETE FROM chapter_permissions WHERE name = 'users-view';
```
Sboonny, line 18: random question, why do we need to add new file, instead of updating the old one? 💭 

https://github.com/freeCodeCamp/chapter/blob/34e5b39e12d3864d484120044ef23428b72afd08/server/prisma/migrations/20221123213911_add_UsersView_to_chapter_permissions/migration.sql

ojeytonwilliams: If we just deleted that file it would be fine in development, but would mean that I'd need to apply it manually to the production db.  It's possible, but not recommended, since you need to apply this SQL manually and update the table that prisma uses to track migrations. Doable, but a chore.

This isn't too much of a hassle since it's just us, but it's harder than simply creating a new migration and it wouldn't be possible if multiple people were running instances of Chapter. Best to get into good habits now.

Also, keeping all the migrations makes it easy to find out why a change was made in the first place. When making this PR I went to https://github.com/freeCodeCamp/chapter/pull/1891 to see why this was done and I could do that easily because it was part of the commit history for the migration you linked.

Sboonny (approved):

ojeytonwilliams:
@gikf do you mind taking a quick look?  This is more your code than mine.

gikf (approved):
Looks good to me :+1:

Issue #2168, opened 2023-01-03
Handling deleted calendars and calendar events


gikf:
### Discuss your ideas or share your views: 
After deleting calendar or calendar event in Google calendar, the events and chapters linked to those will be left in limbo.

In these cases calendar api should return 404 error (https://developers.google.com/calendar/api/guides/errors#404_not_found), so it should be easy to distinguish when that happens and remove `calendar_id` or `calendar_event_id`. Maybe also send email to the instance owner.

Calendar portion is partially related to #1896

Issue #1896, opened 2022-11-14
Create new calendar if a Chapter loses access


ojeytonwilliams:
### Discuss your ideas or share your views: 

Right now we can create Google calendar events if a Chapter event does not have one.  However, it's not possible to create new calendars at all.

In the event that the Chapter app is unable to modify the a calendar (presumably there's a request it can do to check this), it would be nice to be able to create a new one.

P.S. this is very much not MVP worthy, but I want to give people a chance to see this before I leave this to the roadmap.

gikf:
I was going to add creating calendar in chapter dashboard, in similar manner to creating calendar events from event dashboard. It got somewhat delayed due to my not being sure how to approach the _checking if Google calendar is authorized and if integration is working_.

ojeytonwilliams:
Yeah, it's not completely obvious and it's not a *huge* priority.  I only ran into it because I manually updated the staging db to fix an issue, so I don't see it happening much.

Having said all that, my vague thinking was that we could show the button in one of two situtations

- When we catch errors while trying to create an event
- After making a deliberate API call to check if we have access 

I don't know which is better, but I slightly prefer the first approach as long as we're sure that the error is caused by a lack of permission.

gikf:
Closing as completed. Calendar unlinking is possible after making explicit test from chapter dashboard. Applies to deleted calendars and when access to it is lost.

PR #2277, opened 2023-01-25
feat: unlink chapter calendar

gikf:

Ref #1896
Ref #2168

---

- Adds button for unlinking calendar from chapter dashboard.
  - Calendar id is removed from chapter.
  - Calendar event ids are removed from events in the chapter.
- Currently button is displayed always when calendar is linked. Related issue suggests displaying it under some conditions. I'm not sure if that's needed, if we know that access to calendar is lost, need to confirm it explicitly would be unnecessary.
- What bothers me a bit is refetching all `DASHBOARD_EVENTS`, because of unlinking events.

### DIFF
--- a/client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
@@ -3,11 +3,14 @@ import {
   Heading,
   HStack,
   Grid,
+  List,
+  ListItem,
+  ListIcon,
   Spinner,
   Text,
   useToast,
 } from '@chakra-ui/react';
-import { CheckIcon, CloseIcon } from '@chakra-ui/icons';
+import { CheckIcon, CloseIcon, InfoIcon } from '@chakra-ui/icons';
 import NextError from 'next/error';
 import React, { ReactElement, useMemo } from 'react';
 
@@ -21,6 +24,8 @@ import {
   useCalendarIntegrationStatusQuery,
   useCreateChapterCalendarMutation,
   useDashboardChapterQuery,
+  DashboardChapterQuery,
+  useUnlinkChapterCalendarMutation,
 } from '../../../../generated/graphql';
 import { useParam } from '../../../../hooks/useParam';
 import styles from '../../../../styles/Page.module.css';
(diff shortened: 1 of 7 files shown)


client/src/modules/dashboard/Chapters/pages/ChapterPage.tsx
```diff
@@ -133,7 +188,19 @@ export const ChapterPage: NextPageWithLayout = () => {
               {loadingCalendar ? (
                 <Spinner size="sm" />
               ) : data.dashboardChapter.calendar_id ? (
-                <CheckIcon boxSize="5" />
+                <>
+                  <CheckIcon boxSize="5" />
+                  {checkPermission(user, Permission.ChapterEdit, {
```
ojeytonwilliams, line 193:
```suggestion
                  {checkPermission(user, Permission.ChapterCreate, {
```
This is fairly serious and (hopefully) not something people will often need, so I think it should be restricted to instance owners.

server/src/controllers/Chapter/resolver.ts
```diff
@@ -201,6 +201,22 @@ export class ChapterResolver {
     }
   }
 
+  @Authorized(Permission.ChapterEdit)
```
ojeytonwilliams, line 204:
```suggestion
  @Authorized(Permission.ChapterCreate)
```

ojeytonwilliams (changes requested):
Since accidentally triggering this would be incredibly annoying, I think we need some fairly heavy guardrails.  For instance, if chapter still has access to the calendar, it shouldn't be possible to unlink it.

So, it should try to get the calendar first and, if it can, just tell the user that nothing needs to be done.

gikf:
Hm, so maybe making explicit test first, before giving ability to unlink, would be better after all. It would be strange showing the _Unlink_ button, asking if unlink it, and then go: _yeah, no_

ojeytonwilliams:
I know what you mean, but it doesn't seem right to spam the Google api just because a user may need to see a button.

How about handling that with explanatory text?

> If the Google Calendar for this chapter no-longer exists, you can use the Unlink button to stop Chapter from trying to connect to it. Note: this will do nothing if the Calendar exists and Chapter has access.

It's a bit wordy, but this is an edge case that (hopefully) won't come up much.

ojeytonwilliams:
Ah, that makes sense. Thanks for clarifying!

> Test -> Ask to confirm test, with some explanation -> If test fails display Unlink button, or directly use new modal to ask about unlinking.

That sounds perfect.

ojeytonwilliams (approved):
Cheers, @gikf 

LGTM :+1:

PR #2285, opened 2023-01-26
feat: filter recent dashboard event

Sboonny:
Co-authored-by: gikf <60067306+gikf@users.noreply.github.com>

Filter recent events on the dashboard, and sorted some styling. It may seem too big of a change, but I haven't done much except make the component compatible with multiple props, and used grid.

The affected page is 

> /dashboard/events


client/src/modules/dashboard/Events/pages/EventsPage.tsx
```diff
@@ -21,29 +22,37 @@ import { useUser } from '../../../auth/user';
...
   return (
-    <Flex>
-      <FormLabel htmlFor="show-canceled-events">Show canceled events</FormLabel>
+    <Flex alignItems="center" justifyContent="space-between">
+      <FormLabel marginTop=".5em" htmlFor="show-canceled-events">
+        {filterLabel}
+      </FormLabel>
       <Switch
         isChecked={defaultChecked}
         id="show-canceled-events"
```
gikf, line 43: `htmlFor` and `id` are the same for all components.

client/src/modules/dashboard/Chapters/components/DeleteChapterButton.tsx
```diff
@@ -36,8 +36,14 @@ export const DeleteChapterButton = ({
...
-      { query: DASHBOARD_EVENTS, variables: { showCanceled: true } },
-      { query: DASHBOARD_EVENTS, variables: { showCanceled: false } },
+      {
+        query: DASHBOARD_EVENTS,
+        variables: { showCanceled: true, showRecent: true },
+      },
+      {
+        query: DASHBOARD_EVENTS,
+        variables: { showCanceled: false, showRecent: false },
+      },
```
gikf, line 46: Event more I think it might be better to do filtering only in client... The `{ showCanceled: true, showRecent: false }` and `{ showCanceled: false, showRecent: true }` combinations won't be refetched without adding them explicitly. What starts to get a bit crazy.

gikf (commented):
There's couple specific suggestions and a more general ones.

As much as grid helps, buttons sometimes look huge.
image

---
The _Show recent events_ might be a bit vague. Recent as only upcoming events?

client/src/modules/dashboard/Chapters/components/DeleteChapterButton.tsx
Sboonny: TD;LR, I agree we should sort this in the client.

I have created useFilter hook in this PR https://github.com/freeCodeCamp/chapter/pull/2301, that we can use to filter the Events page, it won't change much, will need to add `boolean` action in this PR, and it should result in the same functionality in the server.

When it's merged, I will finish this 👍

client/src/modules/dashboard/Events/pages/EventsPage.tsx
```diff
@@ -21,70 +22,120 @@ import { useUser } from '../../../auth/user';
...
+      </FormLabel>
       <Switch
         isChecked={defaultChecked}
-        id="show-canceled-events"
-        onChange={(e) => setShowCanceled(e.target.checked)}
+        id={id}
+        onChange={(e) => setFilterEvent(e.target.checked)}
       />
-    </Flex>
+    </>
```
Sboonny, line 48: The `Switch` component is weird, it has `label` built in it, but it doesn't have context inside of it, so currently we have two `label` elements.


We can change it to this, but I don't think this is accessible, because the context come after the `input`. here is the sourse code 

https://github.com/chakra-ui/chakra-ui/blob/3e1b3f6b6a7398b71ac08339110f075695fbae94/packages/components/switch/src/switch.tsx#L82-L109

```ts
      <Switch
        marginTop='.25rem'
        htmlFor={id}
        isChecked={defaultChecked}
        id={id}
        onChange={(e) => setFilterEvent(e.target.checked)}
        fontWeight='500'
      >
        {filterLabel}
      </Switch>
```

Sboonny: image

client/src/modules/dashboard/Chapters/components/DeleteChapterButton.tsx
```diff
@@ -36,8 +36,9 @@ export const DeleteChapterButton = ({
     refetchQueries: [
       { query: CHAPTERS },
       { query: DASHBOARD_CHAPTERS },
-      { query: DASHBOARD_EVENTS, variables: { showCanceled: true } },
-      { query: DASHBOARD_EVENTS, variables: { showCanceled: false } },
+      {
+        query: DASHBOARD_EVENTS,
+      },
```
Sboonny, line 41:
```suggestion
      { query: DASHBOARD_EVENTS },
```

client/src/modules/dashboard/Events/components/Actions.tsx
```diff
@@ -47,8 +47,9 @@ const Actions: React.FC<ActionsProps> = ({
         { query: CHAPTER, variables: { chapterId: chapter.id } },
         { query: EVENT, variables: { eventId: event.id } },
         { query: DASHBOARD_EVENT, variables: { eventId: event.id } },
-        { query: DASHBOARD_EVENTS, variables: { showCanceled: true } },
-        { query: DASHBOARD_EVENTS, variables: { showCanceled: false } },
+        {
+          query: DASHBOARD_EVENTS,
+        },
```
Sboonny, line 52:
```suggestion
      { query: DASHBOARD_EVENTS },
```

client/src/modules/dashboard/Events/components/EventCancelButton.tsx
```diff
@@ -29,8 +29,9 @@ const EventCancelButton = (props: EventCancelButtonProps) => {
     refetchQueries: [
       { query: EVENT, variables: { eventId: event.id } },
       { query: DASHBOARD_EVENT, variables: { eventId: event.id } },
-      { query: DASHBOARD_EVENTS, variables: { showCanceled: true } },
-      { query: DASHBOARD_EVENTS, variables: { showCanceled: false } },
+      {
+        query: DASHBOARD_EVENTS,
+      },
```
Sboonny, line 34:
```suggestion
      { query: DASHBOARD_EVENTS },
```

client/src/modules/dashboard/Events/pages/EditEventPage.tsx
```diff
@@ -31,8 +31,9 @@ export const EditEventPage: NextPageWithLayout = () => {
   // https://www.apollographql.com/docs/react/data/mutations/#updating-the-cache-directly
   const [updateEvent] = useUpdateEventMutation({
     refetchQueries: [
-      { query: DASHBOARD_EVENTS, variables: { showCanceled: true } },
-      { query: DASHBOARD_EVENTS, variables: { showCanceled: false } },
+      {
+        query: DASHBOARD_EVENTS,
+      },
```
Sboonny, line 36:
```suggestion
      { query: DASHBOARD_EVENTS },
```

cypress/fixtures/events.json
```diff
@@ -20,7 +20,7 @@
     "image_url": "https://test.event.org/image",
     "streaming_url": "https://test.event.org/video",
     "start_at": "2022-01-01T00:01",
-    "ends_at": "2022-01-02T00:02"
+    "ends_at": "2099-12-31T00:02"
   },
   "eventThree": {
```
Sboonny, line 25: Maybe we can turn `eventTwo` to `runningEvent`, and `eventThree` to `upcomingEvent`, to make it easier in the future

gikf (commented):
I'm not sure if filtering by default is the way to go. Personally I'd keep the full list on load.

cypress/fixtures/events.json
```diff
@@ -20,7 +20,7 @@
     "image_url": "https://test.event.org/image",
     "streaming_url": "https://test.event.org/video",
     "start_at": "2022-01-01T00:01",
-    "ends_at": "2022-01-02T00:02"
+    "ends_at": "2099-12-31T00:02"
```
gikf, line 23: Could you revert this date change and the two below?

Sboonny: the test will always fail, because they will be filtered 🤔.

gikf (approved):

PR #2292, opened 2023-01-27
feat: integration info during event creation

gikf:

---

- Adds information about integration status and creation of the event in the chapter's calendar.
- Changing query to `dashboardChapter` was needed to get the `has_calendar`.

### DIFF
--- a/client/src/modules/dashboard/Events/components/EventForm.tsx
+++ b/client/src/modules/dashboard/Events/components/EventForm.tsx
@@ -30,52 +30,53 @@ import EventSponsorsForm from './EventSponsorsForm';
 import EventVenueForm from './EventVenueForm';
 import { EventFormProps, fields, EventFormData } from './EventFormUtils';
 
+interface IntegrationInfoData {
+  isAuthenticated: boolean | null | undefined;
+  hasCalendar: boolean;
+}
+
 const IntegrationInfo = ({
   isAuthenticated,
   hasCalendar,
-}: {
-  isAuthenticated: boolean | null | undefined;
-  hasCalendar: boolean;
-}) => {
-  const isBroken = isAuthenticated === null;
+}: IntegrationInfoData) => {
+  const infoData = ({ isAuthenticated, hasCalendar }: IntegrationInfoData) => {
+    if (isAuthenticated) {
+      if (hasCalendar) {
+        return {
+          Icon: InfoIcon,
+          text: 'Instance is authenticated with calendar api, and chapter has created calendar. Chapter will attempt creating event in the calendar.',
+        };
+      }
+      return {
+        Icon: WarningTwoIcon,
+        text: "Instance is authenticated with calendar api, but chapter doesn't have calendar created. Event will not be created in the calendar.",
+      };
+    }
+
+    const isBroken = isAuthenticated === null;
+    if (isBroken) {
+      if (hasCalendar) {
+        return {
+          Icon: WarningTwoIcon,
+          text: 'Chapter has calendar created, but calendar integration is not working. Event will not be created in the calendar.',
+        };
+      }
+      return {
+        Icon: WarningTwoIcon,
+        text: 'Calendar integration is not working, and chapter does not have calendar created. Event will not be created in calendar.',
+      };
+    }
+    return {
+      Icon: WarningTwoIcon,
+      text: 'Instance is not authenticated with calendar api. Automatic creation of events in calendar is not possible.',
+    };
+  };
+
+  const { Icon, text } = infoData({ isAuthenticated, hasCalendar });
   return (
     <Container>
-      {isAuthenticated ? (
-        hasCalendar ? (
-          <>
-            <InfoIcon boxSize={5} marginRight={1} />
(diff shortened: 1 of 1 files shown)


client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -20,6 +30,57 @@ import EventSponsorsForm from './EventSponsorsForm';
...
+  isAuthenticated,
+  hasCalendar,
+}: IntegrationInfoData) => {
+  const infoData = ({ isAuthenticated, hasCalendar }: IntegrationInfoData) => {
+    if (isAuthenticated) {
+      if (hasCalendar) {
+        return {
+          Icon: InfoIcon,
+          text: 'Instance is authenticated with calendar api, and chapter has created calendar. Chapter will attempt creating event in the calendar.',
+        };
```
ojeytonwilliams, line 48: It doesn't seem necessary to warn people that things are going to work as expected

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -20,6 +30,57 @@ import EventSponsorsForm from './EventSponsorsForm';
...
+      }
+      return {
+        Icon: WarningTwoIcon,
+        text: "Instance is authenticated with calendar api, but chapter doesn't have calendar created. Event will not be created in the calendar.",
```
ojeytonwilliams, line 52: How about the following?

"This chapter does not have a calendar. To allow calendar event creation, go to `link-to-chapter` and create a calendar"

client/src/modules/dashboard/Events/components/EventForm.tsx
```diff
@@ -20,6 +30,57 @@ import EventSponsorsForm from './EventSponsorsForm';
...
+      }
+      return {
+        Icon: WarningTwoIcon,
+        text: 'Calendar integration is not working, and chapter does not have calendar created. Event will not be created in calendar.',
+      };
+    }
+    return {
+      Icon: WarningTwoIcon,
+      text: 'Instance is not authenticated with calendar api. Automatic creation of events in calendar is not possible.',
+    };
```
ojeytonwilliams, line 72: Do we need to distinguish between these three cases? I think it makes sense to handle `hasCalendar` separately, but if Chapter's tokens are missing or invalid, the fix is simply to reauthenticate.

With that in mind, how about this

> Chapter is not authenticated with Google Calendar. Go to `link to calendar dash` to authenticate.

> Chapter is not authenticated with Google Calendar and this chapter does not have a calendar. Go to `link to calendar dash` to authenticate. If you wish to create a calendar, visit the chapter dashboard.

ojeytonwilliams (changes requested):
It's probably a better UX to provide the authenticate + create calendar buttons on the event page, but that's a little out of scope!

I think we can improve the warnings a little, though:

client/src/modules/dashboard/Events/components/EventForm.tsx
gikf: I wanted to include it as a part if _something goes wrong_, as there isn't clear path to inform client when creating calendar event errors. 

This got me thinking, what if creating calendar and calendar event is pulled out of the create chapter/event resolvers? The actual creation can be then run once entry is created. That gives ability to separate and inform about errors on both these steps:

1. Create chapter/event. In case of errors, stop and inform client. Otherwise go to next step.
1. From client run mutation creating calendar/ calendar event. Errors can be clearly indicated as coming from calendar integration. They could be also displayed as _warnings_.

client/src/modules/dashboard/Events/components/EventForm.tsx
gikf: I'm a bit assuming that person seeing these warnings not necessarily has privileges to (re-)authenticate. Otherwise the alert on the top would be catching the attention to re-authenticate.

client/src/modules/dashboard/Events/components/EventForm.tsx
ojeytonwilliams: I like the sound of that.

Potentially the Google Event creation could fail in a bunch of ways (lack of permissions, rate limits, network error, bad configuration etc. etc.), so rather than trying to warn organisers that the Event will not get created we should focus on handling the errors when it inevitably does fail. Particularly since there are no checks we can carry out that will guarantee that an Event will be created - only that it has been.

Even if the owner has forgotten to authenticate with Google, it should only be a problem for the first event. After which they'll see the warning, authenticate and create the calendar event (once that's implemented).

What do you reckon?

gikf: That seems to be better way of handling it. I'll block this PR for now, start with simple informing if calendar/event in calendar was created, and see how to go from there.

ojeytonwilliams: Cool, that sounds like the way to go.

ojeytonwilliams:
Since we never quite resolved this one, I'll have to close it.

PR #2294, opened 2023-01-27
feat(client): add total to paginatedPage and clear the extra queries

Sboonny:

I am trying to observe the data, and hide the data based on if there is more data or not, and when it's complete and there is no more data, we can hide the buttons, instead of running the function again to findout if there is data left.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -25,7 +25,7 @@
     "@emotion/react": "11.10.5",
     "@emotion/styled": "11.10.5",
     "@hookform/resolvers": "2.9.10",
-    "chakra-confirm": "1.2.17",
+    "chakra-confirm": "1.3.0",
     "chakra-data-table": "1.4.1",
     "chakra-next-link": "1.0.10",
     "class-transformer": "0.5.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,7 +72,7 @@
         "@emotion/react": "11.10.5",
         "@emotion/styled": "11.10.5",
         "@hookform/resolvers": "2.9.10",
-        "chakra-confirm": "1.2.17",
+        "chakra-confirm": "1.3.0",
         "chakra-data-table": "1.4.1",
         "chakra-next-link": "1.0.10",
         "class-transformer": "0.5.1",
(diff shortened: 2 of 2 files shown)


client/src/pages/_app.tsx
```diff
@@ -54,15 +53,7 @@ const retryLink = new RetryLink();
...
   link: from([retryLink, errorLink, httpLink]),
-  cache: new InMemoryCache({
-    typePolicies: {
-      Query: {
-        fields: {
-          paginatedEvents: offsetLimitPagination(),
-        },
-      },
-    },
-  }),
```
Sboonny, line 65: needed to delete this, because the data doesn't exist? 

in policies.d.ts

```ts
export declare type FieldReadFunction<TExisting = any, TReadResult = TExisting, TOptions extends FieldFunctionOptions = FieldFunctionOptions> = (existing: SafeReadonly<TExisting> | undefined, options: TOptions) => TReadResult | undefined;
```

client/src/modules/home/index.tsx
```diff
@@ -45,18 +48,24 @@ const Welcome = ({ user }: { user: User }) => {
...
 
   const toast = useToast();
   const onLoadMore = async () => {
     try {
       const res = await fetchMore({
-        variables: { offset: data?.paginatedEvents.length },
+        variables: { offset: data?.paginatedEventsWithTotal.length },
       });
-      setHasMore(res.data.paginatedEvents.length > 0);
+      setHasMore(res.data.paginatedEventsWithTotal.length > 0);
```
Sboonny, line 68: I have to many possibilities, while knowing the values are always `2` and `7`. BTW, did you know, you can't mutate them, I didn't.

```ts
const total = data?.paginatedEventsWithTotal.map(({total})=> total).reduce(Number)
      setHasMore(res.data.paginatedEventsWithTotal.length < total)
      setHasMore(res.data.paginatedEventsWithTotal.length > total);
      setHasMore(res.data.paginatedEventsWithTotal.length + 2 < total)
      setHasMore(res.data.paginatedEventsWithTotal.length + 2 > total)
      setHasMore(res.data.paginatedEventsWithTotal.length += 2 < total)
      setHasMore(res.data.paginatedEventsWithTotal.length += 2 < total)
      ```

File: client/src/modules/home/index.tsx [hunk 3]
```diff
@@ -45,18 +48,28 @@ const Welcome = ({ user }: { user: User }) => {
...
     try {
       const res = await fetchMore({
-        variables: { offset: data?.paginatedEvents.length },
+        variables: { offset: data?.paginatedEventsWithTotal.length },
       });
-      setHasMore(res.data.paginatedEvents.length > 0);
+      const lookIfDataExist =
+        res.data.paginatedEventsWithTotal.length ===
+        data?.paginatedEventsWithTotal.length;
+      setHasMore(lookIfDataExist);
```
Sboonny, line 72: If the total is odd, it will hide the odd number, although people won't know, it is still a bug, but I have no idea how to fix it.

Sboonny, line 72:
```suggestion
      setHasMore(res.data.paginatedEventsWithTotal.length > 0);
```

client/src/modules/home/index.tsx
```diff
@@ -65,10 +75,13 @@ const Home = () => {
       }
     }
   };
-
   const isLoading = loading || !data;
-  if (isLoading || error) return <Loading error={error} />;
-
+  const isLoadingChapter = chapterLoading || !chapterData;
+  if (isLoading) return <Loading error={error} />;
```
Sboonny, line 80:
```suggestion
  if (isLoading || error) return <Loading error={error} />;
  
```

PR #2295, opened 2023-01-27
refactor: calendar_id to boolean has_calendar

ojeytonwilliams:
While it probably does not matter if we make the calendar_id public, the
client only cares if the calendar exists. So, that's all we need to
provide.

### DIFF
--- a/server/src/graphql-types/Chapter.ts
+++ b/server/src/graphql-types/Chapter.ts
@@ -40,7 +40,6 @@ export class Chapter extends BaseObject {
   @Field(() => Int)
   creator_id: number;
 
-  @Field(() => String, { nullable: true })
   calendar_id?: string | null;
 }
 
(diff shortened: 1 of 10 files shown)


server/src/graphql-types/Chapter.ts
```diff
@@ -40,7 +40,6 @@ export class Chapter extends BaseObject {
   @Field(() => Int)
   creator_id: number;
 
-  @Field(() => String, { nullable: true })
```
Sboonny, line 43: that is possible? 🤯

ojeytonwilliams:
I'll merge https://github.com/freeCodeCamp/chapter/pull/2277 and deal with conflicts on this one since it has no logic changes.

gikf (approved):
Looks great :rocket: 

That's pretty interesting case with the `calendar_id` in graphql type.

PR #2301, opened 2023-01-27
feat: add search functionality in chapterDashboard

Sboonny:

Because of the sort of the data in this PR https://github.com/freeCodeCamp/chapter/pull/2291, I thought maybe a filter function is a due. This should be in Utils later, but for now I am testing it in chapterPages

### DIFF
--- a/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
@@ -47,11 +47,10 @@ const actionLinks = [
   },
 ];
 
-
 export const ChaptersPage: NextPageWithLayout = () => {
+  const [filterChapter, setFilterChapter] = useState('');
   const { loading, error, data } = useDashboardChaptersQuery();
   const { user, loadingUser } = useUser();
-  const { setSearch, search } = useState('');
 
   const hasPermissionToCreateChapter = checkInstancePermission(
     user,
@@ -62,7 +61,7 @@ export const ChaptersPage: NextPageWithLayout = () => {
   if (isLoading || error) return <DashboardLoading error={error} />;
 
   const filteredChapters = data.dashboardChapters.filter(({ name }) =>
-    name.toLowerCase().includes(search.toLowerCase().trim()),
+    name.toLowerCase().includes(filterChapter.toLowerCase().trim()),
   );
 
   return (
@@ -72,20 +71,24 @@ export const ChaptersPage: NextPageWithLayout = () => {
           Chapters
         </Heading>
         <Text srOnly>
-          Type the name of chapter that you are looking for in the filter chapter input
-          to filter out other chapters
+          Type the name of chapter that you are looking for in the filter
+          chapter input to filter out other chapters
         </Text>
-        <Input
-          width="full"
-          type="text"
-          backgroundColor="gray.50"
-          placeholder="Filter Chapter"
-          value={search}
-          gridRowStart="2"
-          gridColumnStart="1"
-          gridColumnEnd="-1"
-          onChange={(e) => setSearch(e.target.value)}
-        />
+        {data.dashboardChapters.length > 1 && (
+          <Input
+            width="full"
+            type="text"
+            backgroundColor="gray.50"
+            placeholder="Filter Chapter"
+            value={filterChapter}
+            gridRowStart="2"
+            gridColumnStart="1"
+            gridColumnEnd="-1"
+            onChange={(e) => {
+              return setFilterChapter(e.target.value);
+            }}
+          />
+        )}
         {hasPermissionToCreateChapter && (
           <LinkButton
             data-cy="new-chapter"
--- a/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
@@ -91,7 +91,7 @@ export const ChaptersPage: NextPageWithLayout = () => {
   if (isLoading || error) return <DashboardLoading error={error} />;
 
   const filteredChapters = data.dashboardChapters.filter(({ name }) =>
-    name.toLowerCase().includes(search.trim()),
+    name.toLowerCase().includes(search.toLowerCase().trim()),
   );
 
   return (
--- a/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
@@ -70,11 +70,12 @@ export const ChaptersPage: NextPageWithLayout = () => {
         <Heading data-cy="chapter-dash-heading" id="page-heading">
           Chapters
         </Heading>
+        {data.dashboardChapters.length > 1 && (
+        <>
         <Text srOnly>
           Type the name of chapter that you are looking for in the filter
           chapter input to filter out other chapters
         </Text>
-        {data.dashboardChapters.length > 1 && (
           <Input
             width="full"
             type="text"
@@ -86,6 +87,7 @@ export const ChaptersPage: NextPageWithLayout = () => {
             gridColumnEnd="-1"
             onChange={(e) => setFilterChapter(e.target.value)}
           />
+          </>
         )}
         {hasPermissionToCreateChapter && (
           <LinkButton
--- a/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
@@ -84,9 +84,7 @@ export const ChaptersPage: NextPageWithLayout = () => {
             gridRowStart="2"
             gridColumnStart="1"
             gridColumnEnd="-1"
-            onChange={(e) => {
-              return setFilterChapter(e.target.value);
-            }}
+            onChange={(e) => setFilterChapter(e.target.value)}
           />
         )}
         {hasPermissionToCreateChapter && (
--- a/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
+++ b/client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
@@ -10,7 +10,7 @@ import {
 } from '@chakra-ui/react';
 import { DataTable } from 'chakra-data-table';
 import { LinkButton } from 'chakra-next-link';
-import React, { ReactElement, useCallback, useReducer } from 'react';
+import React, { ReactElement, useState } from 'react';
 
 import {
   checkInstancePermission,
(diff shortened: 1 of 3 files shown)


client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -50,28 +90,55 @@ export const ChaptersPage: NextPageWithLayout = () => {
...
+  // );
+
+  // Fails, but why?
+  const filteredChapter = useMemo(
+    () =>
+      data.dashboardChapters.filter(({ name }) =>
+        name.toLowerCase().includes(search),
+      ),
+    [data, search],
+  );
```
Sboonny, line 110: Is this failing because it relies on a condition? The hooks run on mount, but the data is got on load?

Sboonny: RIP performance, I guess 🤣. maybe I can move the logic into the top and create a function that run it, but idk if that actually worth it, while this works

gikf (commented):
Works nicely :rocket:.

There isn't any issue with complexity, but was there some specific reason for not attempting using `useState` for the filter string, instead of opting for `useReducer`?

There's couple nitpicks below.

The search field is not very visually separated from the _Add new_ button. It makes it appear on the first look, like both of them are related.
image
Moving filter field to it own line might be the quickest way to clearly give them both their own space:
image
What do you think?

Sboonny:
> but was there some specific reason for not attempting using `useState` for the filter string, instead of opting for `useReducer`?

I had more experience with `setState` and I wanted to try something else, but after this, I wanted to try to make it with `useState`, don't know how to apply `dispatch` to it, tho, maybe with `flushSync` https://beta.reactjs.org/reference/react-dom/flushSync 🤔.

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -50,28 +90,51 @@ export const ChaptersPage: NextPageWithLayout = () => {
   const isLoading = loading || loadingUser || !data;
   if (isLoading || error) return <DashboardLoading error={error} />;
 
+  const filteredChapters = data.dashboardChapters.filter(({ name }) =>
+    name.toLowerCase().includes(search.trim()),
```
gikf, line 94:
```suggestion
    name.toLowerCase().includes(search.toLowerCase().trim()),
```
I forgot to add this earlier

Sboonny: I should have noticed as well 🤣

gikf:
> I had more experience with `setState` and I wanted to try something else, but after this, I wanted to try to make it with `useState`, don't know how to apply `dispatch` to it, tho, maybe with `flushSync` https://beta.reactjs.org/reference/react-dom/flushSync thinking.

With `useState` there wouldn't be really what to dispatch. Set function would replace in `onChange` dispatch from reducer and that's basically it.

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -38,9 +47,40 @@ const actionLinks = [
...
+
+  const setSearch = useCallback((search: string) => {
+    dispatch({
+      type: 'setSearch',
+      payload: search,
+    });
+  }, []);
+
+  return { setSearch, search };
+};
```
ojeytonwilliams, line 78:
```suggestion
```
I think we could clean up the codebase by applying `useReducer` instead of juggling a bunch of `useState`s. However, in this case, it's a complicated way to mimic a single `useState`.

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -38,9 +47,40 @@ const actionLinks = [
...
+
 export const ChaptersPage: NextPageWithLayout = () => {
   const { loading, error, data } = useDashboardChaptersQuery();
   const { user, loadingUser } = useUser();
+  const { setSearch, search } = useFilter();
```
ojeytonwilliams, line 83:
```suggestion
  const { setSearch, search } = useState('');
```

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -50,28 +90,51 @@ export const ChaptersPage: NextPageWithLayout = () => {
...
   return (
     <VStack>
-      <Flex w="full" justify="space-between">
+      <Grid w="full" gap="1em" gridTemplateColumns="1fr 2fr 8em">
         <Heading data-cy="chapter-dash-heading" id="page-heading">
           Chapters
         </Heading>
+        <Text srOnly>
+          Type the name of chapter, that you are looking for, in the FilterBar
+          to find it.
```
ojeytonwilliams, line 105:
```suggestion
          Type the name of chapter that you are looking for in the filter chapter input
          to filter out other chapters
```

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -50,28 +90,51 @@ export const ChaptersPage: NextPageWithLayout = () => {
...
+        </Text>
+        <Input
+          width="full"
+          type="text"
+          backgroundColor="gray.50"
+          placeholder="Filter Chapter..."
+          value={search}
+          aria-Label="FilterBar"
```
ojeytonwilliams, line 113:
```suggestion
          aria-label="FilterBar"
```
Is this necessary? My screen already reads the placeholder when I focus it, it seems unhelpful to repeat that info. That said it reads "Filter Chapter dot dot dot" which isn't great. If it could just say Filter Chapter, that would be perfect.

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -1,7 +1,16 @@
...
+  Heading,
+  HStack,
+  Input,
+  Text,
+  VStack,
+} from '@chakra-ui/react';
 import { DataTable } from 'chakra-data-table';
 import { LinkButton } from 'chakra-next-link';
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useCallback, useReducer } from 'react';
```
ojeytonwilliams, line 13:
```suggestion
import React, { ReactElement, useState } from 'react';
```
See below.

ojeytonwilliams (changes requested):
I feel bad asking you to replace the reducer, but it's more than we need here.

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -50,28 +90,51 @@ export const ChaptersPage: NextPageWithLayout = () => {
...
         </Heading>
+        <Text srOnly>
+          Type the name of chapter, that you are looking for, in the FilterBar
+          to find it.
+        </Text>
+        <Input
+          width="full"
+          type="text"
+          backgroundColor="gray.50"
+          placeholder="Filter Chapter..."
```
Sboonny, line 111:
```suggestion
          placeholder="Filter Chapter"
```

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
Sboonny:
```suggestion
```

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
ojeytonwilliams: Sorry, I somehow missed the entire discussion. I'll revisit this tomorrow, but for a quick answer: if you're planning to implement those things straight away, let's go with the reducer. If you're not, state is fine for now.

The rationale is just [YAGNI](https://en.m.wikipedia.org/wiki/You_aren't_gonna_need_it), so if you _do_ need it, go right ahead.

Sboonny: It won't need to reduce an array state, if I don't include the tags, and it will need to reduce 2 boolean states and string.

I won't lie, I would like to delete it and use state to apply this https://github.com/freeCodeCamp/chapter/pull/2301#issuecomment-1408387596 and gain more experience, up to you tho, if this should stay or not, I am happy either way.

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -39,6 +48,7 @@ const actionLinks = [
 ];
 
 export const ChaptersPage: NextPageWithLayout = () => {
+  const [filterChapter, setFilterChapter] = useState('');
```
Sboonny, line 51: did you know that the type will fail if it's `{filterChapter, setFilterChapter}`. I have spent a good while on it, trying to use generics, until I gave up and copied `hasMore` into it

ojeytonwilliams: `hasMore`? The code you've written seems fine as is.

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -50,28 +60,54 @@ export const ChaptersPage: NextPageWithLayout = () => {
...
+            type="text"
+            backgroundColor="gray.50"
+            placeholder="Filter Chapter"
+            value={filterChapter}
+            gridRowStart="2"
+            gridColumnStart="1"
+            gridColumnEnd="-1"
+            onChange={(e) => {
+              return setFilterChapter(e.target.value);
+            }}
```
ojeytonwilliams, line 89:
```suggestion
            onChange={(e) => setFilterChapter(e.target.value)}
```
Nit: `onChange` doesn't use the return value, so there's no need to provide it.

ojeytonwilliams (dismissed):

client/src/modules/dashboard/Chapters/pages/ChaptersPage.tsx
```diff
@@ -50,28 +60,52 @@ export const ChaptersPage: NextPageWithLayout = () => {
...
+            type="text"
+            backgroundColor="gray.50"
+            placeholder="Filter Chapter"
+            value={filterChapter}
+            gridRowStart="2"
+            gridColumnStart="1"
+            gridColumnEnd="-1"
+            onChange={(e) => setFilterChapter(e.target.value)}
+          />
+        )}
```
Sboonny, line 89:
```suggestion
        {data.dashboardChapters.length > 1 && (
        <>
        <Text srOnly>
          Type the name of chapter that you are looking for in the filter
          chapter input to filter out other chapters
        </Text>
          <Input
            width="full"
            type="text"
            backgroundColor="gray.50"
            placeholder="Filter Chapter"
            value={filterChapter}
            gridRowStart="2"
            gridColumnStart="1"
            gridColumnEnd="-1"
            onChange={(e) => setFilterChapter(e.target.value)}
          />
          </>
        )}
```

gikf (approved):

PR #2308, opened 2023-01-29
feat: use type export for the permission types enums

Sboonny:

I noticed that the dashboard is slower in my devices after making the change in #2121, although I may be overthinking it, and this may be doing nothing, but it's a try.

This PR trys to remove the types import from the runTime https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export


ojeytonwilliams (changes requested):
Sorry for the wait, @Sboonny :cold_face: 

So, this looks good, but could you update it so that `InstancePermission` and `ChapterPermission` are imported as types? It's not a big deal, but it's a little clearer.

ojeytonwilliams (approved):
LGTM :+1:

PR #2310, opened 2023-01-29
feat: form validations for remaining forms

atosh502:

Form validations for `ProfileForm`, `SponsorForm`, `EventForm`, and subforms of `EventForm`
Fixes and refactor related to validations

Depends on https://github.com/freeCodeCamp/chapter/pull/2290. Only merge this after #2290 is merged.


Sboonny (commented):
If someone fills the wrong value for Latitude, then clear the field, they won't be able to add venue, unless they have fixed it.

[rec-tab.webm](https://user-images.githubusercontent.com/88248797/216545279-7c87764d-f4a6-4271-a520-63d83cbb707d.webm)

atosh502:
Hi @Sboonny, fixed the issue by writing a custom validator... `null` is assigned as the default value. Once the input was touched or edited, the value changed to `NaN` and so the `@IsLatitude()` or `@IsLongitude()` validation failed as it expected a valid coordinate value...

There's one issue though, the number input (for latitude and longitude) is editable in firefox (not in Chrome or Edge). So when we enter alphanumeric characters in the number input in firefox the validations won't be run and will simply be ignored. In addition, no value is saved for the number input and `null` is passed to the backend as well...
image

Sboonny:
> Hi @Sboonny, fixed the issue by writing a custom validator... `null` is assigned as the default value. Once the input was touched or edited, the value changed to `NaN` and so the `@IsLatitude()` or `@IsLongitude()` validation failed as it expected a valid coordinate value...
> 
> There's one issue though, the number input (for latitude and longitude) is editable in firefox (not in Chrome or Edge). So when we enter alphanumeric characters in the number input in firefox the validations won't be run and will simply be ignored. In addition, no value is saved for the number input and `null` is passed to the backend as well... image

I should have investigated it more 😨, let's use your previous version, because it requires the user to put input first, before the bugs occur, and this bug is fixed in future version of react-hook-form.

atosh502:
> because it requires the user to put input first, before the bugs occur, and this bug is fixed in future version of react-hook-form.

It's actually a [bug in firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1398528). The inputs/form will work as expected in Chromium based browsers.

Sboonny:
Oh, I am talking about the previous implementation, which had the validation state stop the API call, event when Latitude isn't required. 

This should be similar to this discussion https://github.com/react-hook-form/react-hook-form/discussions/9875#discussioncomment-4877380, and if RTF doesn't currently clear the error, after the value changing the value

image

client/src/modules/dashboard/Venues/components/VenueForm.tsx
```diff
@@ -61,7 +61,9 @@ const VenueForm: React.FC<VenueFormProps> = (props) => {
           isDisabled={loading}
```
Sboonny, line 61: to get the same console.log, you need to change line 41 in this file to this


```ts
resolver: async (data, context, options) => {
      // you can debug your validation schema here
      console.log('formData', data)
      console.log('validation result', await resolver(data, context, options))
      return resolver(data, context, options)
    },
    ```
 
 I think the solution is to clear the error through zod or yub, that's why I mentioned I should have investigated it more, sorry for cutting corners

ojeytonwilliams:
Hey @atosh502, @Sboonny if I understand this correctly, there's no way to get Firefox to handle 'number' input correctly.  It converts empty strings and any non-numeric string into `""`, so any code we write will get `""` for both empty and invalid input.

I don't particularly like it, but it sounds like we need to use a 'text' input and live with the fact that all users can enter invalid input.

client/src/modules/dashboard/Venues/components/VenueForm.tsx
atosh502: The issue before https://github.com/freeCodeCamp/chapter/commit/2bce8e9f18384df111cd60e0b120d0a6fd9644b4 is:

At first we set the default value of `latitude` and `longitude` to `null`. As soon as we touch the inputs its value changes to an empty string `""` or `NaN` (if we're using `{ valueAsNumber: true }`). This seems to be the default behavior of RHF and can be seen in both Firefox and Chrome. Next our validators `IsLatitude()` and `IsLongitude()` gets passed this empty string `""` or `NaN` value for validations, and since both the validators treat `""` and `NaN` as invalid latitude or longitude coordinates we get the error mentioned [above](https://github.com/freeCodeCamp/chapter/pull/2310#issuecomment-1418704540)

Sboonny (approved):

client/src/modules/dashboard/Venues/components/VenueForm.tsx
Sboonny: Thank you for providing more info, I understand the issue more now. 👍 💚

atosh502: > This should be similar to this discussion https://github.com/react-hook-form/react-hook-form/discussions/9875#discussioncomment-4877380, and if RTF doesn't currently clear the error, after the user has changed the value

I think in our case the error wasn't being cleared because our validators were being fed the wrong values and so the errors didn't go away.

In https://github.com/freeCodeCamp/chapter/commit/2bce8e9f18384df111cd60e0b120d0a6fd9644b4, the value for latitude and longitudes were only validated if they were non-empty or valid numbers. 

This allowed us to ignore the error. But then on testing with Firefox I found the issue where Firefox would allow arbitrary strings to be typed into number input.

atosh502:
> but it sounds like we need to use a 'text' input and live with the fact that all users can enter invalid input.

I think we can ignore the issue with Firefox for now as it's a well-known ~feature~issue and there are no clear solutions to handle it. Converting the input to `text` would also take away the ability to change latitude or longitude values using up/down arrow keys. Also, even if we can type in the number input field, the actual form values would still be null and it wouldn't break our backend validations.

Thank you @Sboonny and @ojeytonwilliams for your reviews and suggestions during this and previous PR related to form validations. :rocket:

ojeytonwilliams:
> Converting the input to text would also take away the ability to change latitude or longitude values using up/down arrow keys.

That's a good point, thanks @atosh502. I'm happy to leave it for now. We can see what Firefox users have to say before we attempt a fix.

Regardless, it's a way better experience than what we had before!

PR #2311, opened 2023-01-29
fix: loadingUser stuck on error

### LINKED ISSUE #2259: loadingUser doesn't change on error
### Describe the bug:
When `useMeQuery` in `client/src/modules/auth/user.tsx` errors, the `loadingUser` is not changed.

This makes possible to be stuck in loading when there's some error during the query.

I'm not entirely sure what's the appropriate action in such case. The _simple_ thing that comes to mind is just adding error case to the useEffect:
```
    if (error) {
      setData({ loadingUser: false, isLoggedIn: false });
    }
```

gikf:

Closes #2259

---

- Changes `loadingUser` to `false` when there's `useMeQuery` error.
- Catches errors when creating or destroying session.
- `logout()` along setting `loadingUser` on error is a hacky try to start next time from as clean state as possible. For development that's flipping the `isAuthenticated`.

### DIFF
--- a/client/src/hooks/useSession.ts
+++ b/client/src/hooks/useSession.ts
@@ -61,11 +61,17 @@ export const useSession = () => {
 
     if (toLogout) {
       destroySession()
+        .then((res) => {
+          if (!res.ok) throw Error(`Server returned ${res.status} error.`);
+        })
         .then(() => refetch())
         .then(() => apollo.resetStore())
         .catch((e) => setError(e));
     } else if (isAuthenticated) {
       createSession()
+        .then((res) => {
+          if (!res.ok) throw Error(`Server returned ${res.status} error.`);
+        })
         .then(() => refetch())
         .catch((e) => setError(e));
     }
--- a/client/src/hooks/useSession.ts
+++ b/client/src/hooks/useSession.ts
@@ -40,13 +40,16 @@ export const useSession = () => {
   // isAuthenticated does not change immediately after login/logout, so we
   // need to track the intent to prevent accidental session creation/deletion
   const [toLogout, setToLogout] = useState(false);
+  const [error, setError] = useState<Error | undefined>(undefined);
 
   const login = async () => {
+    setError(undefined);
     await loginAuth();
     setCanAlterSession(true);
     setToLogout(false);
   };
   const logout = async () => {
+    setError(undefined);
     await logoutAuth();
     setCanAlterSession(true);
     setToLogout(true);
@@ -59,11 +62,14 @@ export const useSession = () => {
     if (toLogout) {
       destroySession()
         .then(() => refetch())
-        .then(() => apollo.resetStore());
+        .then(() => apollo.resetStore())
+        .catch((e) => setError(e));
     } else if (isAuthenticated) {
-      createSession().then(() => refetch());
+      createSession()
+        .then(() => refetch())
+        .catch((e) => setError(e));
     }
   }, [isAuthenticated, canAlterSession, toLogout]);
 
-  return { login, logout, isAuthenticated };
+  return { login, logout, isAuthenticated, error };
 };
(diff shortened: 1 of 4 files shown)


client/src/hooks/useSession.ts
```diff
@@ -40,13 +40,16 @@ export const useSession = () => {
   // isAuthenticated does not change immediately after login/logout, so we
   // need to track the intent to prevent accidental session creation/deletion
   const [toLogout, setToLogout] = useState(false);
+  const [error, setError] = useState<Error | undefined>(undefined);
```
ojeytonwilliams, line 43:
```suggestion
  const [error, setError] = useState<Error | null>(null);
```
Bit nicer, imo.

client/src/hooks/useSession.ts
```diff
@@ -40,13 +40,16 @@ export const useSession = () => {
...
 
   const login = async () => {
+    setError(undefined);
```
ojeytonwilliams, line 46:
```suggestion
    setError(null);
```

client/src/hooks/useSession.ts
```diff
@@ -40,13 +40,16 @@ export const useSession = () => {
...
     await loginAuth();
     setCanAlterSession(true);
     setToLogout(false);
   };
   const logout = async () => {
+    setError(undefined);
```
ojeytonwilliams, line 52:
```suggestion
    setError(null);
```

client/src/hooks/useSession.ts
```diff
@@ -59,11 +62,14 @@ export const useSession = () => {
     if (toLogout) {
       destroySession()
         .then(() => refetch())
-        .then(() => apollo.resetStore());
+        .then(() => apollo.resetStore())
+        .catch((e) => setError(e));
```
ojeytonwilliams, line 66: The weird thing about `fetch` is that it doesn't throw errors if there are issues with the request. That means even if we get a status 500 from the server, we don't get to the catch.  You can easily test this by calling `next('message')` in the `/login` middleware.

To have something to catch, you can simply throw an error inside create/destroySession if `ok` is falsy. You _could_ handle it it with a conditional in the `then`, but catching seems fine here!

ojeytonwilliams (changes requested):

client/src/hooks/useSession.ts
gikf: This is interesting. I generally was putting a hammer on server, by deleting some line in `app.ts` so it would error after another compilation. For that case `fetch` is indeed throwing.

ojeytonwilliams: You may be seeing something different, but for what it's worth I was confused by the network errors appearing in the logs. They're not JS errors, they just register that something went wrong during a network request. If they go away when you check "Hide network" in the browser's dev console, we're talking about the same thing.

To the best of my knowledge, you can't catch those errors. However, if you're able to create JS errors by messing with the server, that's really interesting. At least to me, because I'll have learned a thing!

gikf: Try these steps to reproduce:
- Remove `catch` chained to `destroySession()` and `createSession()`.
- While having running server, change something in `/server/src/app.ts` to make it not compile anymore. For example remove line with `const PORT = process.env.PORT || 5000;`
- Now when pressing _Login_ or _Logout_ there will be unhandled error, from the `fetch`.

ojeytonwilliams: Thanks for the quick steps to reproduce, it does indeed error in that case! I did a little digging and it turns out that's just how it works: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#differences_from_jquery

Bit strange, but your code handles both cases, so I think we're good.

ojeytonwilliams (approved):
LGTM :+1:

Issue #2071, opened 2022-12-08
Consolidate email templates


ojeytonwilliams:
Currently the text of the emails that Chapter sends is embedded directly in the controllers.  If we move them into their own file this both removes clutter from the controllers and makes it easier to audit the text and ensure that it has a consistent voice.

`server/src/email-templates.ts` seems like a good enough place for these helper functions for now.

dineshigdd:
can you please assign me this?

UrielOfir:
Hi :)
I would like to assign myself for this issue.
I didn't understand what did you mean that "emails sending is embedded directly in the controllers".
Did you mean to the code in this file-
https://github.com/freeCodeCamp/chapter/blob/08cb5b009187e7ce7530350365b881c37daa8c3e/server/src/controllers/Users/resolver.ts
line 60?

ojeytonwilliams:
Hi @UrielOfir. What I was trying to say was that we should not put the email content directly in the resolvers that handle the user's request.  For example

https://github.com/freeCodeCamp/chapter/blob/08cb5b009187e7ce7530350365b881c37daa8c3e/server/src/controllers/Users/resolver.ts#L60-L67

bloats the resolver logic. If it read like this

```ts
    emailTemplates.sendInstanceRoleChanged({ name: user.name, newRole })
```

it would be a lot cleaner.

Chirag2193:
This issue seems to have been merged under above PR #2319

allella:
It looks like the constants were moved out of the revolvers in PR #2319, but @ojeytonwilliams 's [example above](https://github.com/freeCodeCamp/chapter/issues/2071#issuecomment-1375651014) seemed to go further by moving the mailerService.sendEmail out of the resolver.

Oliver, is there more to go done here?

PR #2319, opened 2023-01-31
refactor: email text templates

### LINKED ISSUE #2271: Email text for event changes could be improved
Currently the text reflects the new _state_ of the venue, date or streaming url, but we are telling people about a _change_ and the text should reflect that. 

The transitions we need to cover are

Venue -> no venue
No venue -> venue
Venue -> different venue

with different text describing the change for each case. The same transitions need to be handled for streaming url.

Also, if the event was online only, but has become a physical event (or vice versa), that's important information we should convey.

For example:

> This was only online, but is now being held at `<venue-details>`.

If it's still online

> Online attendees are still welcome

vs

> It will no longer be possible to attend online

### Tests

Since regressions in the email text keep happening, some unit tests for the various transitions would be great.

### LINKED ISSUE #2278: New event email text does not handle venues or streaming urls correctly
Related to https://github.com/freeCodeCamp/chapter/issues/2271 we need unit tests for the email text, because it's easy to break them accidentally.

For example 
email text

gikf:

Closes #2278 
Closes #2271 
Ref #2071

---

- ~Very in-progress.~ Along consolidating texts I will add tests and therefore fix the two other issues.
- ~So far main part of texts was moved outside of `/Events/resolver.ts`~
- Actually full consolidation of texts, making them consistent across different emails might need separate PR.


Sboonny (approved):

PR #2323, opened 2023-02-01
feat(client): make events card clickable

### LINKED ISSUE #2299: Putting chapter.name on event cards is not necessary on the chapter's page
While it's useful to have the chapter in event cards on the home page, it's redundant on the chapter's page. For example:
chapter name repeated in event list

@Sboonny do you think this is a candidate for a Good First Issue label? If so, could you add it?

Sboonny:

Close #2299 

The solution was simple, I was focused on theme.ts file, and didn't find it while it's under my nose

### DIFF
--- a/client/src/components/EventCard.tsx
+++ b/client/src/components/EventCard.tsx
@@ -87,6 +87,7 @@ export const EventCard: React.FC<EventCardProps> = ({ event }) => {
       overflow="hidden"
       width={'full'}
       {...(hasEnded && { opacity: 0.6 })}
+      position="relative"
     >
       <Image
         display={['none', 'block']}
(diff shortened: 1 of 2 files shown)


client/src/components/EventCard.tsx
```diff
@@ -109,12 +110,22 @@ export const EventCard: React.FC<EventCardProps> = ({ event }) => {
         >
           <Link
             data-cy="event-link"
+            className="card-link"
```
Sboonny, line 113:
```suggestion
```

gikf (commented):
Not sure why, but text with event status and date breaks the link.
image
What is strange I tried to do this with chakra's _Link Overlay_ component and effect was the same. It doesn't appear like it should be like that, but maybe that's normal.

Sboonny:
> What is strange I tried to do this with chakra's Link Overlay component and effect was the same. It doesn't appear like it should be like that, but maybe that's normal.

When adding property that create a stack `position`, the browser pushes the anchor element below the text, if they are in the same layer of the stack.

Which is good because if we are visiting a doctor website and need to copy a medicine name, CSS shouldn't block us.

But we don't need that behavior here, because the dates are sent to google calendar when someone attends

---

You know weird part:
- If you copy the `pseudo-element` and move it to the text, without `z-index` value, it will overshadow the link element, and you can't click it again.
- If you add `z-index` in the dev-tools, it will warn you, even when you have `position: absolute`.

image

gikf (approved):
LGTM :+1:

Sboonny:
revert what I said in the previous comment, the browser gives higher specificity for the latest element, if they are on the same stack's layer.

So without `z-index`, link can take the full space, if it is after the date, otherwise `Text` will have higher specificity on the stack

```html
<Text>Date</Text>
<Link>EventName</Link>
```

PR #2327, opened 2023-02-02
fix: remove duplicate fields

Ismailtlem:

Just removing duplicate fields in this query


Sboonny (commented):
The test complain, that the API has different query than the new one. While the server is running, you can run `npm run gen` and it will sort it

gikf (commented):
Thanks for finding these. Could you run `npm run gen` script and add new commit with changed files?

gikf (approved):

PR #2340, opened 2023-02-05
feat: add total to paginatedPage and clear the extra queries

Sboonny:

I am trying to observe the data, and hide the data based on if there is more data or not, and when it's complete and there is no more data, we can hide the buttons, instead of running the function again to find out if there is data left.

---

Apparently if you ruin the git history, and decide to discard the changes, for new one good commit. GitHub close the open PR and you have to open new one. 🤔 To be honest, I am missing something with how to do it cleanly with Git so no need to focus on that PR anymore, because I can't update commits to it https://github.com/freeCodeCamp/chapter/pull/2294


client/src/modules/home/index.tsx
```diff
@@ -1,25 +1,21 @@
...
+  useChaptersLazyQuery,
+} from '../../generated/graphql';
+import { Pagination } from '../util/pagination';
 import { UserContextType, useUser } from '../auth/user';
 import { getNameText } from '../../components/UserName';
 
-type User = NonNullable<UserContextType['user']>;
+const eventCards = 2;
 
+type User = NonNullable<UserContextType['user']>;
```
Sboonny, line 18:
```suggestion
type User = NonNullable<UserContextType['user']>;

```

client/src/modules/home/index.tsx
```diff
@@ -44,30 +40,34 @@ const Welcome = ({ user }: { user: User }) => {
...
+  const [getChapters, { error: chapterError, data: chapterData }] =
+    useChaptersLazyQuery();
+
+  const { loading, error, data, fetchMore } = usePaginatedEventsWithTotalQuery({
+    variables: { offset, limit: eventCards, showUpcoming: true },
+    onCompleted() {
+      getChapters();
+    },
   });
+
```
Sboonny, line 55:
```suggestion
```

Sboonny:
Will be affect by https://github.com/freeCodeCamp/chapter/pull/2224, so I am blocking it for now

server/src/controllers/Events/resolver.ts
```diff
@@ -230,44 +229,36 @@ const getNameForNewAttendance = (event: EventAttendanceName) => {
...
+        },
+      }),
       include: {
         chapter: true,
       },
       orderBy: [{ start_at: 'asc' }, { name: 'asc' }],
       take: limit ?? 10,
       skip: offset,
     });
+    return events.map((event) => ({ ...event, total }));
```
gikf, line 261: Adding `total` to each event object might be reason for part of the issues. Why not return `total` separately?

gikf (commented):
Couple things I noticed:
- There's warning in console when going from _Events_ to _Home_ page. 
- After depleting on _Events_ page, events by clicking for more until all events are displayed, and going back to _Home_ page, the click for more button there goes bonkers, it doesn't hide when it should. It looks like events having `total` including all events mixed up with events having only upcoming events.

Is it needed to change pagination on _Events_ page to the click for more version?

client/src/pages/_app.tsx
```diff
@@ -58,7 +58,7 @@ const client = new ApolloClient({
     typePolicies: {
       Query: {
         fields: {
-          paginatedEvents: offsetLimitPagination(),
+          paginatedEventsWithTotal: offsetLimitPagination(),
```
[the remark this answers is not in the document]
Sboonny: The issue do happen when navigating from home to events too, but doesn't happen when navigating from chapters to home and vise versa.

I think apollo didn't understand that the data we are fetching is the same as data that it cached, because I didn't define the params. 

Kinda like, hey these objects aren't the same as `paginatedEvents(offset, limit)`, so let's refetch the same data with the same keys, instead of using the cached data.

Which result in having two of the same object with the same key as reference, and React warning us about this.

This is pure guess from reading the docs, which I misread a lot. I haven't seen an actual code.

gikf:
> Warning: Can't perform a React state update on a component that hasn't mounted yet. This indicates that you have a side-effect in your render function that asynchronously later calls tries to update the component. Move this work to useEffect instead.


I've found this error is caused by the changes to load chapters lazily. What do you think about moving that to separate PR, to limit the scope?

Sboonny:
I can't delet the home query without creating conflict

client/src/modules/home/index.tsx
```diff
@@ -84,22 +83,32 @@ const Home = () => {
             <Heading as="h2" size={'md'}>
               Upcoming events
             </Heading>
-            {data.paginatedEvents.map((event) => (
+            {paginatedEventsWithTotal.map((event) => (
```
gikf, line 86:
```suggestion
            {events.map((event) => (
```

client/src/modules/home/index.tsx
```diff
@@ -84,22 +83,32 @@ const Home = () => {
...
               <EventCard key={event.id} event={event} />
             ))}
-            {hasMore ? (
-              <Button onClick={onLoadMore}>Click for more</Button>
-            ) : (
-              <Text size="md">No more</Text>
-            )}
+            <Pagination
+              currentPage={currentPage}
+              setCurrentPage={setCurrentPage}
```
gikf, line 91:
```suggestion
              onClickForMore={() => onClickForMore()
```

client/src/modules/util/pagination.tsx
```diff
@@ -0,0 +1,36 @@
...
+  currentPage: number;
+  setCurrentPage: (page: number) => void;
+  eventCards: number;
+  records: number;
+  displayOnEmpty?: React.ReactNode;
+}
+
+export const Pagination = ({
+  currentPage,
+  setCurrentPage,
```
gikf, line 14:
```suggestion
  onClickForMore,
```

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -1,80 +1,42 @@
...
   });
 
   useEffect(() => {
     if (visitedPages.has(currentPage)) return;
     fetchMore({
-      variables: { offset, limit: pageSize },
+      variables: { offset, limit: eventCards },
     });
     setVisitedPages(new Set(visitedPages).add(currentPage));
   }, [currentPage]);
```
gikf, line 29: I'm unable to suggest this properly in github due to deleted lines. It's the part to replace lines 15 - 29.
```
  const { user } = useUser();
    variables: { offset: 0, limit: eventCards, showOnlyUpcoming: false },
  });

  const events = useMemo(
    () =>
      data?.paginatedEventsWithTotal.flatMap((eventData) => eventData.events) ??
      [],
    [data?.paginatedEventsWithTotal],
  );

  const currentPage = useMemo(
    () => Math.floor(events.length / eventCards),
    [events],
  );

  const onClickForMore = () => {
    const offset = currentPage * eventCards;
    console.log(offset, eventCards);
    fetchMore({
      variables: { offset, limit: eventCards },
    });
  };
```

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -1,80 +1,42 @@
...
 
   const isLoading = loading || !data;
   if (isLoading || error) return <Loading error={error} />;
+  // we need to do this because the data is array of event for every total?
+  const paginatedEventsWithTotal = data?.paginatedEventsWithTotal.flatMap(
+    (eventData) => eventData.events,
+  );
```
gikf, line 36:
```suggestion
```

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -87,14 +49,32 @@ export const EventsPage: NextPage = () => {
             </LinkButton>
           )}
         </Flex>
-        {data?.paginatedEventsWithTotal.events.map((event) => (
-          <EventCard key={event.id} event={event} />
-        ))}
+        {
+          <>
+            {paginatedEventsWithTotal.map((event) => (
```
gikf, line 54:
```suggestion
            {events.map((event) => (
```

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -87,14 +49,32 @@ export const EventsPage: NextPage = () => {
...
+              <EventCard key={event.id} event={event} />
+            ))}
+          </>
+        }
         <Pagination
           currentPage={currentPage}
           setCurrentPage={setCurrentPage}
```
gikf, line 61:
```suggestion
          onClickForMore={() => onClickForMore()}
```

gikf (commented):
What stands out is the _Click for more_ button in some situations. To reproduce:
- On home page (it applies also to the events page) click _Click for more_ once or twice.
- Navigate to different page.
- Go back to home page.

Notice that:
- If in first step clicks depleted available events, the _Click for more_ button is again visible.
- Despite of having more event cards displayed than on fresh visit, _Click for more_ button, behaves as if, only initial events are loaded. That means for example: 
  - If on the first visit you clicked once, now you need to click twice, to see new events actually loaded.
  - If on the first visit you depleted events by clicking the button. To make button go away again, it's needed to click the same number of times.

I've managed to make work as expected by using two `useMemo`. Suggestions below are implementing that. I haven't touched anything else, there's couple nitpicks that I'm going to nitpick after ironing out these more important kinks.

It's very possible most of the logic, if not all, could be kept in the `Pagination` component, but that's for once all is working.

gikf (commented):
I forgot about imports.

client/src/modules/events/pages/eventsPage.tsx
Sboonny: Added, without the console log 👍.

gikf:
@Sboonny would you mind if I push couple commits directly? I've made further changes to make `Pagination` more generic. It'd be more convenient to share it like that, instead of making dozen suggestions to make github happy.

Sboonny:
Nope, feel free to 👍, remove it from draft when you are done

gikf:
@ojeytonwilliams If you could take a look and review this in spare time, with fresh pair of eyes, that'd be great.

client/src/modules/util/pagination.tsx
```diff
@@ -0,0 +1,45 @@
...
+  dataFlattener,
+  displayOnEmpty,
+  itemsPerPage,
+  mapper,
+  onClickForMore,
+  totalItemsFromData,
+}: PaginationProps<T, V>) => {
+  const flattenedData = data.flatMap(dataFlattener);
+  const currentPage = Math.ceil(flattenedData.length / itemsPerPage);
+  const totalPages = Math.ceil(totalItemsFromData(data) / itemsPerPage);
```
ojeytonwilliams, line 25: This approach seems a bit abstract. All the pagination component really needs is an array of things to render, a total, `itemsPerPage`, `onClickForMore` and `displayOnEmpty`. `data`, `dataFlattener` and `mapper` could just be a single `items` array, that's the result of

```
data.flatMap(dataFlattener).map(mapper)
```

That way, the Pagination component's really focused on pagination specific stuff, not on what's being paginated.

client/src/modules/home/index.tsx
```diff
@@ -84,22 +66,35 @@ const Home = () => {
...
+                fetchMore({ variables: { offset, limit: eventsPerPage } })
+              }
+              totalItemsFromData={(data) => data[0].total}
+              displayOnEmpty={
+                <Text size="md">
+                  No more, check out the{' '}
+                  <Link href="/events" fontWeight="bold">
+                    Event page
+                  </Link>
+                  .
```
ojeytonwilliams, line 86:
```suggestion
                  to see past events.
```
It wasn't obvious to me why I should go there otherwise.

client/src/modules/events/pages/eventsPage.tsx
```diff
@@ -87,14 +32,28 @@ export const EventsPage: NextPage = () => {
...
+          }
+          totalItemsFromData={(data) => data[0].total}
+          displayOnEmpty={
+            checkInstancePermission(user, Permission.EventsView) && (
+              <Text size="md">
+                No more, you can create event in{' '}
+                <Link href="/dashboard/events" fontWeight="bold">
+                  Event dashboard
+                </Link>
+                .
```
ojeytonwilliams, line 53:
```suggestion
                No more events. Go to the{' '}
                <Link href="/dashboard/events" fontWeight="bold">
                  event dashboard
                </Link>
                to create more.
```

server/src/controllers/Events/resolver.ts
```diff
@@ -230,44 +229,36 @@ const getNameForNewAttendance = (event: EventAttendanceName) => {
 
 @Resolver()
 export class EventResolver {
-  @Query(() => PaginatedEventsWithTotal)
+  @Query(() => [PaginatedEventsWithChapters])
```
ojeytonwilliams, line 232:
```suggestion
  @Query(() => PaginatedEventsWithChapters)
```
Seems redundant since the array only has a single element.

server/src/controllers/Events/resolver.ts
```diff
@@ -230,44 +229,36 @@ const getNameForNewAttendance = (event: EventAttendanceName) => {
...
-    const events = await prisma.events.findMany({
-      include: {
-        chapter: true,
-      },
-      orderBy: [{ start_at: 'asc' }, { name: 'asc' }],
-      take: limit ?? 10,
-      skip: offset,
+    @Arg('showOnlyUpcoming', () => Boolean, { nullable: true })
+    showOnlyUpcoming = true,
+  ): Promise<PaginatedEventsWithChapters[]> {
```
ojeytonwilliams, line 238:
```suggestion
  ): Promise<PaginatedEventsWithChapters> {
```

server/src/controllers/Events/resolver.ts
```diff
@@ -230,44 +229,36 @@ const getNameForNewAttendance = (event: EventAttendanceName) => {
...
+        },
+      }),
       include: {
         chapter: true,
       },
       orderBy: [{ start_at: 'asc' }, { name: 'asc' }],
       take: limit ?? 10,
       skip: offset,
     });
+    return [{ total, events }];
```
ojeytonwilliams, line 261:
```suggestion
    return { total, events };
```

ojeytonwilliams (changes requested):
Hey @gikf @Sboonny, I like the look and feel. I think there's a few minor tweaks we can make, though:

gikf:
Thanks @ojeytonwilliams. I was very happy with my making `Pagination` as abstract :laughing:, you are right that's going a bit too far and it can be simpler.

@Sboonny let me know if you'd want me to make suggested changes to the `Pagination`.

Sboonny:
Sorry for the delay, @gikf. I don't understand the requested changes, so part of me want to change it and find out more.

But I know how obnoxious I can be, when I don't understand something. Give me tomorrow, I will try to implement it, if I can't, I will ping you.

server/src/controllers/Events/resolver.ts
ojeytonwilliams: @Sboonny I did try this locally and I'm completely at a loss as to why it fails.

The GQL query works just fine in the studio, but for some reason the client gets `data.paginatedEventsWithTotal === []`. This makes no sense to me.

Unless either of you have any clue why this is failing, I think we can just add an explanatory comment and be done with it.

gikf: That's due to query policy merging the results
https://github.com/Sboonny/chapter/blob/80f58331fee00649361e936bba9b01ad680cfc78/client/src/pages/_app.tsx#L61

For the changed resolver something like this works, and it actually merges them flattened, which is pretty cool.
```typescript
          paginatedEventsWithTotal: {
            keyArgs: ['showOnlyUpcoming'],
            merge(existing = { total: undefined, events: [] }, incoming) {
              return {
                total: existing.total ?? incoming.total,
                events: [...existing.events, ...incoming.events],
              };
            },
          },
```

ojeytonwilliams: Oh, thanks, I completely missed that!

client/src/modules/util/pagination.tsx
Sboonny: is the current Pagination good enough? 🤔

ojeytonwilliams: It is now :) I figured you'd been through enough, so I just pushed the nitpicks.

ojeytonwilliams (approved):
LGTM :+1: 

Since I pushed a bit of code, could you sign off on this @gikf ?

Also, thanks to you both for your hard work on this. I didn't expect cleaning up the Click for more button to be anywhere near this involved, but what you came up with is a big improvement.

gikf (approved):

PR #2342, opened 2023-02-05
fix: upsert user record during login

### LINKED ISSUE #1984: Server can attempt to create an existing user
```
PrismaClientKnownRequestError: 
Invalid `prisma.users.create()` invocation:


Unique constraint failed on the fields: (`email`)
    at RequestHandler.handleRequestError (/usr/chapter/node_modules/@prisma/client/runtime/index.js:34310:13)
    at RequestHandler.request (/usr/chapter/node_modules/@prisma/client/runtime/index.js:34293:12)
    at async PrismaClient._request (/usr/chapter/node_modules/@prisma/client/runtime/index.js:35273:16)
    at async /usr/chapter/server/src/app.js:91:95 {
  code: 'P2002',
  clientVersion: '4.6.1',
  meta: { target: [ 'email' ] }
}
```

I'm not sure how this happens, because the `createUser(email)` call is immediately after a `findUser(email)` check, but I think it must be two simultaneous login attempts.

TODO:

- [ ] Investigate how/why/if that is happening
- [ ] `upsert` the user record instead (it's at least more idiomatic)

There's a nice discussion https://github.com/prisma/prisma/discussions/5815#discussioncomment-404334 about how to use `upsert` as a `findOrCreate` call and how to handle race conditions.

atosh502:

Closes #1984


ojeytonwilliams (changes requested):
The code looks good!  The only thing I'd like to change is how the error is logged.

server/src/app.ts
```diff
@@ -82,17 +78,33 @@ export const main = async (app: Express) => {
...
     });
   }
 
+  async function findOrCreateUser(email: string) {
+    try {
+      return await upsertUser(email);
+    } catch (e) {
+      if (
+        e instanceof Prisma.PrismaClientKnownRequestError &&
+        e.code === 'P2002'
```
gikf, line 91: This error code has named variable in the `server/src/prisma.ts` https://github.com/freeCodeCamp/chapter/blob/bcd226b9bf5f0b14e749690c0d5a37fb0058bb74/server/src/prisma.ts#L10

server/src/app.ts
```diff
@@ -82,17 +78,33 @@ export const main = async (app: Express) => {
...
+      ) {
+        console.log(`user with ${email} already exists`);
+        return await upsertUser(email);
+      }
+      throw e;
+    }
```
gikf, line 97: I'm a bit confused about the overall logic here.

If there's error thrown during the first `upsertUser` call and the error code is `P2002`, `upsertUser` is called again. Won't that just throw error once again?

ojeytonwilliams: That's a very good question. The answer is 'no', but it's also not necessary to call `upsert` again. Reason being, `upsert` throws if two requests come in at the same time and resolve to two `createUser` calls. That can happen because there's a race between `findUser` and `createUser` (effectively that's what `upsert` is doing). If both `findUser`s resolve before the first `createUser`, then the second `createUser` will throw.

Since, the `upsert` is throwing because the `createUser` part has already resolved, we can just log that something weird happened and move on.
```suggestion
    try {
      return await upsertUser(email);
    } catch (e) {
      if (
        e instanceof Prisma.PrismaClientKnownRequestError &&
        e.code === 'P2002'
      ) {
        console.log(`user with ${email} already exists`);
      } else {
        throw e;
      }
    }
```

atosh502: > That can happen because there's a race between findUser and createUser (effectively that's what upsert is doing). If both findUsers resolve before the first createUser, then the second createUser will throw.

:100: 

@gikf [Here's an explanation from the Prisma docs](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#cause) on why the error occurs.

server/src/app.ts
```diff
@@ -82,17 +78,37 @@ export const main = async (app: Express) => {
...
     const token = getBearerToken(req);
     if (token) {
       const userInfo = fetchUserInfo(token);
       userInfo
         .then(async ({ email }) => {
           if (!email) throw Error('No email found in user info');
-          const user = (await findUser(email)) ?? (await createUser(email));
+
+          const user = await findOrCreateUser(email);
+          if (!user) throw Error('No user found');
```
ojeytonwilliams, line 111:
```suggestion
```
Sorry, I hadn't realised this would be the consequence of removing the second `upsert`. I think it's a better UX if we only allow errors we can't handle to terminate the request.

With that in mind, I'll add a suggestion in `findOrCreateUser` to remove the need to handle `undefined` here.

server/src/app.ts
```diff
@@ -82,17 +78,37 @@ export const main = async (app: Express) => {
...
+    try {
+      return await upsertUser(email);
+    } catch (e) {
+      if (
+        e instanceof Prisma.PrismaClientKnownRequestError &&
+        e.code === UNIQUE_CONSTRAINT_FAILED
+      ) {
+        console.error(
+          'Received concurrent login requests - the client is likely making too many requests.',
+        );
```
ojeytonwilliams, line 95:
```suggestion
        );
        // Since the error was a unique constraint violation, we know that the
        // user exists, so we can just return their record.
        return await prisma.users.findUniqueOrThrow({ where: { email } });  
```
This _should_ be more efficient because it's definitely a single query.  Also, it's more declarative since we know that we would never be inserting a new record.

ojeytonwilliams (changes requested):

ojeytonwilliams (approved):
LGTM :+1:

PR #2350, opened 2023-02-07
feat: remove required fields from not required fields

Sboonny:

Event is now can be created without these fields, unless I missed something 🤔, so we can remove them


gikf (commented):
Capacity is definitely still required in the server, and partially in client - the validation won't let empty capacity.

Venue type and venue(_id) are a bit funny. Technically they always will have some value, so changing anything for them is indeed not required. However if somehow they'd go missing from the submit data, server side would error, I believe.

Sboonny:
Do we remove capacity, I haven't checked, but people should join an event even with capacity of 1🤔

gikf:
They still can _opt_ for attending when event is at full capacity, but they will be placed on the waitlist.

PR #2351, opened 2023-02-07
feat: stop the card flashing in eventCard

Sboonny:

Without `onError` the eventCards will fallback to the image, while we need it to fallback if there is no image

### DIFF
--- a/client/src/components/EventCard.tsx
+++ b/client/src/components/EventCard.tsx
@@ -95,7 +95,10 @@ export const EventCard: React.FC<EventCardProps> = ({ event }) => {
         <Image
           width="100%"
           height="100%"
-          src={event.image_url}
+          src={
+            event.image_url ||
+            'https://cdn.freecodecamp.org/chapter/brown-curtain-small.jpg'
+          }
           fit="cover"
           fallbackSrc="https://cdn.freecodecamp.org/chapter/brown-curtain-small.jpg"
           fallbackStrategy="onError"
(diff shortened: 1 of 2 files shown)


client/src/components/EventCard.tsx
```diff
@@ -98,6 +98,7 @@ export const EventCard: React.FC<EventCardProps> = ({ event }) => {
           src={event.image_url}
```
gikf, line 98:
```suggestion
          src={
            event.image_url ||
            'https://cdn.freecodecamp.org/chapter/brown-curtain-small.jpg'
          }
```

gikf (commented):
Could you post steps for reproducing the issue, or what is solved? After I've removed event _Event Image Url_, there's broken image displayed.
image

Maybe when `!event.image_url` then fallback image should be used as the image `src`. Suggestion for that below.

Sboonny:
[rec-tab.webm](https://user-images.githubusercontent.com/88248797/217737092-712a4412-ea65-4e32-8c01-28cd06e27131.webm)

The issue is the fallback is shown while the image is lazy loading

client/src/components/EventCard.tsx
Sboonny: Works 👍, nice call

Sboonny:
Did you remember the [image falling gracefully](https://discord.com/channels/692816967895220344/988495800852152450/1033992660559609896) discussion in discord, and made suggestion based on it? 🤣 

I have forgot about it completely

gikf:
Thanks, this makes sense. The `fallbackSrc` and `fallbackStrategy` still might be useful though, for cases when image_url is provided, but cannot be loaded.

gikf (approved):

PR #2352, opened 2023-02-07
feat: add fieldset and legend in eventForm for screen readers

Sboonny:

Screen Readers may need additional context of what does checking `invite-only` does, for now they will know it's the type of the event


gikf:
_Event Type_ sounds a bit weird to me when describing invite-only event. Not that this isn't technically correct.

I don't have idea for what that could be changed. What came to my mind is grouping the _Invite only_ checkbox together with the _Capacity_ and call them both _Attendee restrictions_ or _Event restrictions_. What do you think?

Sboonny:
Sorry, I have missed the comment while I was sick, I am happy with adding the changes.

Just to add another option, this change is purely for screen readers, if it isn't visually pleasing, we can add a legend with too many details and set it as visibility hidden.

gikf (approved):

PR #2354, opened 2023-02-07
chore: check integrationStatus before trying to use calendar

gikf:

---

- Adds check of the `integrationStatus` before using calendar.
- Change isn't that big as it might look like, external api wouldn't be actually used anyway, but I find more sensible to do the check first, instead of letting the service call error.


ojeytonwilliams (approved):
Thanks, @gikf, I think it makes sense to put the burden on our db rather than sending requests  to Google that we know will fail.

LGTM :+1:

PR #2363, opened 2023-02-10
fix: prevent the same start_at and ends_at

gikf:

---

- Event form doesn't allow `ends_at` to be the same as `start_at`.
- Therefore when test is trying to edit event having them the same, without changing dates - it will fail.


Sboonny (approved):

PR #2371, opened 2023-02-13
feat: warn when calendar or calendar event wasn't created

gikf:

---

- Displays basic warning when during chapter and event creation, calendar api didn't create calendar or calendar event.
- Adds `has_calendar_event` custom `FieldResolver` on `Event` graphql-type.


ojeytonwilliams:
Hey @gikf, I like the `has_calendar_event` field, that's much nicer than abusing the ids.  My only concern is when to show the warnings.

Warning that the Chapter does not have a calendar seems good. People may forget to authenticate with Google, after all. Also, only instance owners will see this warning and it's reasonable to expect them to have understand how Chapter works.

However, I'm not sure the warning 'Calendar event was not created.' is appropriate unless there's a calendar. If someone chooses not to make use of the integration (due to rate limits, say, https://github.com/freeCodeCamp/chapter/issues/2373) then showing warnings on every event creation is likely to make people think something has gone wrong with Chapter.

If that makes sense, I think the simplest implementation would be to pull `useChapterQuery` out of the EventForm and into the page components. That's a better separation of concerns, too, since the form shouldn't really have to worry about data fetching.

gikf:
Good ideas @ojeytonwilliams!

client/src/modules/dashboard/Events/pages/NewEventPage.tsx
```diff
@@ -60,6 +69,17 @@ export const NewEventPage: NextPageWithLayout<{
...
         status: 'success',
       });
+
+      let hasChapterCalendar = chapterData?.dashboardChapter.has_calendar;
+      if (chapter_id !== chapterId) {
+        const { data: currentChapter } = await refetch({
+          chapterId: chapter_id,
+        });
+        hasChapterCalendar = currentChapter.dashboardChapter.has_calendar;
+      }
```
ojeytonwilliams, line 79: The fact this is necessary, got me to thinking: why not use the same source of truth for chapter name when on `dashboard/chapters/[id]/new-event` and `dashboard/events/new`? Namely `admined_chapters`.

Reason being, we only care about three things: `id`, `name` and `has_calendar`. If `admined_chapters` gets `has_calendar` we don't need to make (let alone pass around) an additional query. As a side benefit, it should be possible to simplify `NewEventPage` a little.

What do you reckon?

ojeytonwilliams (approved):
Sorry for the delay getting to this, @gikf 

Everything looks great, though, cheers!

PR #2390, opened 2023-02-17
feat: delete the test site alert from the pages

Sboonny:

The site is no longer testing site, because  it's moved to .org, so the alert can go

### DIFF
--- a/client/src/components/PageLayout/index.tsx
+++ b/client/src/components/PageLayout/index.tsx
@@ -30,14 +30,6 @@ const PageLayout = ({ children }: { children: React.ReactNode }) => {
   return (
     <>
       <Header />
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (changes requested):
Could you make a couple of changes?

* Make this conditional on `process.env.NEXT_PUBLIC_DEPLOYMENT_ENVIRONMENT === 'staging'` (same as 
and the `noindex` `meta` element).
* change .env.example to have `NEXT_PUBLIC_DEPLOYMENT_ENVIRONMENT=production`

ojeytonwilliams:
Blocked until we officially release, since we may need to purge the db before then.

ojeytonwilliams (commented):
LGTM :+1:

PR #2396, opened 2023-02-18
fix: ensure more consistent events ordering

gikf:

---

- Ensures more consistent ordering of events.
- For some reason events with the same start date not always were ordered the same way.
- This was creating issues for one of the tests, which uses the first event in the dashboard. If after editing that event, while keeping date without changes, order would be flipped with another event having the same date, the test would fail.
- If needed I can also augment mentioned test, but as far as test goes, chances to have seeded event with the same date and name is way much closer to none.


Sboonny (approved):

PR #2397, opened 2023-02-18
fix: wait for two mails

gikf:

---

- When attending there are two emails send - one to the attendee and second one to the admin, in that order.
- Test is interested in the second of them, but was waiting only for single email. In most cases the email to admin must have been send in the meantime, so it was already there when was specifically looked for. In other rare cases test was failing.


Sboonny (approved):

PR #2412, opened 2023-02-22
feat: count chapter users in db

gikf:

---
- Uses `_count` from prisma to obtain number of chapter users.
- Changes returned object for `chapter` query.
- Fixes couple refetch issues in the chapter <-> event relation, mainly concentrating on counted users, but not only.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -959,7 +959,7 @@
               "name": null,
               "ofType": {
                 "kind": "SCALAR",
-                "name": "Float",
+                "name": "Int",
                 "ofType": null
               }
             },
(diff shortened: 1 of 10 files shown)


client/graphql.schema.json
```diff
@@ -953,6 +945,33 @@
...
+            "name": "chapter_users",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Float",
+                "ofType": null
```
Sboonny, line 963: Noob question, do you know why this type of data is float, and not int?

and will we have to deal with a user in the other side of the decimal? and how? 🤕

gikf: That's because I used `Number` instead of `Int` in the graphql type.

Sboonny (approved):

PR #2420, opened 2023-02-25
chore: replace toast with alert

### LINKED ISSUE #2107: Alternatives/improvements to toasts

Typically in freeCodeCamp UIs we do not use toasts. From talking to @QuincyLarson the main concern is a11y. We want to make sure that all users have time to read the information and that screenreaders announce it.

From my testing, our current toasts are most of the way there, but leave a few things to be desired.

1. They disappear quite quickly (this is easy enough to configure, though)
2. They do not appear in the tab hierarchy, so it's unclear how a keyboard user would manually dismiss them

Are there superior alternatives to give this kind of instant feedback?  Or, could the current toasts do the job if we make them last longer, fit them into the focus order and ensure they can be dismissed via the keyboard?

@ahmadabdolsaheb @bbsmooth this is a little outside of my comfort zone, so any insight would be very welcome.

---

If anyone wants to quickly test out the toasts feel free to join/leave https://chapter.freecodecamp.dev/chapters/4.  The annoying repetition of the "This is a testing site" message is because we implemented that as an alert. It's not ideal, but it's not the fault of the toasts.

gikf:

TODO
- [ ] Check multi-step tests and add closing alerts, if they could affect the outcome.

(Maybe) Closes #2107

---

- Ground work for replacing `toast` with ~~`alert`~~ `addAlert`. I might be a bit careful, but using name of built-in function could lead to some interesting mistakes.
- I've changed all toasts to alert in `events/{id}`, so feel free to take a look.


Sboonny (approved):

PR #2423, opened 2023-02-27
feat: only warn about missing keys

ojeytonwilliams:
- chore: use colours to better differentiate logs
- chore: downgrade oauth2.keys.json error to warning

The main thing is we can't rely on `oauth2.keys.json` existing. Notably it won't exist in [railway](https://railway.app/), so we need an alternative approach.

For now, I just want to downgrade the error to a warning.  Longer term we could make use of environment variables, but I'm focusing on getting the MVP deployable for now.


ojeytonwilliams:
I added the colours to (hopefully) make it easier to read the logs that concurrently produces. If anyone hates them, feel free to revert that change.

PR #2425, opened 2023-02-28
feat: add instance settings to schema

### LINKED ISSUE #2408: Self-hosted privacy policy, terms of service and code of conduct
People who self-host chapter will need a way to provide their own versions of these.

As @allella mentioned https://github.com/freeCodeCamp/chapter/issues/2373#issuecomment-1438686164 the simplest approach would be to leave them as links, but allow them to be customised.

There are two main tasks

- [ ] Persist the URLs
- [ ] Use the URLs instead of the hard-coded values

While the best UX is _probably_ to provide a GUI to customise the variables (and persist them in the database) the simplest approach (sufficient for the MVP) would be to store them in environment variables.

Sboonny:

Closes #2408


server/prisma/seed/seed.ts
```diff
@@ -52,7 +53,7 @@ async function seed() {
     sponsorIds,
     15,
   );
-
+  await createInstanceSettings();
```
Sboonny, line 56:
```suggestion
  await createInstanceSettings();
  
```

server/prisma/schema.prisma
```diff
@@ -11,6 +11,15 @@ datasource db {
   url      = env("DATABASE_URL")
 }
 
+model instance_settings {
+  id                    Int    @id
```
gikf, line 15:
```suggestion
  id                    Int     @id @default(autoincrement())
```
Only thing I can think of, without trying to go into how it would be used.

Sboonny: Wait, I was planning to have one instance, but we can keep the change if we have save state for multiple instance setting 🤣

server/prisma/seed/factories/instance_settings.factory.ts
```diff
@@ -0,0 +1,21 @@
+import { faker } from '@faker-js/faker';
+import { Prisma } from '@prisma/client';
+
+import { prisma } from '../../../src/prisma';
+
+const { lorem } = faker;
+
+const createInstanceSettings = async () => {
+  const instanceSettingsData: Prisma.instance_settingsCreateInput = {
+    id: 1,
```
ojeytonwilliams, line 10:
```suggestion
```
This won't be necessary with the above schema.

ojeytonwilliams (changes requested):

server/prisma/migrations/20230324125932_create_instance_settings/migration.sql
```diff
@@ -0,0 +1,14 @@
+-- CreateTable
+-- We aren't generating ids automatically, because the instance will use one record.
+-- But we allow for multiple record, because we can add the ability for instance owner,
+-- to save multiple settings, and use them as templates.
+CREATE TABLE "instance_settings" (
+    "id" SERIAL NOT NULL,
```
ojeytonwilliams, line 6: Ah, that's a little different. If the plan is to have multiple records, then we don't need to do anything special. It's fine to use `@default(autoincrement())` and there's nothing weird that demands explanation.
```suggestion
CREATE TABLE "instance_settings" (
    "id" SERIAL NOT NULL,
```

ojeytonwilliams (changes requested):
Thanks for adding the comment @Sboonny, it helped me understand the requirements.  Ironically, that means we don't need it!

ojeytonwilliams:
Sorry for the delay on this one, but I think it's a bad idea to update the database without including the code that makes use of those updates. Reason being I can't follow the logic about how this is used. Also, more generally each PR should "do one thing", but this PR "does half a thing". I hope that makes sense!

Could you combine this PR with https://github.com/freeCodeCamp/chapter/pull/2467 ?

PR #2434, opened 2023-03-03
feat: lazy load the chapters

Sboonny:
- [x]I have read [Chapter's contributing guidelines](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md).

Lazy load the chapters in the home page, to make it possible to interact with the events while chapters are loading

### DIFF
--- a/client/src/modules/home/index.tsx
+++ b/client/src/modules/home/index.tsx
@@ -1,5 +1,5 @@
 import { Heading, VStack, Grid, GridItem, Flex, Text } from '@chakra-ui/react';
-import React from 'react';
+import React, { Suspense, useEffect } from 'react';
 import { Link } from 'chakra-next-link';
 
 import { Loading } from '../../components/Loading';
@@ -7,7 +7,7 @@ import { ChapterCard } from '../../components/ChapterCard';
 import { EventCard } from '../../components/EventCard';
 import {
   usePaginatedEventsWithTotalQuery,
-  useChaptersQuery,
+  useChaptersLazyQuery,
 } from '../../generated/graphql';
 import { Pagination } from '../util/pagination';
 import { UserContextType, useUser } from '../auth/user';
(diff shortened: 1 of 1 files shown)


client/src/modules/home/index.tsx
```diff
@@ -41,15 +41,26 @@ const Welcome = ({ user }: { user: User }) => {
...
-  const { error: chapterError, data: chapterData } = useChaptersQuery();
+  const [eventsLoaded, setEventsLoaded] = useState(false);
+  const [getChapters, { error: chapterErrors, data: chapterDatas }] =
+    useChaptersLazyQuery();
 
   const { loading, error, data, fetchMore } = usePaginatedEventsWithTotalQuery({
     variables: { offset: 0, limit: eventsPerPage },
+    onCompleted: () => {
+      setEventsLoaded(true);
+    },
```
gikf, line 52:
```suggestion
```

client/src/modules/home/index.tsx
```diff
@@ -41,15 +41,26 @@ const Welcome = ({ user }: { user: User }) => {
...
   });
   const { user } = useUser();
 
-  const isLoading = loading || !data || !chapterData;
-  if (isLoading) return <Loading error={error || chapterError} />;
+  useEffect(() => {
+    if (eventsLoaded && !chapterDatas) {
+      getChapters();
+    }
+  }, [eventsLoaded]);
```
gikf, line 60:
```suggestion
    getChapters();
  }, [loading]);
```

gikf (commented):
I've noticed the issue with warning was still occurring. More tinkering lead me to suggestions that this specific warning might have been false positive.

Nevertheless it also turned out the whole thing can be done in a simpler manner. The `loading` variable doesn't change when fetching more.

client/src/modules/home/index.tsx
gikf: I might have glitched :thinking:. This, double checked, should start loading chapters after events are loaded:
```suggestion
    if (!loading) {
      getChapters();
    }
  }, [loading]);
```

gikf (approved):
LGTM :+1:

PR #2438, opened 2023-03-04
feat: increase the min height for content in laptops layout

Sboonny:

Screens with width between 1000px and 1440px have the footer away from the bottom of the layout.

image

### DIFF
--- a/client/src/components/PageLayout/index.tsx
+++ b/client/src/components/PageLayout/index.tsx
@@ -53,7 +53,7 @@ const PageLayout = ({ children }: { children: React.ReactNode }) => {
       <SkipNavContent />
       <Box
         as="main"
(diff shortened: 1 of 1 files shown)


Sboonny:
the more that I interact with it, the more that I hate `min-height` solution, the reason I have it, because body has `#__next` as a child, so I can't use `body` as a way to set the layout.

If you want, I can change `pagelayout` to have two extra elements that can create grid layout.

ojeytonwilliams (approved):
I'm happy with whichever solution you prefer. This already looks better, so I'm okay with merging, but if you want to work on it more that's cool.

Sboonny:
After turning it to grid, it looked more consistent, but it didn't sort what I desire.

Lets push the current change, and if someone wants a different way, I have already did it so adding it would be easier

PR #2443, opened 2023-03-06
fix: ChapterCard displays past events after visiting chapter page

### LINKED ISSUE #2440: Past events displayed in chapter card after chapter page visit
### Describe the bug:
Example for reproducing:

1. Specific chapter has all events in the past - no _Next Events_ displayed in the chapter card on home page or `/chapters`.
1. Navigate to that chapters public page, see that chapter has some events.
1. Navigate back to page with chapter card. Past events are in the chapter card as _Next Events_, labelled as _Running_.

gikf:

Closes #2440

---

- Filters out past events in the chapter card.
- Also modifies what graphql-type is returned by chapter query. It's right now not using `_count`, so no need for it.

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -5276,7 +5276,7 @@
               "name": null,
               "ofType": {
                 "kind": "OBJECT",
-                "name": "ChapterCardRelations",
+                "name": "ChapterWithEvents",
                 "ofType": null
               }
             },
--- a/client/src/components/ChapterCard.tsx
+++ b/client/src/components/ChapterCard.tsx
@@ -124,37 +124,39 @@ export const ChapterCard: React.FC<ChapterCardProps> = ({ chapter }) => {
             Next Events
           </Text>
           <GridItem area="event" paddingInline={'1em'}>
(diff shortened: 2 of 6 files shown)


Sboonny (commented):
Is this the fix for this bug? 


[rec-tab (14).webm](https://user-images.githubusercontent.com/88248797/223191355-9472c657-6449-4f3f-a4d3-645e976acd8f.webm)

---

Edit: I am going to review it in depth tomorrow 👍

gikf:
Yes, it addresses that in two ways.

Sboonny (approved):
I have tried hard to create a custom field for chapters, that stop the caching data from `/chapter` to home page, but it always results in a buggy behavior, either:

- the data of past event showing in the homepage
- the data of the past events hidden in the `/chapter`


Although I am unhappy (edit: I am unhappy with apollo, not with the solution) because I think the caching still happens and get filtered, and it shouldn't happen, and the pages shouldn't affect each other, so far this is the best solution we currently have.


https://www.apollographql.com/docs/react/caching/cache-field-behavior/#merging-arrays-of-non-normalized-objects

PR #2450, opened 2023-03-07
fix: refetch after event actions

gikf:

---
- Fixes updating `/events` after edit, canceling or deletion (including chapter deletion).
- Fixes updating events displayed in chapter card after event deletion.
- Unfortunately queries having distinct arguments (like `showOnlyUpcoming`) need refetches with argument combinations.


ojeytonwilliams:
> Unfortunately queries having distinct arguments (like showOnlyUpcoming) need refetches with argument combinations.

That is definitely unfortunate. I guess it makes sense since they'll likely have different results.  Still a pain, though!

ojeytonwilliams (approved):
LGTM :+1:

PR #2464, opened 2023-03-12
chore: add grid to event actions on eventPage

gikf:

---

- Adds grid to event actions on eventPage to slightly unify the look with chapterPage.

### DIFF
--- a/client/src/modules/events/pages/eventPage.tsx
+++ b/client/src/modules/events/pages/eventPage.tsx
@@ -367,7 +367,7 @@ export const EventPage: NextPage = () => {
               paddingBlock="1"
               paddingInline="2"
             >
-              {data.event.invite_only ? 'Request' : 'Attend'}
+              {data.event.invite_only ? 'Request Invite' : 'Attend Event'}
             </Button>
           ) : (
             <>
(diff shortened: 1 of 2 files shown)


client/src/modules/events/pages/eventPage.tsx
```diff
@@ -353,72 +354,77 @@ export const EventPage: NextPage = () => {
...
-        {userEvent && (
-          <HStack>
-            {userEvent.subscribed ? (
-              <>
-                <Text fontSize={'md'} fontWeight={'500'}>
-                  You are subscribed to event updates
+          ) : (
+            <>
+              {attendanceStatus === 'waitlist' ? (
+                <Text fontSize="md" fontWeight="500">
```
Sboonny, line 373: these styles are different from the other text, which weirding me out.

but this isn't related to the original issue, I am happy with current changes, if you think this deserves to be in another PR.

```suggestion
                <Text>
```

Sboonny (approved):

client/src/modules/events/pages/eventPage.tsx
gikf: I suppose there's couple more inconsistencies between the text - ie. text by subscribe/unsubscribe buttons have such styling, so I'd say separate PR would be better to get them all together. And to figure out whether the styling is needed or not.

PR #2479, opened 2023-03-15
feat: chapter subscription opt-in

### LINKED ISSUE #2273: Opt in to Chapter notifications when attending first event
The first time a user tries to attend an event in a new chapter, they're given the following warning

> Join this event?
> Note: joining this event will make you a member of the event's chapter.

New users do not necessarily know what a chapter is or membership of a chapter comes with a subscription.  That means it is unclear exactly what this warning is about.

Since the relevant thing they're opting into is notifications of subsequent events, we should make that explicit.  Also, it should actually be opt in.  Since https://github.com/freeCodeCamp/chapter/pull/2103 `auto_subscribe` is true, which I think is fine for Chapter pages, since the user immediately sees the Unsubscribe button. However, this is different since it's unreasonable to expect someone to intuit what joining a chapter means and we don't offer an obvious way to unsubscribe. 

### Possible fix

- Replace the warning with unchecked checkbox "Subscribe to subsequent events" and only subscribe them to the Chapter if they check it.

### Downsides

- This makes it more likely that people accidentally opt out of something they would want
- It's an additional annoyance to users who don't forget - they have to click the box or go to the chapter to subscribe
- auto_subscribe's function becomes more specialised and less obvious

Perhaps we can revert https://github.com/freeCodeCamp/chapter/pull/2103 (again) simply not show the checkbox if they set auto_subscribe to true.

In addition, if we put "You're unsubscribed" warnings on the chapter page and it's event pages, then people who are unsubscribed should realise and, if they want to, can easily change their mind.

Finally, alongside the warning, we should put a subscribe to chapter button on the event page.

gikf:

Closes #2273 

---

- Adds optional `subscribe` parameter to `joinChapter` mutation.
- Adds checkbox allowing subscribe or not to chapter in case of not being chapter member already. It will be displayed only if:
  - User is logged in
  - User is not chapter member
- I'd say displaying it also for not logged users is currently not feasible. Not only due to potentially showing it to somebody already being chapter member, but also because of increased complexity that would entail right now. I think ideally in such case, `chakra-confirm` would need to be replaced partially, or completely by new component.


Sboonny (approved):

PR #2480, opened 2023-03-16
fix: update chapterPage after login

### LINKED ISSUE #2476: Chapter page doesn't update after login
### Describe the bug:
1. While logged out, navigate to chapter page `chapters/{id}`. Action buttons (ie. _Join chapter_) are not visible.
2. Press login.
3. After login action buttons are still not visible.

chapterPage uses `chapterUser` query, which doesn't get refetched on login. It returns `chapterUser` for a specific `chapterId`, so it might not be too feasible to refetch all _chapter users_ for user. Some option could be using `user_chapters` in the `meQuery`, however that needs at least modifying query to get name of the chapter role.

gikf:

Closes #2476

### DIFF
--- a/client/graphql.schema.json
+++ b/client/graphql.schema.json
@@ -5357,35 +5357,6 @@
             "deprecationReason": null
           },
           {
-            "name": "chapterUser",
-            "description": null,
-            "args": [
-              {
-                "name": "chapterId",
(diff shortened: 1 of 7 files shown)


Sboonny (approved):

PR #2516, opened 2023-03-25
chore: use permission to find chapter admin

gikf:

### DIFF
--- a/server/src/controllers/Events/resolver.ts
+++ b/server/src/controllers/Events/resolver.ts
@@ -348,12 +348,17 @@ export class EventResolver {
       },
     });
 
-    // TODO: find by permission, not role.
     const chapterAdministrators = await prisma.chapter_users.findMany({
       where: {
         chapter_id: chapterId,
         subscribed: true,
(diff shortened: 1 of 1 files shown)


server/src/controllers/Events/resolver.ts
```diff
@@ -348,12 +348,17 @@ export class EventResolver {
...
       where: {
         chapter_id: chapterId,
         subscribed: true,
-        chapter_role: { is: { name: 'administrator' } },
+        chapter_role: {
+          chapter_role_permissions: {
+            some: {
+              chapter_permission: { name: 'chapter-edit' },
+            },
+          },
```
Sboonny, line 360: random question, I haven't tested it yet 😞, does this get the instances_owners too, they should have `chapter-edit` as well.

Do we need to add `NOT` that clear the owners 🤔

gikf: As queried is `chapter_users` table, it will not include instance owner. Unless they have explicitly, for specific chapter, chapter role with `chapter-edit` permission.

Sboonny (approved):

PR #2545, opened 2023-03-31
feat: use transform in jest config

Sboonny:

With the new jest it's moving to transform, instead of global. this add its config

https://jestjs.io/docs/next/code-transformation#writing-custom-transformers

Ref: #2544

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -3,10 +3,13 @@ module.exports = {
   testEnvironment: 'node',
   roots: ['<rootDir>'],
   moduleDirectories: ['node_modules', 'server'],
-  globals: {
-    'ts-jest': {
-      tsconfig: './tsconfig.test.json',
-    },
+  transform: {
+    '.(ts|tsx)': [
+      'ts-jest',
+      {
+        tsconfig: './tsconfig.test.json',
+      },
+    ],
   },
   watchPathIgnorePatterns: ['/node_modules'],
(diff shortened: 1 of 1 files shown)


jest.config.js
```diff
@@ -3,10 +3,13 @@ module.exports = {
   testEnvironment: 'node',
   roots: ['<rootDir>'],
   moduleDirectories: ['node_modules', 'server'],
-  globals: {
-    'ts-jest': {
-      tsconfig: './tsconfig.test.json',
-    },
+  transform: {
+    '.(ts|tsx)': [
```
Sboonny, line 7: Btw, I don't know the difference in regex, but here is the one in ts-jest docs https://kulshekhar.github.io/ts-jest/docs/getting-started/options

```suggestion
    '^.+\\.tsx?$': [
```

ojeytonwilliams (changes requested):
What's the motivation for this change? I ask because the tests are working in both configurations.

If this change is going to be necessary at some point, we should wait until it is. On the other hand, if it lets us make improvements, then they should be included. Otherwise it's unclear what this is for.

Sboonny:
This change and the PR of jest were added to solve the goose hunt of why Chakra PR isn't passing <#2520>, TD;LR, it didn't help, even with the current change. 😞 

I happy to close this and revert the jest PR

PR #2549, opened 2023-04-01
Revert "chore(deps): update jest and ts-jest to v29 (major)"

Sboonny:
Reverts freeCodeCamp/chapter#2544

I have added this package, because we couldn't merge chakra new monorepo, thankfully gikf found a way around it. so the upgrade to v29 isn't needed anymore, we can revert it if wanted


ojeytonwilliams:
Hey @Sboonny it's fine to be on the latest version of Jest, so there's no need to downgrade.

Sboonny:
Awesome 👍

PR #2557, opened 2023-04-04
feat: ability to unlink calendar event

gikf:

Ref #2168

---
- Adds ability to unlink calendar event after _test_.
- Test is a bit a lie. It doesn't check access to calendar event directly. It lists events from the chapter calendar (via the `calendar_id`), and checks, if `calendar_event_id` is among events for the chapter calendar. There are three possible results:
  - No matching event - calendar event either was deleted or (assumed) there's no access to it. Unlinking is possible.
  - Matching event - calendar event exists and there _should_ be access to it, as we have access to chapter calendar. No unlinking.
  - Events can't be listed - this likely means there's no access to chapter. Right now this case is thrown together with other unexpected errors. No unlinking.
- I thought about just _getting_ event instead. While it might prove complete lack of access, I'm unsure it could prove that chapter has full access to calendar event.


ojeytonwilliams (approved):
LGTM :+1:

Issue #2379, opened 2023-02-14
Waitlist when event is full


gikf:
Few tasks related to events at full capacity.

```[tasklist]
- [x] Send different email when user is placed on the waitlist, than when is added to attendees. (`attendEvent` mutation)
- [x] Don't add user to the calendar event, if user is placed on the waitlist. (`attendEvent`)
- [ ] ~~Move user from the waitlist~~, if attendance of another user is deleted. Send email about being off the waitlist. (`deleteAttendee`)
- [ ] ~~Add user to the calendar event~~, when moved from the waitlist to attendees, in case of another user cancelling attendance. Send email about being off the waitlist. (`cancelAttendance`)
```

PR #2563, opened 2023-04-07
feat: distinct emails for possible attendances

gikf:

Ref #2379

---
- Adds email for invite only event and when user _attends_ to full event.
- Fixes cause for https://github.com/freeCodeCamp/chapter/actions/runs/4637113197/jobs/8206065493 (Actually unsubscribed event was different than checked, difference was due to sending email only when user is added to attendees.)

### DIFF
--- a/server/src/util/event-email.ts
+++ b/server/src/util/event-email.ts
@@ -7,6 +7,7 @@ import {
   dateChangeText,
   eventAttendanceCancelationText,
   eventAttendanceConfirmationText,
+  eventAttendanceRequestText,
   eventAttendeeToWaitlistText,
   eventCancelationText,
   eventConfirmAtendeeText,
@@ -16,6 +17,7 @@ import {
   eventNewAttendeeNotificationText,
   eventUnsubscribeText,
   eventUpdateText,
+  eventWaitlistConfirmationText,
   physicalLocationChangeText,
   physicalLocationShortText,
   SPACER,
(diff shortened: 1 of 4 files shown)


server/src/util/event-email.ts
```diff
@@ -325,6 +327,38 @@ export const eventAttendanceConfirmation = ({
...
+}
+
+export const eventWaitlistConfirmation = ({
+  event,
+  userName,
+}: WaitlistConfirmation) =>
+  withUnsubscribe(
+    eventWaitlistConfirmationText({
+      eventName: event.name,
+      userName: userName ? ` ${userName}` : '',
```
Sboonny, line 342: the type is always `true`, and they will be used when there is a user, the if statement can go, unless I am missing something

```suggestion
      userName: ` ${userName}`,
```

server/src/util/event-email.ts
```diff
@@ -325,6 +327,38 @@ export const eventAttendanceConfirmation = ({
...
+}
+
+export const eventAttendanceRequest = ({
+  event,
+  userName,
+}: AttendanceRequest) =>
+  withUnsubscribe(
+    eventAttendanceRequestText({
+      eventName: event.name,
+      userName: userName ? ` ${userName}` : '',
```
Sboonny, line 358:
```suggestion
      userName: ` ${userName}`,
```

server/src/util/event-email.ts
gikf: Nah, it can be empty, _anonymous_ is only displayed in client. Resolver takes user directly from db, or from ctx, both don't have empty string replaced with _anonymous_.

As a side note - I'd be a bit surprised if page on which I didn't give anything as my name would start email with _Hi anonymous_ :slightly_smiling_face:

Sboonny (approved):

PR #2569, opened 2023-04-08
feat: add warning message in case there are people on the waiting list

### LINKED ISSUE #2407: Add warning when canceling the events about the waitlist
When a user leaves an event which has a waiting list, someone else will join the event in their stead, I don't think this is an issue, just warning about it happening should be enough, if there is a waitlist.

https://user-images.githubusercontent.com/88248797/220564704-57d3d355-0e8f-4bd9-9551-76e401f626b5.webm

Including a body, if there is a waitlist in the section below, should do the trick.

https://github.com/freeCodeCamp/chapter/blob/45be96dc812617511c7c0cd33feef0068b872cea/client/src/modules/events/pages/eventPage.tsx#L137-L139

---

To get set up locally, take a look at our [contribution docs](https://github.com/freeCodeCamp/chapter/blob/main/CONTRIBUTING.md). And, if you'd like to discuss more, you can head on over to [Discord](https://discord.gg/QbQd7BpaaH) and say hi, or comment here if you prefer.

Ismailtlem:

Closes #2407

### DIFF
--- a/client/src/modules/events/components/Actions.tsx
+++ b/client/src/modules/events/components/Actions.tsx
@@ -78,6 +78,9 @@ export const Actions = ({
   const userEvent = user?.user_events.find(
     ({ event_id }) => event_id === eventId,
   );
+  const waitlist = event_users.filter(
+    ({ attendance }) => attendance.name === AttendanceNames.waitlist,
+  );
 
   const attendanceStatus = eventUser?.attendance.name;
 
@@ -145,6 +148,10 @@ export const Actions = ({
   async function onCancelAttendance() {
     const ok = await confirm({
       title: 'Are you sure you want to cancel your attendance?',
+      ...(waitlist.length > 0 &&
+        !inviteOnly && {
+          body: 'The next person on the waitlist will take your seat.',
+        }),
     });
 
     if (ok) {
@@ -152,7 +159,6 @@ export const Actions = ({
         await cancelAttendance({
           variables: { eventId },
         });
-
         addAlert({ title: 'You canceled your attendance 👋', status: 'info' });
       } catch (err) {
         addAlert({ title: 'Something went wrong', status: 'error' });


client/src/modules/events/components/Actions.tsx
```diff
@@ -143,16 +146,22 @@ export const Actions = ({
...
-    const ok = await confirm({
+    const cancelMessage: {
+      title: string;
+      body?: string;
+    } = {
       title: 'Are you sure you want to cancel your attendance?',
-    });
+    };
+    if (waitlist.length > 0) {
+      cancelMessage.body = 'The next person on the list will take your seat';
```
gikf, line 156:
```suggestion
      cancelMessage.body = 'The next person on the waitlist will take your seat';
```

client/src/modules/events/components/Actions.tsx
```diff
@@ -143,16 +146,22 @@ export const Actions = ({
...
   async function onCancelAttendance() {
-    const ok = await confirm({
+    const cancelMessage: {
+      title: string;
+      body?: string;
+    } = {
       title: 'Are you sure you want to cancel your attendance?',
-    });
+    };
+    if (waitlist.length > 0) {
```
gikf, line 155:
```suggestion
    if (waitlist.length > 0 && !inviteOnly) {
```
There's one edge case - invite only event, in which case text shouldn't be displayed. For invite only event it's also not guaranteed canceling user is already off the waitlist at this point.

client/src/modules/events/components/Actions.tsx
```diff
@@ -143,16 +146,22 @@ export const Actions = ({
...
+      cancelMessage.body = 'The next person on the list will take your seat';
+    }
+    const ok = await confirm(cancelMessage);
```
gikf, line 158:
```suggestion
    const ok = await confirm({
      title: 'Are you sure you want to cancel your attendance?',
      ...(waitlist.length > 0 &&
        !inviteOnly && {
          body: 'The next person on the waitlist will take your seat.',
        }),
    });
```
Creating `cancelMessage` with optional `body` property, and then conditionally adding it separately is a bit verbose. What do you think about conditionally including it in this way?

Ismailtlem: Agree, cleaner this way

Sboonny (approved):

PR #2574, opened 2023-04-09
feat: display user joined date in event dashboard

gikf:

---
- Display event_user _joined_date_ in event dashboard.
- Order of the _joining_ can be important for the waitlist processing order.
- Displaying it is intended for the event administrator processing invites, or moving from (overcrowded) attendees back to waitlist.
- It's a bit stretch to be using _joined_date_ with the current name for this purpose. But that's the only existing field that could be used right now.
  - _created_at_ - doesn't change, no matter how many times user will cancel their attendance.
  - _updated_at_ - is changed event if user changes just subscription status. 
  - _joined_date_ - initially starts the same as _created_at_. PR adds updating it every time user _attends_ keeping the track of the last time. So ie. attend at Monday -> cancel at Friday -> attend again at Sunday. Won't push user in the line before those wanting to attend between Monday and Sunday.
- Any particular date format requests?

### DIFF
--- a/client/src/modules/dashboard/Events/pages/EventPage.tsx
+++ b/client/src/modules/dashboard/Events/pages/EventPage.tsx
@@ -283,7 +283,7 @@ export const EventPage: NextPageWithLayout = () => {
                 <DataTable
                   title={`${title}: ${users.length}`}
                   data={users}
-                  keys={['user', 'role', 'action'] as const}
+                  keys={['user', 'joined', 'role', 'action'] as const}
                   emptyText="No users"
                   mapper={{
                     user: ({ user }) => <UserName user={user} />,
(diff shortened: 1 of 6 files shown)


client/src/modules/dashboard/Events/pages/EventPage.tsx
```diff
@@ -308,69 +308,77 @@ export const EventPage: NextPageWithLayout = () => {
...
+                          ),
+                          action: () => (
+                            <VStack
+                              align={'flex-start'}
+                              fontSize={['sm', 'md']}
+                              spacing={'2'}
+                              marginBottom={4}
+                            >
+                              <UserName user={user} />
+                              <Text>{joined_date}</Text>
```
Sboonny, line 356:
```suggestion
                              <Text>{joined_date.slice(0, 10)}</Text>
```

client/src/modules/dashboard/Events/pages/EventPage.tsx
```diff
@@ -308,69 +308,77 @@ export const EventPage: NextPageWithLayout = () => {
                     role: ({ event_role }) => (
                       <Text data-cy="role">{event_role.name}</Text>
                     ),
+                    joined: ({ joined_date }) => <Text>{joined_date}</Text>,
```
Sboonny, line 311:
```suggestion
                    joined: ({ joined_date }) => <Text>{joined_date.slice(0, 10)}</Text>,
```

Sboonny (commented):
my only issue is the date format, I have tried to mess with the resolver so it will look like this, but with no luck.

Tried this With no luck

```ts
joined_date: new Date().toLocaleDateString("en-US")
joined_date: new Date().toLocaleString("en-US")
```

image


Is there a way to do format it in resolver, so it would nicer in the table? or do we have to use `slice` 🤔

gikf:
From the resolver - not really. For all formatting we can use [format](https://date-fns.org/v2.29.3/docs/format) or [formatRelative](https://date-fns.org/v2.29.3/docs/formatRelative) from `date-fns`. For the former one question is what options use. I'd imagine date (day, month, year) and time (hour, minutes) would be included.

Sboonny (approved):

Cool, I will mess with it tomorrow morning, thank you. 💚

PR #2592, opened 2023-04-14
refactor: split EventPage into components

gikf:

---
- More moving logic from `EventPage` to components

### DIFF
--- a/client/src/util/getLocationString.ts
+++ b/client/src/util/getLocationString.ts
@@ -1,4 +1,4 @@
-interface PartialLocation {
+export interface PartialLocation {
   street_address?: string | null;
   city: string;
   postal_code?: string;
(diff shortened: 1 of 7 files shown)


client/src/util/getLocationString.ts
```diff
@@ -1,4 +1,4 @@
-interface PartialLocation {
+export interface PartialLocation {
```
Sboonny, line 1: What do you think of linking the type to `Venue`, because it's the same type and same scope.


```ts
import { Venue } from '../generated/graphql';

export type Location = Pick<
  Venue,
  'street_address' | 'city' | 'postal_code' | 'region' | 'country'
>;
```

Sboonny (approved):

Issue #1927, opened 2022-11-18
The ability to transfer event to different chapter 


Sboonny:
### Discuss your ideas or share your views: 
Instance_owner can open event in any chapter, which may cause mistake (opening one in different chapter). 

They need to input the Event data again in different chapter, to create the same Event, and the chapter will send email about it cancelation, while it still available but in different chapter.

We can create a transfer function to take care of these issue automatically.

Kudos to Oliver for the awesome idea 👍

gikf:
While thinking on this, I thought it might be useful to give user ability to change some event details, before creating _transferred_ event. That might be some intermediate feature, to use existing event to pre-fill form.

Sboonny:
Good point, we may want to set the venue to TBD as default as well 🤔, they are part of chapter as far as I remember.

allella:
I wouldn't hold up the MVP for this feature and it looks like it's already on the Roadmap, so perhaps close it for now and revisit post-MVP?

I guess an alternative to a transfer would be to show the organizer a confirmation warning, or page, showing critical values that can't be changed after creation, like the _chapter_.

Eventually, we may also allow for different _organizations_ on the same _instance_, so a warning or transfer could eventually cover the _organization_ value too.

PR #2601, opened 2023-04-18
feat: event transfer

gikf:

Ref #1927 

---
- Fully reusing existing mutations. That's pretty cool (:sunglasses:), but let me know if I should rather create separate mutation for it in resolver, and offload logic there.
- I've added custom hooks to reuse canceling and event creating logic from client as much as possible.
- Original chapter is filtered out from the list of available chapters when transferring.
- Button to transfer is displayed only if user has admined `>= 2` chapters.
- Added header to the form, to make clearer what page is that (_Create Event_, _Edit Event_, _Transfer Event_).
- I'm going to still add tests, either to this PR or separate. But it can be looked at already.

### DIFF
--- a/client/src/hooks/useCancelEvent.ts
+++ b/client/src/hooks/useCancelEvent.ts
@@ -1,0 +1,35 @@
+import { useCancelEventMutation } from '../generated/graphql';
+import {
+  DASHBOARD_EVENT,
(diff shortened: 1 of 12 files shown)


ojeytonwilliams (approved):
This looks great, @gikf. Amazing work as always.

PR #2604, opened 2023-04-19
fix: don't display zero

gikf:

---
- I'm always getting it wrong when it's going to be displayed as `0`.
- I'm going to actually blame chakra for it, as I've just noticed it's even hiding the zeros, at least when parent component is `HStack` or `VStack`.


ojeytonwilliams (approved):
LGTM :+1: 

Also, thanks for the reminder that relying on the truthiness of length can have consequences!

